PR #14, opened 2015-07-17
:bug: Fix requirements in Full .NET tutorial

rowanmiller:
We only require net45 (rather than net451)
Remove reference to VS2010 since it only supports net40


natemcmaster:
:shipit:

PR #18, opened 2015-07-27
:memo: First cut of providers page

rowanmiller:


rowanmiller:
Will resubmit on beta6 branch

PR #25, opened 2015-07-29
:memo: Getting Started on UWP

rowanmiller:
First cut at getting started on UWP
Staged: http://ef.readthedocs.org/en/staging/getting-started/uwp.html
I'm still sorting out why syntax highlighting isn't working on the page


rowanmiller:
cc @bricelam @natemcmaster to see if there is anything else you hit that you think should be called out as a known issue etc.

natemcmaster:
Yes, there are issues in SQLite. See https://github.com/aspnet/EntityFramework.Docs/pull/26. There may be others specific to UWP.

rowanmiller:
@natemcmaster think I have those covered already (the Startup.cs one doesn't exists because UWP uses Package Manager Console rather than DNX)

natemcmaster:
:shipit: (assuming you've walked through the steps with VS RTM and it works). I didn't actually try it, but it looks right.

rowanmiller:
Yep I wrote it as I was building the sample :smile:

PR #26, opened 2015-07-29
Update OSX/Linux guides to include work arounds for issues in beta6

natemcmaster:
Can remove workarounds when aspnet/EntityFramework#2357 and aspnet/EntityFramework#2545 are resolved.

### DIFF
--- a/docs/getting-started/linux.rst
+++ b/docs/getting-started/linux.rst
@@ -32,10 +32,7 @@ Minimum system requirements
 .. caution::
     **Known Issues**
 
-     - Bugs in Mono 4.0.2 may cause Entity Framework to crash when using async methods. This is resolved with Mono >4.2.0, which has not yet been publicly released. `See this issue on GitHub <https://github.com/aspnet/EntityFramework/issues/2708>`_
-     - Migrations on SQLite do not support more complex schema changes due to limitations in SQLite itself.
-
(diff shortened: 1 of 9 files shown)


docs/getting-started/linux.rst
```diff
@@ -81,4 +78,4 @@ The following steps will install `dnvm <https://github.com/aspnet/home#running-a
 If you have trouble installing dnvm, consult this `Getting Started guide <http://dotnet.github.io/core/getting-started/>`_.
 
 
-.. include:: xplat.rst
\ No newline at end of file
+.. include:: x-plat/guide.rst
```
rowanmiller: assuming this fixes the build warning about the file not being included in a TOC?

rowanmiller:
:shipit:

docs/getting-started/linux.rst
natemcmaster: It doesn't. I just moved into x-plat so the files in this folder represent articles and includes come from the x-plat subfolder. I have noticed, btw, that the TOC on the left includes a :heavy_plus_sign:  button next the article title, even though there no subitems. Not sure how to fix this...

rowanmiller: Yeah I just copied how ASP.NET lays things out and they have the same issue with the :heavy_plus_sign:. I think we probably need to change the directory structure so that the subfolders are excluded from the build.

PR #27, opened 2015-07-30
Use .. contents:: directive

natemcmaster:
Autogen the "in this article" section


natemcmaster:
nvm... this is added in other PRs.

PR #28, opened 2015-07-30
Fix minor grammar issues

rhires:


rowanmiller:
@rhires thanks for the first external contribution to the docs!

PR #34, opened 2015-09-01
:arrow_up: Update docs to Beta 7

rowanmiller:
Updating docs and samples to Beta 7 - except for cross-plat ones which
will come in a separate commit.
Note that UWP is currently broken on Beta 7, so adding docs to use Beta
6 for the moment.


natemcmaster:
:shipit: I didn't actually test the samples on UWP, but I'm assuming you did.

rowanmiller:
Yep :smile: I started by updating the samples, then the docs

PR #50, opened 2015-10-24
Correct HTML in `Create.cshtml`

dougbu:
- #49
- `<div>`, `<label>` and `<span>` should always be written with end tags

nit: let VS have its way with `_references.js` file


natemcmaster:
:shipit:

dougbu:
d5f8910

PR #52, opened 2015-10-31
Update full-dotnet.rst

ejsd1989:
Clarified installation of Entity Framework when using Nuget Package Manager Tool.
- Added instruction on how to traverse to the Pack Sources settings in the menu
- Added instruction on appending the source to the command if desired.

### DIFF
--- a/docs/getting-started/full-dotnet.rst
+++ b/docs/getting-started/full-dotnet.rst
@@ -47,6 +47,13 @@ Install Entity Framework
 To use EF7 you install the package for the database provider(s) you want to target. This walkthrough uses SQL Server. For a list of available providers see :doc:`/providers/index`.
 
 * :menuselection:`Tools --> NuGet Package Manager --> Package Manager Console`
+* Prior to installation:
+* Either, ensure Package source is set to nuget.org:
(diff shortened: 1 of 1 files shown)


docs/getting-started/full-dotnet.rst
```diff
@@ -47,6 +47,13 @@ Install Entity Framework
...
 
 * :menuselection:`Tools --> NuGet Package Manager --> Package Manager Console`
+* Prior to installation:
+* Either, ensure Package source is set to nuget.org:
+* :menuselection:`Tools --> NuGet Package Manager --> Package Manager Settings --> Package Sources`
+* Check the box for nuget.org
+* :menuselection:`Package Manager Console --> Package source: --> nuget.org`
+* Or, Append to the end of the command ``-Source nuget.org``
+
 * Run ``Install-Package EntityFramework.SqlServer –Pre``
```
rowanmiller: I think this is a good idea, but I would rather see it called out in a `.. note::` box rather than just inline. The vast majority of customers on our stable releases will only have the NuGet.org feed... folks that have multiple are typically our early adopters who experiment with nightly builds (even folks that use our alpha/beta builds get them from NuGet.org). So I think of this as a reminder to our more advanced users.

rowanmiller:
Closing as no updates have been made

PR #53, opened 2015-11-06
Update index.rst to clarify .NET 4.5.1 requirement

ErikEJ:
Bump up to .NET 4.5.1 required

### DIFF
--- a/docs/providers/index.rst
+++ b/docs/providers/index.rst
@@ -14,7 +14,7 @@ EntityFramework.SqlServer
 
 **Database Engine:** Microsoft SQL Server (2008 onwards)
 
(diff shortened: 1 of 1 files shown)


natemcmaster:
LGTM.

rowanmiller:
I think the current release does work on .NET 4.5 right? I've been bumping the min-version to 4.5.1 as I work on the updates for RC1. It's fine to leave this in, but master is for the current stable release.

natemcmaster:
I haven't tested. I thought we bumped to 4.5.1 before beta8.

ErikEJ:
I have never tested on a Vanilla 4.5 Windows 8.0 machine with VS 2012. I think the required Nuget tools in VS do not work with VS 2012 anyway

PR #63, opened 2015-11-10
:memo: Add annotations and conventions to modeling docs

rowanmiller:
Also changing the modeling docs to be split into topics
The relationship content is still missing but I'll add that in a follow up PR


rowanmiller:
You can look at the staged content here http://ef.readthedocs.org/en/staging/modeling/index.html

ErikEJ:
@rowanmiller Any thought about the possibility of incorporating different modelling conventions used by other providers? Or will I just add these to my wiki?

As in this section:
(For example, SQL Server uses datetime2(7) for DateTime properties, and nvarchar(max) for string properties (or nvarchar(450) for string properties that are used as a key).

docs/modeling/generated-properties.rst
```diff
@@ -0,0 +1,99 @@
...
+    :depth: 3
+
+Value Generation Patterns
+-------------------------
+There are three value generation patterns that can be used for properties.
+
+No value generation
+^^^^^^^^^^^^^^^^^^^
+No value generation means that you will always supply a valid value to be saved to the database. This valid value must be assigned to new entities before they are added to the context.
+* **Value generated on add** means that if you don't specify a value, one will be generated for you.
```
natemcmaster: Duplicated from [this line](https://github.com/aspnet/EntityFramework.Docs/pull/63/files#diff-73bff1fb2b23f025c5afceb22f8f62eeR21)?

rowanmiller:
@ErikEJ the purpose of those is just as an example of the general principal. I could have used any provider, but SQL Server is the most common (at least for now) which is why I went with that. 

I am going to add some more content to the providers section to cover conventions, limitations, provider specific settings, etc., for the providers we ship (SQL Server, SQLite, and InMemory). I think similar content should exist in the docs for each provider - wherever they live.

Sound OK?

docs/modeling/generated-properties.rst
rowanmiller: Yep, I reorganized this content a few times... that would be a copy/paste mess up

docs/modeling/keys.rst
```diff
@@ -0,0 +1,67 @@
+.. include:: /rc1-notice.txt
+
+Keys (primary)
+==============
+
+A key serves as the primary unique identifier for each entity instance. When using a relational database this maps to the concept of a *primary key*.
```
natemcmaster: Could be useful to add a reference to the alternate keys page, something to the effect of:

    "You can also configure a unique identifier that is not the primary key. See `alternate keys`_"

docs/modeling/max-length.rst
```diff
@@ -0,0 +1,39 @@
+.. include:: /rc1-notice.txt
+
+Maximum Length
+==============
+
+Configuring a maximum length provides a hint to the data store about the appropriate data type to use for a given property. Maximum length only applies to array data types, such as ``string`` and ``byte[]``.
+
+.. note::
+    Entity Framework does not do any validation of maximum length before passing data to the provider. It is up to the provider or data store to validate if appropriate. For example, when targeting SQL Server, exceeding the maximum length will result in an exception as the data type of the underlying column will not allow excess data to be stored.
```
natemcmaster: And for comparison, when targeting SQLite, exceeding the maxlength does not throw an error because SQLite does not have max length contraints.

docs/modeling/relational-column.rst
```diff
@@ -0,0 +1,37 @@
...
+
+Conventions
+-----------
+
+By convention, each property will be setup to map to a column with the same name as the property.
+
+Data Annotations
+----------------
+
+You can use Data Annotations to configure the column that a property maps to.
```
natemcmaster: wording nit: "...to configure the column to which a property is mapped."

docs/modeling/relational-data-type.rst
```diff
@@ -0,0 +1,44 @@
+Relational Databases: Data Types
+================================
+
+Data type refers to the database specific type of the column a given property is mapped to.
```
natemcmaster: Same *wording nit* as above

docs/modeling/relational-table.rst
```diff
@@ -0,0 +1,53 @@
...
+.. note::
+    The configuration in this section is applicable to relational databases in general. The extension methods shown here will become available when you install a relational database provider (due to the shared *EntityFramework.Relational* package).
+
+.. contents:: In this article:
+    :depth: 3
+
+Conventions
+-----------
+
+By convention, each entity will be setup to map to a table with the same name as the entity.
```
natemcmaster: May be worth mentioning that SQLite's table names are case-insensitive, so adding mappings for casing-differences is unnecessary.

ErikEJ:
@rowanmiller  Sounds great!

docs/modeling/required-optional.rst
```diff
@@ -0,0 +1,39 @@
...
+.. contents:: In this article:
+    :depth: 3
+
+Conventions
+-----------
+
+By convention, a property whose CLR type can contain null will be configured as optional (``string``, ``int?``, ``byte[]``, etc.). Properties whose CLR type can not contain null will be configured as required (``int``, ``decimal``, ``bool``, etc.).
+
+.. note::
+    A property whose CLR type can not contain null can not be configured as optional, it will always be considered required by Entity Framework.
```
natemcmaster: Two wording nits: 
Cannot should be one word.
The first clause is a complete sentence. Use a semicolon or period.

docs/modeling/shadow-properties.rst
```diff
@@ -0,0 +1,52 @@
+.. include:: /rc1-notice.txt
+
+Shadow Properties
+=================
+
+Shadow properties are properties that do not exist in your entity class. The value and state of these properties is maintained purely in the Change Tracker.
```
natemcmaster: Will we have a Change Tracker article?

docs/modeling/shadow-properties.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Shadow property values can be obtained and changed thru the ``ChangeTracker`` API.
```
natemcmaster: through

docs/modeling/shadow-properties.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Data Annotations
+----------------
+
+Shadow properties can not be created with data annotations.
+
+Fluent API
+----------
+
+You can use the Fluent API to configure shadow properties. Once you have called the string overload of ``Property`` you can chain any of the configuration calls you would for other propertied.
```
natemcmaster: propertied

docs/modeling/included-types.rst
```diff
@@ -0,0 +1,45 @@
...
+.. contents:: In this article:
+    :depth: 3
+
+Conventions
+-----------
+
+By convention, types that are exposed in ``DbSet`` properties on your context are included in your model. In addition, types that are mentioned in the ``OnModelCreating`` method are also included. Finally, any types that are found by recursively exploring the navigation properties of discovered types are also included in the model.
+
+For example, in the following code listing all three types are discovered:
+ - ``Blog`` because it is exposed in a ``DbSet`` property on the context
```
natemcmaster: What about `context.Set<T>()`?

natemcmaster:
So far this looks good. Do we plan to add documentation for the rest of the Fluent API? We are missing examples for:

 - `HasAnnotation` 
 - `HasBaseType`
 - `HasDiscriminator`
 - `HasSequence`
 - `HasDefaultSchema`
 - `HasComputedColumnSql`
 - `HasDefaultValue`
 - `HasDefaultValueSql`

docs/modeling/shadow-properties.rst
rowanmiller: Eventually yes, or I may just point it to the API docs when they are available

docs/modeling/included-types.rst
rowanmiller: That won't include a type in the model since that API is only used after the model is already built.

rowanmiller:
Yes, I have a list of more things to incrementally add, I'm working on relationships at the moment and inheritance will be after that. Sequences are SQL Server specific, so they will go in provider docs I think.

rowanmiller:
@natemcmaster are you signed off if I address the comments you had?

smitpatel:
`[Timestamp]` is missing.

rowanmiller:
@smitpatel thanks, I will add it to my list

natemcmaster:
@rowanmiller `.HasSequence` is in relational, not just SqlServer. But yes, :shipit:

PR #64, opened 2015-11-11
:memo: Relationship docs

rowanmiller:
Note that this is not supposed to be a 100% exhaustive document of everything you can possibly do - since that would be a very large book. It's supposed to give folks the info they need to get started with the API. Our API docs provide all the specific little nuances of each variation of the API.

You can see the rendered version of the doc at http://ef.readthedocs.org/en/staging/modeling/relationships.html

### DIFF
--- a/docs/modeling/relationships.rst
+++ b/docs/modeling/relationships.rst
@@ -1,0 +1,266 @@
+.. include:: /rc1-notice.txt
+
+Relationships
+=============
+
+A relationship defines how two entities relate to each other. In a relational database, this is represented by a foreign key constraint.
+
+.. note::
+    Most of the samples in this article use a one-to-many relationship to demonstrate concepts. For examples of one-to-one and many-to-many relationships see the `Other Relationship Patterns`_ section at the end of the article.
+
+.. contents:: In this article:
+    :depth: 3
+
+Definition of Terms
+-------------------
+
+There are a number of terms used to describe relationships
+ - **Dependent entity:** This is the entity that contains the foreign key property(s). Sometimes referred to as the 'child' of the relationship.
+ - **Principal entity:** This is the entity that contains the primary/alternate key property(s). Sometimes referred to as the 'parent' of the relationship.
+ - **Foreign key:** The property(s) in the dependent entity that is used to store the values of the principal key property that the entity is related to.
+ - **Principal key:** The property(s) that uniquely identifies the principal entity. This may be the primary key or an alternate key.
+ - **Navigation property:** A property defined on the principal and/or dependent entity that contains a reference(s) to the related entity(s).
+ - **Collection navigation property:** A navigation property that contains references to many related entities.
+ - **Reference navigation property:** A navigation property that holds a reference to a single related entity.
+ - **Inverse navigation property:** When discussing a particular navigation property, this term refers to the navigation property on the other end of the relationship.
+
+The following code listing shows a one-to-many relationship between ``Blog`` and ``Post``
+  - ``Post`` is the dependent entity
+  - ``Blog`` is the principal entity
+  - ``Post.BlogId`` is the foreign key
+  - ``Blog.BlogId`` is the principal key (in this case it is a primary key rather than an alternate key)
+  - ``Post.Blog`` is a reference navigation property
+  - ``Blog.Posts`` is a collection navigation property
+  - ``Post.Blog`` is the inverse navigation property of ``Blog.Posts`` (and vice versa)
+
+.. literalinclude:: configuring/sample/EFModeling.Conventions/Samples/Relationships/Full.cs
+        :language: c#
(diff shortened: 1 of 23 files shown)


docs/modeling/relationships.rst
```diff
@@ -0,0 +1,236 @@
...
+
+.. contents:: In this article:
+    :depth: 3
+
+Definition of Terms
+-------------------
+
+There are a number of terms used to describe relationships
+ - **Dependent entity:** This is the entity that contains the foreign key property(s). Sometimes referred to as the 'child' of the relationship.
+ - **Principal entity:** This is the entity that contains the primary key property(s). Sometimes referred to as the 'parent' of the relationship.
```
bricelam: "...the ~~primary~~**principal** key properties..."?

docs/modeling/relationships.rst
```diff
@@ -0,0 +1,236 @@
...
+        :lines: 12-28
+        :linenos:
+
+Conventions
+-----------
+
+By convention, a relationship will be created when there is a navigation property discovered on a type. A property is considered a navigation property if the type it points to can not be mapped as a scalar type by the current database provider.
+
+.. note::
+    Relationships that are discovered by convention will always target the primary key of the principal entity. To target an alternate key, additional configuration must be performed using the Fluent API.
```
bricelam: Random thought, if alternate keys are defined, we could add `<EntityType><AlternateKey>` to the search for foreign key properties.

docs/modeling/relationships.rst
```diff
@@ -0,0 +1,236 @@
...
+
+If you want the foreign key to reference a property other than the primary key, you can use the Fluent API to configure the principal key property for the relationship. The property that you configure as the principal key will automatically be setup as an alternate key (see :doc:`alternate-keys` for more information).
+
+.. literalinclude:: configuring/sample/EFModeling.Configuring.FluentAPI/Samples/RelationshipPrincipalKey.cs
+        :language: c#
+        :lines: 7-39
+        :emphasize-lines: 11
+        :linenos:
+
+The following code listing shows how to configure a composite principal key.
```
bricelam: It might be worth mentioning in a caution that the order of the properties needs to match the order of the foreign key properties.

bricelam:
:shipit:

I hate to say it, but I actually like where this API finally landed.

docs/modeling/relationships.rst
smitpatel: What would `<AlternateKey>` read?

docs/modeling/relationships.rst
```diff
@@ -0,0 +1,236 @@
...
+       :linenos:
+
+.. caution::
+    If there are multiple navigation properties defined between two types (i.e. more than one distinct pair of navigations that point to each other), then no relationships will be created by convention and you will need to manually configure them to identify how the navigation properties pair up.
+
+No Foreign Key Property
+^^^^^^^^^^^^^^^^^^^^^^^
+
+While it is recommended to have a foreign key property defined in the dependent entity class, it is not required. If no foreign key property is found, a shadow foreign key property will be introduced with the name ``<principal type name><principal key property name>`` (see :doc:`shadow-properties` for more information).
+
```
smitpatel: Name would be `<Navigation property name><principal key property name>` if navigation pointing to principal is available.

docs/modeling/relationships.rst
```diff
@@ -0,0 +1,236 @@
...
+.. literalinclude:: configuring/sample/EFModeling.Configuring.DataAnnotations/Samples/RelationshipForeignKey.cs
+        :language: c#
+        :lines: 13-31
+        :emphasize-lines: 17
+        :linenos:
+
+.. note::
+    The ``[ForeignKey]`` annotation can be placed on either navigation property in the relationship. It does not need to go on the navigation property in the dependent entity class.
+
```
smitpatel: `[ForeignKey]` can also be put on property and also putting on property essentially defines the dependent end.

rowanmiller: Yeah I decided to just leave that out... the endless list of things you can do are so complicated I opted for keeping things "concise".

docs/modeling/relationships.rst
bricelam: E.g. `PersonSSN` would by-convention be the FK to the `Person` navigation property if a property named `PersonId` doesn't exist.

smitpatel: To match a property, we first use the referenced property name. So based on names of principal key properties, fk properties with similar names will be found. If this gives us no match then we use `Id` suffix if there is only 1 property to reference. If `Person` navigation property is configured to use `PrincipalKey(e => e.SSN)` then the `PersonSSN` will be fk property even if `PersonId` also exists in the entity type. `PersonId` will be configured only if `PersonSSN` is not available.

PR #66, opened 2015-11-12
:art: Copy CSS form ASP.NET

rowanmiller:
This is a direct copy of the CSS from the ASP.NET repo with two minor tweaks:
* Removed the ASP.NET logo file. We don't have an approved EF logo that will work in such a small place yet.
* Remove the reference to the logo from the css file.

There is some CSS that won't be used on our site... but I left the file as unchanged as possible so that future updates are easier.


natemcmaster:
:shipit:

natemcmaster:
So they won't let us use this?
ef

PR #67, opened 2015-11-12
[TINY] Resolve build warnings

rowanmiller:
Several build warnings had crept in, builds cleanly with this change.


natemcmaster:
:shipit:

PR #68, opened 2015-11-13
[SMALL] Moved shared content to separate directory

rowanmiller:
Moving the files that are used throughout the documents to an _shared
directory


docs/getting-started/aspnet5.rst
```diff
@@ -1,5 +1,3 @@
-.. include:: /stub-topic.txt
-
```
smitpatel: intentional removal?

smitpatel:
:shipit:

docs/getting-started/aspnet5.rst
rowanmiller: Yeah it was left over from when this article wasn't written (this include just makes the icon available for use... it doesn't actually render any content).

PR #71, opened 2015-11-16
:memo: Existing database tutorial for ASP.NET 5

rowanmiller:


docs/getting-started/aspnet5/existing-db.rst
```diff
@@ -0,0 +1,207 @@
...
+        :language: json
+        :linenos:
+        :lines: 26-29
+        :emphasize-lines: 3
+
+Reverse engineer your model
+---------------------------
+
+.. caution::
+    The reverse engineer experience in ASP.NET 5 is still a work-in-progress. The following steps are overly complex and will be simplified by the time we reach a stable release.
```
natemcmaster: > overly complex

Agreed. We can improve this UX. Can we link to any plans or issues tracking our hopes for better reveng UX?

rowanmiller: The main thing is enabling Init.ps1 in xproj, so that we can use PowerShell commands... then it just becomes the same as the other project types (i.e. no dropping out to cmd)

PR #72, opened 2015-11-17
:memo: Complete relational modeling docs

rowanmiller:
Covering the remaining relational specific configuration APIs.
Also splitting some relational docs that were inline in general configuration into the new relational sub-section


natemcmaster:
:shipit:

PR #82, opened 2015-11-24
Fix in sequence name and schema

michaelrp:
Changed to "shared.OrderNumbers" to match Sequence.cs and SequenceConfigured.cs samples.


rowanmiller:
Thanks for the contribution!

Issue #80, opened 2015-11-21
Mac OS X ef migration and SQLite.Design


rhires:
I don't know if it's just me, or something else. In upgrading from beta8 to rc1, I deleted my migrations folder, and the database, and changed all the names for the assemblies in all the right places, and changed by project.json file to -rc1-final for everything...and then re-ran my migration `dnx ef migrations add MyFirstMigration` - I got stopped by a missing "EntityFramework.Sqlite.Design": "7.0.0-rc1-final" assembly. I figured it out after a little bit of research, but it's not in the docs for the example. Is this something that should be added to the example for the OS X getting started piece?

rhires:
Part of the error I got included `Could not load type 'Microsoft.Data.Entity.Scaffolding.Sqlite Design Time Services, EntityFramework.Sqlite.Design'.` - I'm adding this for the search engines. Maybe help someone else.

natemcmaster:
Can you share your project.json? I can't repro again [this sample](https://github.com/aspnet/EntityFramework.Docs/tree/master/docs/getting-started/x-plat/sample) (which is what we walkthorugh in the OSX getting started guide).

rhires:
Here's the whole thing. I worry that I didn't adequately explain before. I was running with beta8, then upgraded to rc1-final, and when I did, I had to add the Sqlite.Design line in order for my migrations to go forward. I upgraded previously from beta6 or beta7, so I threw out my entire migrations folder and the database and ran the migration from scratch. Also, running on Mac OS X Yosemite w/ Mono. 

{
  "webroot": "wwwroot",
  "version": "1.0.0-*",
  "tooling": {
    "defaultNamespace": "TestApp"
  },

  "dependencies": {
    "Microsoft.AspNet.Diagnostics": "1.0.0-rc1-final",
    "Microsoft.AspNet.Diagnostics.Entity": "7.0.0-rc1-final",
    "Microsoft.AspNet.IISPlatformHandler": "1.0.0-rc1-final",
    "Microsoft.AspNet.Mvc": "6.0.0-rc1-final",
    "Microsoft.AspNet.Mvc.TagHelpers": "6.0.0-rc1-final",
    "Microsoft.AspNet.Server.Kestrel": "1.0.0-rc1-final",
    "Microsoft.AspNet.StaticFiles": "1.0.0-rc1-final",
    "Microsoft.AspNet.Tooling.Razor": "1.0.0-rc1-final",
[...]

PR #83, opened 2015-11-24
Add issue to getting started guides about case-sensitive package names

natemcmaster:
See #80 and a few others.
cc @rowanmiller


rowanmiller:
:shipit:

PR #88, opened 2015-12-07
[SMALL] Clarify value generation may require setup

rowanmiller:
Per discussion on
https://github.com/aspnet/EntityFramework/issues/39551, adding some
notes to clarify that just marking a property as generated is not always
enough to actually get values generated.

### DIFF
--- a/docs/modeling/generated-properties.rst
+++ b/docs/modeling/generated-properties.rst
@@ -17,7 +17,12 @@ No value generation means that you will always supply a valid value to be saved 
 Value generated on add
 ^^^^^^^^^^^^^^^^^^^^^^
 
-Value generated on add means that if you don't specify a value, one will be generated for you.
+Value generated on add means that a value is generated for new entities.
+
+.. caution::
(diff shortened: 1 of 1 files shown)


ajcvickers:
:shipit:

bcbeatty:
Documentation for http://ef.readthedocs.org/en/latest/modeling/generated-properties.html#id7 is inconsitent. 

With the caution section on [Add or Update](http://ef.readthedocs.org/en/latest/modeling/generated-properties.html#value-generated-on-add-or-update),  it states

> However, if you specify that a DateTime property is generated on add or update, then you must setup a way for the values to be generated (such as a database trigger).

but in the example for [Fluent](http://ef.readthedocs.org/en/latest/modeling/generated-properties.html#value-generated-on-add-fluent-api) 
>modelBuilder.Entity<Blog>()
>                .Property(b => b.LastUpdated)
              .ValueGeneratedOnAddOrUpdate()

has no such example.

rowanmiller:
@bcbeatty the detailed explanation is included in the main section, and there is a caution box in the Fluent and Data Annotation sections that points back to the detailed note.

> This just lets EF know that values are generated for added entities, it does not guarantee that EF will setup the actual mechanism to generate values. See Value generated on add section for more details.

PR #94, opened 2015-12-15
make it clear that this is EF7

rhires:


rowanmiller:
We actually already have a header on every page that calls out that this doc site is for EF7
header

rhires:
Ok, https://docs.asp.net/en/latest/data/index.html doesn't show the same way as your screen shot above...does that mean it hasn't yet been merged/published to the main site? 

screenshot

smitpatel:
This repo is published at http://ef.readthedocs.org/en/latest/
The site you referred is from https://github.com/aspnet/Docs repo

rhires:
Ok, so it looks like I got lost in the file structure somehow. Closing.

PR #103, opened 2016-01-04
:memo: Document min PowerShell version

### LINKED ISSUE #101: Powershell version on Windows 7
If one follows the tutorial [here](http://ef.readthedocs.org/en/latest/getting-started/full-dotnet/new-db.html) and does so on a box running Windows 7, chances are that they'll receive the following error ...

```
The term 'Add-Migration' is not recognized as the name of a cmdlet, function, script file,
or operable program. Check the spelling of the name, or if a path was included, verify
that the path is correct and try again.
At line:1 char:14
+ Add-Migration <<<<  MyFirstMigration
    + CategoryInfo          : ObjectNotFound: (Add-Migration:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```
... when trying to execute `Add-Migration MyFirstMigration`.

This is due to the fact that Powershell may be Version 2 or 3. It needs to be Version 4 for this to work.
I had this issue on two separate W7 boxes on which I tried playing with EF7, so I think other people may run into this as well. The first time it took me almost a day to figure this out. So a hint towards this in the tutorial may be a good idea.
Installing an update from [here](https://www.microsoft.com/en-us/download/details.aspx?id=40855) fixed it.

rowanmiller:
For Package Manager Console commands you need PowerShell 3.0 or newer.
Adding this to the pre-req section. Doesn't apply to UWP since you need
Windows 10 for that anyway.
Resolves #101

### DIFF
--- a/docs/getting-started/full-dotnet/existing-db.rst
+++ b/docs/getting-started/full-dotnet/existing-db.rst
@@ -19,6 +19,7 @@ The following prerequisites are needed to complete this walkthrough:
 
 * Visual Studio 2013 or Visual Studio 2015
 * `Latest version of NuGet Package Manager`_
+* `Latest version of Windows PowerShell <https://www.microsoft.com/en-us/download/details.aspx?id=40855>`_
 * `Blogging database`_
 
 Latest version of NuGet Package Manager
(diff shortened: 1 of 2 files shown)


natemcmaster:
:shipit:

mehrt:
I don't think that Version 3.0 mentioned in the initial comment  will suffice. I explicitly tried installing 3.0 first, and it didn't work on my W7 machine. Only installing 4.0 fixed it. Since the link added to the docs points to version 4, that should work though.

rowanmiller:
@mehrt the content in the docs just says "The latest version", so I think we are good

PR #104, opened 2016-01-09
Document powershell commands

natemcmaster:
This matches the RC1 commands.

#78 

cc @rowanmiller

PR #106, opened 2016-01-11
Update UWP guide to include rdxml gen

natemcmaster:
:pig: Piggy back off #104 
cc @rowanmiller


docs/cli/dotnet.rst
```diff
@@ -0,0 +1,8 @@
+.. include:: /_shared/stub-topic.txt
+
+|stub-icon| dotnet CLI
+===================
```
rowanmiller: You should be getting an "underline too short" warning here?

docs/getting-started/uwp.rst
```diff
@@ -23,13 +23,9 @@ In this walkthrough, you will build a Universal Windows Platform (UWP) applicati
...
-In this article:
-	- `Prerequisites`_
-	- `Create a new project`_
-	- `Install Entity Framework`_
-	- `Create your model`_
-	- `Create your database`_
-	- `Use your model`_
+.. contents:: `In this article:`
+    :depth: 1
+    :local:
```
rowanmiller: Just double check this renders correctly... I think it might only give you the H1 heading of the page and you may need depth=2 to give you the actual section headings

docs/getting-started/uwp.rst
```diff
@@ -166,3 +162,33 @@ You can now run the application to see it in action.
 .. image:: uwp/_static/create.png
 
 .. image:: uwp/_static/list.png
+
+Generate runtime directives
+---------------------------
+
```
rowanmiller: Add a note that this is only needed for deploying... you don't need to do this over and over again during local development

rowanmiller: In fact, I would phrase this section as "stuff you need to know for deployment" rather than making a step in the step-by-step walkthough

docs/cli/dotnet.rst
natemcmaster: I didn't notice the warnings. I'll fix this

docs/getting-started/uwp.rst
natemcmaster: I checked. It gets the headings correctly

natemcmaster:
I've rephrased the scaffold-directives section. Does it sound better?

rowanmiller:
:shipit:

PR #109, opened 2016-01-15
[TINY] Min PowerShell version in package manager console docs

rowanmiller:
Based on feedback https://github.com/aspnet/EntityFramework.Docs/issues/101#issuecomment-171230564
cc @natemcmaster


natemcmaster:
:shipit:

PR #112, opened 2016-01-21
Improvements to provider docs

rowanmiller:
Split each provider out into a separate section to avoid cramming all
the info onto a single page
Introduce a limitations doc for SQLite

@bricelam can you take a look at this one? docs/providers/SQLite/limitations.rst is the only new content, the rest is just restructuring what we already had.


bricelam:
:shipit:

PR #117, opened 2016-01-26
Testing with InMemory provider

### LINKED ISSUE #95: Testing with InMemory
The most important thing is how to scope the database to a single test...

```c#
using Microsoft.Data.Entity;
using Microsoft.Data.Entity.Infrastructure;
using Microsoft.Extensions.DependencyInjection;
using System;
using System.Linq;

namespace ConsoleApp1
{
    public class Program
    {
        public static void Main(string[] args)
        {
            var serviceCollection = new ServiceCollection();
            serviceCollection
                .AddEntityFramework()
                .AddInMemoryDatabase()
                .AddDbContext<SampleContext>(c => c.UseInMemoryDatabase());


            using (var db = serviceCollection.BuildServiceProvider().GetService<SampleContext>())
            {
                db.Blogs.Add(new Blog { Url = "Test" });
                db.SaveChanges();
                Console.WriteLine(db.Blogs.Count());
            }

            using (var db = serviceCollection.BuildServiceProvider().GetService<SampleContext>())
            {
                db.Blogs.Add(new Blog { Url = "Test" });
                db.SaveChanges();
                Console.WriteLine(db.Blogs.Count());
            }
        }
    }

    public class SampleContext : DbContext
    {
        public DbSet<Blog> Blogs { get; set; }
    }

    public class Blog
    {
        public int BlogId { get; set; }
        public string Url { get; set; }
    }
}
```

rowanmiller:
Resolves #95

### DIFF
--- a/docs/getting-started/aspnet5/new-db.rst
+++ b/docs/getting-started/aspnet5/new-db.rst
@@ -35,7 +35,7 @@ Create a new project
 	* Click **OK**
 
 .. caution::
-    If you use **Individual User Accounts** instead of **None** for **Authentication** then an Entity Framework model will be added to your project in `Models\\IdentityModel.cs`. Using the techniques you will learn in this walkthrough, you can chose to add a second model, or extend this existing model to contain your entity classes.
+    If you use **Individual User Accounts** instead of **None** for **Authentication** then an Entity Framework model will be added to your project in `Models\\IdentityModel.cs`. Using the techniques you will learn in this walkthrough, you can choose to add a second model, or extend this existing model to contain your entity classes.
 
 Install Entity Framework
 ----------------------------------------
--- a/docs/miscellaneous/index.rst
+++ b/docs/miscellaneous/index.rst
@@ -6,3 +6,4 @@ Miscellaneous
     :caption: The following articles are available
 
     logging
+    testing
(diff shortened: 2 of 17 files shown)


ajcvickers:
Looks good.

Consider pointing out that the in-memory database is not a fell relational database and that using SqLite in in-memory mode might be a better option if this is needed.

Consider suggesting that if the context can be modified to take an IServiceProvider in its constructor, then all the CreateScope stuff should not be needed. This is what we typically do in our tests.

:shipit:

rowanmiller:
I made some pretty significant changes based on your suggestions (to try and minimize the interaction with DI as much as possible). Of note, I pulled out DbContextOptions from DI so that the test code looks more like it will when we have named instances for InMemory (and IServiceProvider is used purely to scope the InMemory database instance).

The updated version is staged at http://ef.readthedocs.org/en/staging/miscellaneous/testing.html

rowanmiller:
@ajcvickers could you take another quick look based on comments above

ajcvickers:
Looks great. :shipit:

rowanmiller:
Merged https://github.com/aspnet/EntityFramework.Docs/commit/47d1ccef3db8e2c9e1aba04a073bae1cc545a1e7

PR #119, opened 2016-01-28
.NET Native

natemcmaster:


natemcmaster:
cc @rowanmiller

rowanmiller:
As discussed, move this under a UWP folder along with the getting started. I'll so some restructuring later to make it "Platforms" rather than "Getting Started"
:shipit:

PR #121, opened 2016-01-28
Writing an EF provider

natemcmaster:
First draft of the "getting-started" equivalent for provider writers.

cc @rowanmiller @divega


docs/internals/sample/.gitattributes
```diff
@@ -0,0 +1,63 @@
+###############################################################################
```
rowanmiller: Should go at root level and not in this subdirectory

docs/internals/sample/.gitignore
```diff
@@ -0,0 +1,156 @@
+## Ignore Visual Studio temporary files, build results, and
```
rowanmiller: Shouldn't be needed as we have a .gitignore in the root directory

docs/internals/writing-a-provider.rst
```diff
@@ -0,0 +1,134 @@
+===========================
+Writing an EF Core Provider
```
rowanmiller: "Writing a Database Provider"

docs/internals/writing-a-provider.rst
```diff
@@ -0,0 +1,134 @@
...
+
+``Microsoft.EntityFrameworkCore.Infrastructure.DbContextOptions`` is the API surface for **users** to configure ``DbContext``. Provider
+writers are responsible for creating API to configure options and to make
+services responsive to these options. For example, most providers require a
+connection string. These options are
+typically created using ``DbContextOptionsBuilder``.
+
+Services
+^^^^^^^^
+
```
rowanmiller: Should also cover how the services get registered when the user is not using an external ServiceCollection.

docs/internals/writing-a-provider.rst
```diff
@@ -0,0 +1,134 @@
...
+  :lines: 8-18
+
+.. tip::
+  The ``UseX()`` method can also be used to return a special wrapper around
+  ``DbContextOptionsBuilder`` that allows users to configure multiple options
+  with change calls. See ``SqlServerDbContextOptionsBuilder`` as an example.
+
+The `Add` Method
+^^^^^^^^^^^^^^^^
+
```
rowanmiller: We should provide a list of all the services that need to be implemented (or tell them how to derive it from the code). I think this is as simple as needing to implement everything that is exposed from `DatabaseProviderServices`? Though we should cover that those things should be resolved from DI and not just explicitly returned from an implementation of `DatabaseProviderServices`. 

We should also explain what `DatabaseProviderServices` is somewhere in this doc, since it is pretty fundamental to how things work.

rowanmiller:
There is obviously a ton of stuff we could add to this doc, but I think if you cover what I mentioned then this a good start for now.

rowanmiller:
:shipit: once that info is added

docs/internals/writing-a-provider.rst
rowanmiller: Ah, that is in the Understanding Services page... perhaps under the "Add Method" section you could call out that in some scenarios users will call the AddX method directly, and in other cases it will be called by EF.

rowanmiller: And point to the section in the services page where that process is explained

docs/internals/writing-a-provider.rst
rowanmiller: Scrap the above comment, I see that is explained in the Understanding Services page.

However, this code listing is a little misleading because it doesn't really register all the services needed to make things work. I would expand this to actually show it registering all the services - which are then resolved and returned from the `DatabaseProviderServices` implementation.

PR #125, opened 2016-01-29
Configuring a dbcontext

### LINKED ISSUE #90: Configuring a DbContext
Moved from https://github.com/aspnet/EntityFramework/issues/3919

We should add a header to https://github.com/aspnet/EntityFramework/wiki/Configuring-a-DbContext that points to the proper docs

natemcmaster:
Fix #90 

cc @rowanmiller  @divega

### DIFF
--- a/docs/miscellaneous/configuring-dbcontext.rst
+++ b/docs/miscellaneous/configuring-dbcontext.rst
@@ -1,0 +1,139 @@
+Configuring a DbContext
+=======================
+
+This article shows patterns for configuring a ``DbContext`` with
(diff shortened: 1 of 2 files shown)


rowanmiller:
This is a good start so :shipit: 

I have a draft outline that I want to use to "tell more of a story" around how contexts are configured - so that folks have a better idea of when to use which one... but that is just some re-arranging and extra text on what you already have here, so I'll just retrofit it :smile:

natemcmaster:
Merged in https://github.com/aspnet/EntityFramework.Docs/commit/3f01199fa28406530cd5e9226d87b6e9acb3aa8c

PR #134, opened 2016-02-11
Backing fields

rowanmiller:


docs/modeling/backing-field.rst
```diff
@@ -0,0 +1,45 @@
...
+  :depth: 2
+  :local:
+
+Conventions
+-----------
+
+By convention, the following fields will be discovered as backing fields for a given property (listed in precedence order):
+  * <propertyName> differing only by case
+  * _<propertyName>
+  * m_<propertyName>
```
rowanmiller: This is slightly simplified from the actual rules, as we have some additional precedence around casing of the field name... but it seems like such a corner case to have two fields that match a property name and only differ by case that I chose to keep it simple.

ajcvickers:
:shipit:

PR #141, opened 2016-02-27
Fix build warnings

rowanmiller:
Fixing various build warnings that had crept in.
C# syntax highlighter doesn't support string interpolation yet, so just
moving to concatenation for the moment.


smitpatel:
:shipit:

PR #159, opened 2016-05-10
Add document with preview1 known issues for dotnet ef

natemcmaster:
This page is the target of the fwlink we want to add in https://github.com/aspnet/EntityFramework/pull/5312.

cc @rowanmiller

### DIFF
--- a/docs/cli/dotnet.rst
+++ b/docs/cli/dotnet.rst
@@ -3,16 +3,31 @@
 
 EF command-line tools for .NET Core Command Line Interface (CLI).
 
+.. contents:: `In this article:`
+    :depth: 2
+    :local:
+
 .. note::
     Command-line tools for .NET Core CLI has known issues. See `Preview 1 Known Issues`_ for more details.
 
+Installation
+------------
+
 Prerequisites
--------------
+~~~~~~~~~~~~~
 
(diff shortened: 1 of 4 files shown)


docs/cli/dotnet.rst
```diff
@@ -1,8 +1,161 @@
...
+    {
+        "frameworks": {
+            "net451": {
+                "buildOptions": {
+                    "emitEntryPoint": true
+                }
+            }
+        }
+    }
+
```
rowanmiller: If their context doesn't have a default constructor they will need to add an implementation of IDbContextFactory, we should include an example for the default context created by the templates (which has a DbContextOptions constructor param).

docs/cli/powershell.rst
```diff
@@ -1,7 +1,7 @@
 Package Manager Console (Visual Studio)
 =======================================
```
rowanmiller: The known issues from CLI also apply to PMC commands for ASP.NET Core and .NET Core applications... we should tell folks that for those project types we just invoke the CLI commands, and link to the CLI known issues section.

rowanmiller:
:shipit: once the comments are addressed

docs/cli/dotnet.rst
natemcmaster: Added a small sample of using IDbContextFactory to this PR. See https://github.com/aspnet/EntityFramework.Docs/pull/159/commits/120d3a686beb3d977aa052367c74d348479b00b6

PR #160, opened 2016-05-11
[rc2] Update configuring context guide

natemcmaster:
Changes:

 - Samples show `DbContextOptions<T>` instead of non-generic option
 - Separate the section about AddDbContext from explanation of constructor arg vs onconfiguring

cc @ajcvickers @rowanmiller


rowanmiller:
:shipit:

PR #164, opened 2016-05-17
Only add a space between the code blocks

TiagoSoczek:
Without this space, the layout is broken


rowanmiller:
Thanks for the fix!

PR #168, opened 2016-05-20
Updated the latest nuget library version.

tanaka-takayoshi:
The latest version of "Microsoft.EntityFrameworkCore.Tools" Nuget library is 1.0.0-preview1-final.
https://www.nuget.org/packages/Microsoft.EntityFrameworkCore.Tools/1.0.0-preview1-final


natemcmaster:
Thanks, good catch. We missed this one.

PR #170, opened 2016-05-24
Document side-by-side powershell commands

natemcmaster:
Add explanation of how to invoke EF6 and EF Core commands in sxs installation.

cc @rowanmiller


rowanmiller:
:shipit: 
Ultimately this probably belongs in an "EF6/EFCore SxS" doc, but we can move it when we have that.

natemcmaster:
Merged to staging branch as this change won't be effective until RTM.

https://github.com/aspnet/EntityFramework.Docs/commit/8abf6ff7c8c7b3d1b257f6b01bcf6472b0fafcb6

PR #186, opened 2016-06-10
Update .NET Core getting started guide

natemcmaster:
Unify the OS X and Linux getting started guide as the only difference between the two was how to install the .NET Core SDK.

Update the guide for RTM/Preview 2.


rowanmiller:
:shipit:

PR #187, opened 2016-06-10
Update dotnet-ef doc

natemcmaster:
Update document including explanation of workarounds and common errors.

### DIFF
--- a/docs/cli/dotnet.rst
+++ b/docs/cli/dotnet.rst
@@ -31,7 +31,7 @@ Install by editing project.json
 
 EF command-line tools for .NET Core CLI are installed by manually editing ``project.json``.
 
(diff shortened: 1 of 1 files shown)


lajones:
Looks good to me, but perhaps @rowanmiller should review the error messages. But just to check, is it true that e.g. a `-c` option just after `dotnet ef` will _not_ clash with a `-c` option specified after the command? So e.g. `dotnet ef -c <some_configuration> some_command -c <some_other_option>` will work correctly?

natemcmaster:
Correct, these parameters don’t clash. Order matters though. This is something we want to revisit in https://github.com/aspnet/EntityFramework/issues/5188

lajones:
OK - as far as I'm concerned :shipit:

rowanmiller:
:shipit:

PR #189, opened 2016-06-21
Adding some "Saving" articles

rowanmiller:
Here is a local build of the content. To view it, download, unzip, open up html/index.html.
[html.zip](https://github.com/aspnet/EntityFramework.Docs/files/326677/html.zip)

The docs to be reviewed are:
* Saving Data -> Related Data
* Saving Data -> Concurrency Conflicts
* Saving Data -> Transactions

### DIFF
--- a/docs/saving/sample/EFSaving/Concurrency/Sample.cs
+++ b/docs/saving/sample/EFSaving/Concurrency/Sample.cs
@@ -1,0 +1,96 @@
+﻿using Microsoft.EntityFrameworkCore;
+using System;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+
+namespace EFSaving.Concurrency
+{
+    public class Sample
+    {
+        public static void Run()
+        {
+            // Ensure database is created and has a person in it
+            using (var context = new PersonContext())
+            {
+                context.Database.EnsureDeleted();
+                context.Database.EnsureCreated();
+
+                context.People.Add(new Person { FirstName = "John", LastName = "Doe" });
+                context.SaveChanges();
+            }
+
+            using (var context = new PersonContext())
+            {
+                // Fetch a person from database and change phone number
+                var person = context.People.Single(p => p.PersonId == 1);
+                person.PhoneNumber = "555-555-5555";
+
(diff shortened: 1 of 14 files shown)


rowanmiller:
cc @ajcvickers @AndriySvyryd if one (or both) of you would have time to review

ErikEJ:
The sample code here looks odd: file:///C:/Users/Erik/Downloads/html/saving/related-data.html#removing-relationships

docs/saving/sample/EFSaving/RelatedData/Sample.cs
```diff
@@ -0,0 +1,95 @@
...
+                context.SaveChanges();
+            }
+
+            // Removing relationships
+            using (var context = new BloggingContext())
+            {
+                var blog = context.Blogs.Include(b => b.Posts).First();
+                var post = blog.Posts.First();
+
+                post.Blog = blog;
```
ajcvickers: This doesn't look like a remove.

docs/saving/sample/EFSaving/Concurrency/Sample.cs
```diff
@@ -0,0 +1,96 @@
...
+                            }
+                        }
+                        else
+                        {
+                            throw new NotSupportedException("Don't know how to handle concurrency conflicts for " + entry.Metadata.Name);
+                        }
+                    }
+
+                    // Retry the save operation
+                    context.SaveChanges();
```
ajcvickers: This really should be part of a loop, otherwise a concurrency exception here will not be handled.

ajcvickers:
:shipit:

docs/saving/sample/EFSaving/RelatedData/Sample.cs
rowanmiller: copy/paste/fail... thanks 😵

docs/saving/sample/EFSaving/Concurrency/Sample.cs
rowanmiller: Will do, I'm going to make this change in a follow up commit

PR #196, opened 2016-06-24
Add "Querying Data" section

rowanmiller:
Here is a local build of the content. To view it, download, unzip, open up html/index.html.
[html.zip](https://github.com/aspnet/EntityFramework.Docs/files/332575/html.zip)

The new content under the "Querying Data" section from the left menu of the site. All pages should be reviewed, as it is all new.

cc @anpete @ajcvickers as it would be good to get your eyes over this

### DIFF
--- a/docs/querying/overview.rst
+++ b/docs/querying/overview.rst
@@ -1,0 +1,45 @@
+How Query Works
+===============
+
+Entity Framework Core uses Language Integrate Query (LINQ) to query data from the database. LINQ allows you to use C# (or your .NET language of choice) to write strongly typed queries based on your derived context and entity classes.
+
+.. contents:: `In this article:`
+    :depth: 2
+    :local:
+
+The life of a query
+-------------------
+
+The following is a high level overview of the process each query goes through.
+
+1. The LINQ query is processed by Entity Framework Core to build a representation that is ready to be processed by the database provider
+
+  a. The result is cached so that this processing does not need to be done every time the query is executed
+
+2. The result is passed to the database provider
+
+  a. The database provider identifies which parts of the query can be evaluated in the database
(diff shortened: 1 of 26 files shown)


ajcvickers:
It feels like having the "How Query Works" section first might not be right since it has a lot of concepts in it. It's good for people to know this stuff, and I like the way it is written, but I would probably not have it has the first bullet in the list.

rowanmiller:
@ajcvickers yep that makes sense, possibly as the last bullet?

ajcvickers:
@rowanmiller I'm still reading through the rest, but yes, making it the last bullet seems right.

docs/querying/overview.rst
```diff
@@ -0,0 +1,48 @@
...
+3. For each item in the result set
+
+  a. If this is a tracking query, EF checks if the data represents an entity already in the change tracker for the context instance
+
+    * If so, the existing entity is returned
+    * If not, a new entity is created, change tracking is setup, and the new entity is returned
+
+  b. If this is a no-tracking query, EF checks if the data represents an entity already in the result set for this query
+
+    * If so, the existing entity is returned
```
ajcvickers: This isn't 100% true. If the application starts to run out of memory, then identity resolution stops working. Because it is impossible to know if/when this might happen I would consider it dangerous to write an application that depends on this behavior. I'm not sure what the best way to document this is.

ajcvickers:
:shipit:

docs/querying/overview.rst
rowanmiller: Going to file an issue to discuss this more with @anpete and then update the docs. We need to work out when/if/how this would impact the code you write for an app.

PR #197, opened 2016-06-24
Reorganize top level structure

rowanmiller:
Reduces number of nodes at the root level and adjusts names/ordering to make the docs flow better

cc @natemcmaster


natemcmaster:
:shipit:

PR #198, opened 2016-06-24
Landing page

rowanmiller:
Provide a landing page that covers what EF is, and directs off to the
various parts of the docs.
Using shared content rendered on two pages so that it shows up on the
"index" page, but also has a node in the left menu.


rowanmiller:
@ajcvickers would be great to get your feedback on this one 😄

ajcvickers:
Looks really good. :shipit:

PR #62, opened 2015-11-10
Update x-plat guide for RC1

natemcmaster:
Use CoreCLR by default in the tutorials.

PR #203, opened 2016-06-28
on line #62, change "context" to "db"

github5775:
You have not created "context" in the using statement, you created "db"


github5775:
ty

PR #206, opened 2016-06-29
More visible docs on installing

rowanmiller:
Also tidying up some of the provider docs to use a more consistent format etc.


natemcmaster:
:shipit:

PR #207, opened 2016-06-29
Update instructions for installing .NET Core 1.0 on UWP

natemcmaster:
Apparently Microsoft.NETCore.UniversalWindowsPlatform 5.2.0 is unlisted on nuget.org. This means `Update-Package` doesn't work and can only be installed with Package Manager Console and adding the `-Version` parameter.

cc @weshaggard @rowanmiller


rowanmiller:
:shipit:

PR #219, opened 2016-07-11
Changed ``dotnet ef --help`` to ``dotnet -ef --help``

ronaldnsabiyera:
The command ``dotnet ef --help`` throws "No executable found matching the command" but when I run ``dotnet -ef --help`` it run fine

### DIFF
--- a/docs/miscellaneous/rc1-rc2-upgrade.rst
+++ b/docs/miscellaneous/rc1-rc2-upgrade.rst
@@ -103,7 +103,7 @@ If you have an ASP.NET Core application and you want EF to resolve internal serv
 DNX Commands => .NET CLI (ASP.NET Core Projects Only)
 -----------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


Kagamine:
I don't think this is ok to merge, you were wrong with the command line.

To execute `dotnet -ef --help` is equal to execute `dotnet --help`

You should enter a directory which contains a `project.json` and added ef tools into your `project.json`

Then `dotnet ef --help` works well.

natemcmaster:
@ronaldnsabiyera Thanks for attempting to help update our docs. As @Kagamine noted however, the syntax in the documentation `dotnet ef --help` is correct so we won't be adding this change. 

If you are still encountering the error "No executable found matching the command", then check that your project matches the installation criteria (see https://docs.efproject.net/en/latest/miscellaneous/cli/dotnet.html#installation) and that you are executing dotnet.exe from the directory containing your project.json.

PR #221, opened 2016-07-12
Add Pomelo MySQL provider

Kagamine:
/cc @rowanmiller @natemcmaster


rowanmiller:
Hey,

Can you help us understand a little more about your project. Our main concerns with listing providers is knowing that they have a decent quality level and that they are going to be supported long term. We don't want to push customers toward a provider that is partially implemented, or that they aren't going to get support for in a years time. There is very little history on your GitHub project, so I'm curious where the project came from, who is behind it, etc.

~Rowan

rowanmiller:
Hey,

Sorry for the long delay on this one. We're going to take the listing of this provider, but I need some changes to what is submitted here. We really just want changes to `docs/providers/index.rst` and `docs/providers/pomelo-mysql/index.rst`. The samples folder is really just for our more detailed articles, I think you can just inline the code samples for this page. You may also want to consider just linking off to your own documentation, which is what the other provider writers do.

If you rather I make the changes, then just let me know.

~Rowan

Kagamine:
@rowanmiller Certainly, you can make these changes.

Besides, we have implemented a new ado layer for mysql, it is not forked form the official repository and port to .net core, it is a independent library which called [MySqlConnector](https://www.nuget.org/packages/MySqlConnector/0.1.0-alpha19), and it is a real MIT licensed and async supported lib.

So our mysql provider supported async operations and more efficiency than the official one. It's  also under a friendly open source license: MIT license.

Thanks for support us.

Kagamine:
We have another product, I opened #252, Could you have a look if you have time?

rowanmiller:
New PR opened https://github.com/aspnet/EntityFramework.Docs/pull/260

PR #224, opened 2016-07-22
Probaly it's just a documentation issue.

ste4net:
I opened also an issue to efcore repo
https://github.com/aspnet/EntityFramework/issues/6127


rowanmiller:
This looks like a bug in EF rather than a documentation issue. Marking a property as NotMapped should not exclude the type from the model - unless that property was the only way it was included in the model. We'll follow up on the issue in the runtime repository.

PR #229, opened 2016-08-07
Workaround for Add-Migration command failure

NageshAndani:
Added a workaround note to reopen Visual Studio in case if Package Manager fails to recognize Add-Migration command. This is a tested behaviour of Package Manager

### DIFF
--- a/docs/platforms/aspnetcore/new-db.rst
+++ b/docs/platforms/aspnetcore/new-db.rst
@@ -107,7 +107,7 @@ Create your database
 Now that you have a model, you can use migrations to create a database for you.
 
 * :menuselection:`Tools –> NuGet Package Manager –> Package Manager Console`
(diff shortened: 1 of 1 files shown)


natemcmaster:
@NageshAndani Can you clarify what you meant by "This is a tested behaviour of Package Manager"?  I've verified that the 'Add-Migration' command is available without restart when following the tutorial steps. Do you have the latest tooling installed?   cref https://docs.efproject.net/en/latest/platforms/aspnetcore/new-db.html#prerequisites

NageshAndani:
@natemcmaster I mean, when I followed this doc to setup a sample app to work on Entity Framework, I got an error due to Add-Migration command was not recognized. From StackOverflow threads, I got to know the workaround of restarting VS.

The same workaround is mentioned in EF's other doc too (Database First page).

I am using VS 2015 with ASP.NET Core 1.0 + Tools Preview 2

rowanmiller:
Thanks for the contribution!

PR #232, opened 2016-08-11
Sequence sample lines reference fix

nikibobi:
The original included part of the Order class:
```
public class Order
```
Now it is gone. If the intention was to include the whole class then it should reference lines `5-20` instead.


natemcmaster:
Thanks for noticing. The intention is to include 'Order' class. If you can update this patch, I'll merge it.

PR #234, opened 2016-08-13
Fix broken Microsoft.EntityFrameworkCore.InMemory package link

scottaddie:
There was an extraneous "h" in the `Microsoft.EntityFrameworkCore.InMemory` NuGet package link, which caused it to not work.


natemcmaster:
Thanks @scottaddie

PR #263, opened 2016-09-14
Add note about tracking use for nav property fix-up

tdykstra:
cc @divega

### DIFF
--- a/docs/querying/tracking.rst
+++ b/docs/querying/tracking.rst
@@ -1,7 +1,7 @@
 Tracking vs. No-Tracking
 ========================
 
-Tracking behavior controls whether or not Entity Framework Core will keep information about an entity instance in its change tracker. If an entity is tracked, any changes detected in the entity will be persisted to the database during ``SaveChanges()``.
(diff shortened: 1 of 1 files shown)


docs/querying/tracking.rst
```diff
@@ -1,7 +1,7 @@
 Tracking vs. No-Tracking
 ========================
 
-Tracking behavior controls whether or not Entity Framework Core will keep information about an entity instance in its change tracker. If an entity is tracked, any changes detected in the entity will be persisted to the database during ``SaveChanges()``.
+Tracking behavior controls whether or not Entity Framework Core will keep information about an entity instance in its change tracker. If an entity is tracked, any changes detected in the entity will be persisted to the database during ``SaveChanges()``. Tracking is also used by Entity Framework Core to automatically fix-up navigation properties in entities that were previously loaded into the context instance.
```
divega: Consider rephrasing to something like this:

> Entity Framework Core will also fix-up navigation properties between entities that are obtained from a tracking query and entities that were previously loaded into the `DbContext` instance. 

cc @rowanmiller

PR #268, opened 2016-09-23
Update dotnet.rst

Ketrex:
The original text did not account for the fact that most class library projects only target NETStandardApp. Projects of this type cannot reference Microsoft.EntityFrameworkCore.Tools. Attempting run any migrations even when referencing a valid startup project will still fail with the error 'No executable found matching command "dotnet-ef"' as the tools are not referenced in the target project. Both workarounds require that the target project not be a class library.


rowanmiller:
Can you update the underlines on the headings you changed to match the new length. There is a build warning if they don't match.

Ketrex:
Thanks for the advice, @rowanmiller!

rowanmiller:
Thanks for the contribution!

PR #269, opened 2016-10-04
:memo: Better clarify that alternate key != unique index

rowanmiller:
Resolves aspnet/EntityFramework/issues/6644


rowanmiller:
cc @ajcvickers for review

PR #276, opened 2016-10-18
Delete index.rst

DalSoft:
Paid for only, documentation should be impartial / contain open or at least free providers only. Otherwise it will end up as a advertising hub.


rowanmiller:
The purpose of our provider listing is to let folks know what is available. DevArt is a valid option and we've had feedback from customers that their providers are very good. We clearly call out that this is a commercial provider, so folks have that info up front.

DalSoft:
On reflection I agree,  I guess it might change if lots of commercial options spring up. And as you say it's clearly called out and devart support providers where there are no other options at the moment.

PR #282, opened 2016-10-28
Initial Markdown Migration

v-anpasi:


rowanmiller:
Looks like this has the `_site` directory being added, I think you want to add that to `.gitignore` right?

v-anpasi:
@rowanmiller I am not familiar with a lot of the configuration details of docfx. If this folder needs to be removed we can take care of that at a later date.

rowanmiller:
`_site` is the rendered output when you build the site locally. So it's build output.

rowanmiller:
But yes, we can remove it later if you prefer

PR #305, opened 2016-11-23
Update index.md

Ponant:
Change title "Saving Data" to "Saving and Deleting Data"

### DIFF
--- a/entity-framework/core/saving/index.md
+++ b/entity-framework/core/saving/index.md
@@ -10,7 +10,7 @@ ms.technology: entity-framework-core
  
 uid: core/saving/index
 ---
-# Saving Data
+# Saving and Deleting Data
 
 > [!NOTE]
 > This documentation is for EF Core. For EF6.x, see [Entity Framework 6](../../ef6/index.md).


bradygaster:
### :white_check_mark: Validation status: passed


File | Status | Preview URL | Details
---- | ------ | ----------- | -------
[entity-framework/core/saving/index.md](https://github.com/Ponant/EntityFramework.Docs/blob/patch-1/entity-framework/core/saving/index.md) | :white_check_mark:Succeeded | |

For more details, please refer to the [build report](https://opbuildstorageprod.blob.core.windows.net/report/2016%5C11%5C23%5Cc1705ea6-142e-cb02-f3b5-8dd98c6d1733%5CPullRequest%5C201611231119481563-305%5Cworkflow_report.html).

**Note:** If you changed an existing file name or deleted a file, broken links in other files to the deleted or renamed file are listed only in the full build report.

rowanmiller:
Saving covers insert, update, and delete... so I think we'll leave the title as-is

Ponant:
Being explicit is always a better option in my view, especially for docs. I came up with this proposition because it took me some time to find info about deletion....

rowanmiller:
Ok, I can see that. How about `Saving Data (Add, Update, and Delete)` to be super specific. BTW you would also need to update toc.md in the `./entity-framework` folder to update what shows up in the left menu.

divega:
What about "Saving changes" or "Saving changes made to data"?

Ponant:
@divega, in my opinion Saving Changes requires an inference, whereas **Saving Data (Add, Update, Delete)** or a shorter version **Save, Add, Update, Delete** is more epxlicit
As a side note, as I exchanged a few [lines](https://github.com/aspnet/EntityFramework/issues/7110) about docs with Rowan I find that the docs contain some jargons which is fine for people knowing the matter, but as Scott Guthrie/Hanselman pointed out in Connect2016(), .Net Core has circa 40% of new comers, which is my case (I come from scientific programming Fortran 77, C++, matlab). This means that focussing on streamlined content that allows the user to read and write code at the same time and bring him to the higher level of understanding of EF is good practice if you want people to be encouraged to continue instead of switching to simpler languages/frameworks.

The guide so far goes along those lines, I think one should only be more specific and simplify the wording. Eventually, this doc is already way better than the one of EF6 and I learn a lot. I also read a few blogs from @ajcvickers, such the one on INotifyPropertyChanged, perhaps they could be included if permission is granted.
I am on EF Core for the next month as it represents a step in our building of an app; we can profit from this time that I give you feedback. What I can do, if you want is to try to come up with a structure for the docs where I do not see major changes but where I will have in mind the beginner's/mid-experienced reader while explaining a bit more what is going on. For instance when we scaffold a migration the text says somewhere in the Get Started
"Run dotnet ef migrations add MyFirstMigration to scaffold a migration to create the initial set of tables for your model."
Here I would just add a 2-3 lines explaining which files have been created and which files have been modified. Etc. I am not sure if GitHub allows for annotations where I push to you some recommendations. I could use the livefyre sidenotes in the docs website but I expect them to become public.

PR #358, opened 2017-02-27
Explain transaction support with connection resiliency

### LINKED ISSUE #334: Consider explaining transaction support with connection resiliency
This is a follow up to *Connection Resiliency does not support transactions?  https://github.com/aspnet/EntityFramework/issues/7318* and in particular including some explanation on why connection resiliency cannot work with user initiated transactions (and attempt as explaining this was made at https://github.com/aspnet/EntityFramework/issues/7318#issuecomment-269822171).

rowanmiller:
Resolves #334


rowanmiller:
@AndriySvyryd can you review this? Need to look at changes to `connection-resiliency.md` and the `Sample` region in `Program.cs`

bradygaster:
### :white_check_mark: Validation status: passed


For more details, please refer to the [build report](https://opbuildstorageprod.blob.core.windows.net/report/2017%5C2%5C27%5Cc1705ea6-142e-cb02-f3b5-8dd98c6d1733%5CPullRequest%5C201702272331355248-358%5Cworkflow_report.html).

**Note:** If you changed an existing file name or deleted a file, broken links in other files to the deleted or renamed file are listed only in the full build report.

rowanmiller:
@AndriySvyryd the `ExecuteInTransaction` API seems kinda hard to use... but perhaps I'm missing something 😄. Here is what I could come up with...

```c#
using (var db = new BloggingContext())
            {
                var strategy = db.Database.CreateExecutionStrategy();

                strategy.ExecuteInTransaction<object, int>(
                    operation: (_) =>
                    {
                        using (var transaction = db.Database.BeginTransaction())
                        {
                            db.Blogs.Add(new Blog { Url = "http://blogs.msdn.com/dotnet" });
                            db.SaveChanges();

                            db.Blogs.Add(new Blog { Url = "http://blogs.msdn.com/visualstudio" });
                            db.SaveChanges();

                            transaction.Commit();
                        }

                        return 0;
                    },
                    verifySucceeded: (_) =>
                    {
                        return db.Blogs
                            .Where(b => b.Url == "http://blogs.msdn.com/dotnet" 
                                        || b.Url == "http://blogs.msdn.com/visualstudio")
                            .Count() == 2;
                    },
                    state: null,
                    context: db);
            }
```

rowanmiller:
The main issues were:
* Having to specify a `TResult` when there isn't a result. I just used a fake int result to make it compile
* Having to specify `TState` when there isn't any state
* Having to specify `state` and `context` parameters

Are there ways to avoid this? At the moment it seems more complex than the code example I included.

rowanmiller:
I pushed these changes, but will definitely add the ExecuteInTransaction example once we lock on it

AndriySvyryd:
@rowanmiller Thanks, I think these can be solved by extension methods, I've added your comments to https://github.com/aspnet/EntityFramework/issues/7431

rowanmiller:
@AndriySvyryd do you think it is worth adding the example I came up with, or should we just wait for the API to be simpler? Does the example without `ExecuteInTransaction` just fail if the transaction fails to commit, or does it retry anyway, possibly duplicating an operation that succeeded?

AndriySvyryd:
I think it would be good to have the example now and we can simplify it later.

A common use for the state parameter is to supply the context instance to the lambdas so that it's not captured from the outer scope, you can do that to avoid using placeholders.

PR #391, opened 2017-04-19
Update logging.md

luisgepeto:
Review simple error


divega:
@luisgepeto thanks for the fix.

PR #522, opened 2017-10-13
Add a page for IModelCacheKeyFactory

### LINKED ISSUE #484: Explain how to use ModelCacheKeyFactory
Please provide a documentation on how to use `ModelCacheKeyFactory` to change model cache keys. The need to use `ModelCacheKeyFactory` comes from this ticket: https://github.com/aspnet/EntityFrameworkCore/issues/9985

AndriySvyryd:
Fixes #484

### DIFF
--- a/entity-framework/core/modeling/dynamic-model.md
+++ b/entity-framework/core/modeling/dynamic-model.md
@@ -1,0 +1,18 @@
+---
+title: Alternating between multiple models with the same DbContext type - EF Core
+author: AndriySvyryd
+
+uid: core/modeling/dynamic-model
+---
+# Alternating between multiple models with the same DbContext type
+
+The model built in `OnModelCreating` could use a property on the context to change how the model is built. For example it could be used to exclude a certain property:
+
+[!code-csharp[Main](../../../samples/core/DynamicModel/DynamicContext.cs?name=Class)]
+
+## IModelCacheKeyFactory
+However if you tried doing the above without additional changes you would get the same model every time a new context is created for any value of `IgnoreIntProperty`. This is caused by the model caching mechanism EF uses to improve the performance by only invoking `OnModelCreating` once and caching the model.
+
+By default EF assumes that for any given context type the model will be the same. To accomplish this the default implementation of `IModelCacheKeyFactory` returns a key that just contains the context type. To change this you need to replace the `IModelCacheKeyFactory` service. The new implementation needs to return an object that can be compared to other model keys using the `Equals` method that takes into account all the variables that affect the model:
+
+[!code-csharp[Main](../../../samples/core/DynamicModel/DynamicModelCacheKeyFactory.cs?name=Class)]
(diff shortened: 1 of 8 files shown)


AndriySvyryd:
@divega Could you check style/formatting?

entity-framework/core/modeling/dynamic-model.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Dynamic Models - EF Core
```
divega: @AndriySvyryd @ajcvickers I am a bit concerned that the term "Dynamic" may convey magic capabilities that we don't have, but I am struggling to come up with something better for the title, e.g. "Changing the model at runtime" isn't really more constrained, or for the class name, e.g. `ParameterizedDbContext` isn't great. Any better ideas? If it is too hard I can settle for "dynamic".

entity-framework/core/modeling/dynamic-model.md
```diff
@@ -0,0 +1,59 @@
...
+author: ansvyryd
+
+uid: core/modeling/dynamic-model
+---
+# Dynamic Models
+
+The model built in `OnModelCreating` could use a property on the context to change how the model is built. For example it could be used exclude a certain property:
```
divega: missing word: used ***to*** exclude

entity-framework/core/modeling/dynamic-model.md
```diff
@@ -0,0 +1,59 @@
...
+
+<!-- [!code-csharp[Main](samples/core/DynamicModel/DynamicContext.cs?range=6-31)] -->
```
divega: You should be able to uncomment this code reference and get rid of the snippet below. Also, consider relying on regions (e.g. DynamicContext.cs?#Class) rather than on line numbers because the latter are easier to break.

entity-framework/core/modeling/dynamic-model.md
```diff
@@ -0,0 +1,59 @@
...
+            }
+        }
+    }
+}
+```
+
+## IModelCacheKeyFactory
+However if you tried doing the above without additional changes you would get the same model every time a new context is created for any value of `IgnoreIntProperty`. This is caused by the model caching mechanism EF uses to improve the performance.
+
+By default EF assumes that for any given context type the model will be the same. To change this you need to replace the `IModelCacheKeyFactory` service. The new implementation needs to return an object that can be compared to other model keys using the `Equals` method that takes into account all the variables that affect the model:
```
divega: Consider expanding, e.g.: hence `OnModelCreating` only needs to be invoked once.

divega: Consider mentioning what the default implementation of the service does.

entity-framework/core/modeling/dynamic-model.md
ajcvickers: Maybe something along the lines of, "Using multiple models with the same context type"? Or "One DbContext with many models"?

divega: I like that. It can be read as using all the models at the same time, but not sure it is a major issue... Anyway, what about "Alternating between multiple models with the same DbContext type"?

ajcvickers: Sounds good.

PR #533, opened 2017-10-17
Update the UWP page to 2.x

### LINKED ISSUE #446: Update UWP sample to 2.0 when .NET Standard 2.0 is supported
Currently it is on Microsoft.NETCore.UniversalWindowsPlatform 5.4 and EF Core 1.1.2.

### LINKED ISSUE #467: Update UWP guidance to recommend creating the model on a separate project
Currently EF Core 2.0 tooling will not work directly with EF Core models in .NET Core. This is described in more detail at https://github.com/aspnet/EntityFrameworkCore/issues/9666. 

In the short term, we need to update the guidance for working with UWP projects to place the EF Core model in a separate project that is compatible with our tooling.

This could be a .NET Standard library which either cross compiles for .NET Core, is used by a .NET Core console app.

AndriySvyryd:
Fixes #446 
Fixes #467


entity-framework/core/get-started/uwp/getting-started.md
```diff
@@ -13,9 +13,6 @@ uid: core/get-started/uwp/getting-started
 
 # Getting Started with EF Core on Universal Windows Platform (UWP) with a New Database
 
-> [!NOTE]  
-> Temporarily this tutorial uses EF Core 1.1. UWP has not been updated yet to support .NET Standard 2.0 which is required for compatibility with EF Core 2.0. Once it is, we will update the tutorial to use the new version. 
-
 In this walkthrough, you will build a Universal Windows Platform (UWP) application that performs basic data access against a local SQLite database using Entity Framework.
 
 > [!WARNING]
```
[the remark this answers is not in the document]
AndriySvyryd: I don't think it would result in a crash, but I suspect that perf will still be noticeably lower.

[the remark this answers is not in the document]
divega: I would like to change it to an "important" note and use less prescriptive language. Ideally we would have a list of catches of using EF Core with .NET Native in a visible place, e.g. in the "Supported .NET Implementations" page and link to it form here, but that doesn't need to happen in this PR. I will create an issue.

entity-framework/core/get-started/uwp/getting-started.md
```diff
@@ -13,49 +13,42 @@ uid: core/get-started/uwp/getting-started
 
 # Getting Started with EF Core on Universal Windows Platform (UWP) with a New Database
 
-> [!NOTE]  
-> Temporarily this tutorial uses EF Core 1.1. UWP has not been updated yet to support .NET Standard 2.0 which is required for compatibility with EF Core 2.0. Once it is, we will update the tutorial to use the new version. 
+-> [!NOTE]
+-> This tutorial uses EF Core 2.0.1. EF Core 2.0.0 lacks some crucial bugfixes required for a good UWP experience.
```
ajcvickers: If we're going to mention 2.0.1, we should probably also mention that it will be released as the 2.0.3 meta package, or at least add an appropriate link when available.

entity-framework/core/get-started/uwp/getting-started.md
```diff
@@ -13,49 +13,42 @@ uid: core/get-started/uwp/getting-started
...
+* [.NET Core 2.0.0 SDK](https://www.microsoft.com/net/core) or later.
+
+* [Visual Studio 2017](https://www.visualstudio.com/downloads/) version 15.4 or later with the **Universal Windows Platform Development** workload.
 
-* The latest version of [Windows 10 Developer Tools](https://dev.windows.com/downloads)
+## Create a new model project
 
-## Create a new project
+> [!WARNING]
+> Due to a limitation in the Entity Framework tools the model needs to be placed in a non-UWP project to be able to run migrations commands in the Package Manager Console
```
ajcvickers: I don't think we should make this sound like it is just EF tools that are limited. Maybe, "Due to limitations in the way .NET Core tools interact with UWP..."

entity-framework/core/get-started/uwp/getting-started.md
divega: Do you think UWP customers will use the ASP.NET Core metapackage useful?

divega:
There are too many things in this tutorial that won’t work until we release EF Core 2.0.1 patch. Not planning to merge until then.

entity-framework/core/get-started/uwp/getting-started.md
ajcvickers: Good point. 😸 However, the announcements/tweets/etc. people will see will likely say 2.0.3, so I think it's still worth saying something like, "released as part of the wider 2.0.3 ASP.NET/EF Core patch release" or something like that.

divega: Ok, with mentioning it between parenthesis.

entity-framework/core/get-started/uwp/getting-started.md
```diff
@@ -248,6 +146,9 @@ You can now run the application to see it in action.
 
 ## Next steps
 
+> [!TIP]
+> `SaveChanges()` performance can be improved by implementing [`INotifyPropertyChanged`](https://msdn.microsoft.com/en-us/library/system.componentmodel.inotifypropertychanged.aspx), [`INotifyPropertyChanging`](https://msdn.microsoft.com/en-us/library/system.componentmodel.inotifypropertychanging.aspx), [`INotifyCollectionChanged`](https://msdn.microsoft.com/en-us/library/system.collections.specialized.inotifycollectionchanged.aspx) in your entity types and using `ChangeTrackingStrategy.ChangingAndChangedNotifications`.
+
 Tada! You now have a simple UWP app running Entity Framework.
 
-Check out the numerous articles in this documentation to learn more about Entity Framework's features.
+Check out other articles in this documentation to learn more about Entity Framework's features.
```
divega: I prefer referring to the product as Entity Framework Core.

divega:
@anpete while doing the UWP verification for 2.0.1 found that VS fails to add a project reference to a class library that has both targets: `<TargetFrameworks>netcoreapp2.0;netstandard2.0</TargetFrameworks>`. If you add the reference first and then add the netcoreapp2.0 target, it will kind of work but complain with a yellow bang on the reference. We don't know what else could be broken. Given that, I believe we should just recommend customers create a separate netcoreapp2.0 console app that targets the same netstandard2.0 library.

AndriySvyryd:
I think @ryanbrandenburg ran into the same issue

divega:
@AndriySvyryd Now that the patch is out, customers should be able to follow the updated walkthrough, right? Is there anything else that needs to be updated?

AndriySvyryd:
@divega It's up-to-date AFAIK

PR #535, opened 2017-10-17
Add a sample showing how to configure scale and precision

### LINKED ISSUE #527: Document HasPrecision

AndriySvyryd:
Fixes #527

### DIFF
--- a/entity-framework/core/modeling/relational/data-types.md
+++ b/entity-framework/core/modeling/relational/data-types.md
@@ -25,20 +25,14 @@ For example, SQL Server uses `datetime2(7)` for `DateTime` properties, and `nvar
 
 ## Data Annotations
 
-You can use Data Annotations to specify an exact data type for the column.
+You can use Data Annotations to specify an exact data type for a column.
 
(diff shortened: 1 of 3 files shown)


entity-framework/core/modeling/relational/data-types.md
```diff
@@ -27,18 +27,10 @@ For example, SQL Server uses `datetime2(7)` for `DateTime` properties, and `nvar
...
-    public int BlogId { get; set; }
-    [Column(TypeName = "varchar(200)")]
-    public string Url { get; set; }
-}
-```
+[!code-csharp[Main](../../../../samples/core/Modeling/DataAnnotations/Samples/Relational/DataType.cs?name=Entities&highlight=4,6)]
 
 ## Fluent API
 
 You can use the Fluent API to specify an exact data type for the column.
```
divega: Perhaps we should say mention what facets we are configuring in both examples: "e.g. this configures a string with x length, and a decimal with precision y, scale z". This may help search engines find this page, which is important, especially given we don't have separate APIs for this as we did in EF6.

PR #538, opened 2017-10-19
Update Connection Resiliency docs

### LINKED ISSUE #405: Fix and expand the connection resiliency docs
For https://docs.microsoft.com/en-us/ef/core/miscellaneous/connection-resiliency

Add docs for `ExecuteInTransaction` overloads added in https://github.com/aspnet/EntityFramework/pull/8751 showing how to use the `verifySucceeded` parameter.
* The supplied operation lamda needs to be idempotent.
* Need to call `SaveChanges(acceptAllChangesOnSuccess: false)` as the transaction is committed outside of `SaveChanges` and might fail. Call `context.ChangeTracker.AcceptAllChanges()` after calling `ExecuteInTransaction`

AndriySvyryd:
Fixes #405

### DIFF
--- a/entity-framework/core/miscellaneous/connection-resiliency.md
+++ b/entity-framework/core/miscellaneous/connection-resiliency.md
@@ -10,10 +10,7 @@ uid: core/miscellaneous/connection-resiliency
 
 # Connection Resiliency
 
-> [!NOTE]  
-> This feature is new in EF Core 1.1.
-
-Connection resiliency automatically retries failed database commands. It is specific to relational databases.
+Connection resiliency automatically retries failed database commands. It is supported by all relational databases.
 
 ## SQL Server
 
@@ -25,7 +22,7 @@ An execution strategy is specified when configuring the options for your context
 
 ## Custom execution strategy
 
-There is a mechanism to register a custom execution strategy of your own.
+There is a mechanism to register a custom execution strategy of your own if you wish to change any of the defaults.
 
 ``` csharp
 protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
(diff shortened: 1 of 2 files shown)


entity-framework/core/miscellaneous/connection-resiliency.md
```diff
@@ -10,10 +10,7 @@ uid: core/miscellaneous/connection-resiliency
 
 # Connection Resiliency
 
-> [!NOTE]  
-> This feature is new in EF Core 1.1.
-
-Connection resiliency automatically retries failed database commands. It is specific to relational databases.
+Connection resiliency automatically retries failed database commands. It is supported by all relational databases.
```
ajcvickers: Does "supported" here imply that it works out-of-the-box for any relational database?

entity-framework/core/miscellaneous/connection-resiliency.md
```diff
@@ -39,12 +36,57 @@ protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
...
+By default, the execution strategy will retry the operation as if the transaction was rolled back, but if it's not the case this will result in an exception if the new database state is incompatible or could lead to **data corruption** if the operation does not rely on a particular state, for example when inserting a new row with auto-generated key values.
+
+There are several ways to deal with this.
+
+### Option 1 - Do (almost) nothing
+
+The likelihood of a connection failure during transaction commit is low so it may be acceptable for your application to just fail if this condition actually occurs.
+
+However, you need to avoid using store-generated keys in order to ensure that an exception is thrown instead of adding a duplicate row. Consider using a client-generated GUID value or a client-side value generator such as `ForSqlServerUseSequenceHiLo`.
+
```
ajcvickers: We shouldn't recommend people use ForSqlServerUseSequenceHiLo. It has other implications and is confusing for people who don't understand what is happening.

entity-framework/core/miscellaneous/connection-resiliency.md
AndriySvyryd: An execution strategy can be specified for any relational provider. But we only provide one for SQL Server

ajcvickers: Yes, I understand that, I just think "supported" is ambiguous here. @divega can make the final call.

divega: I would say something like connection resiliency _can_ be supported by any relational provider by implementing an execution strategy, and that the SQL Server provider does it.

divega: I ended up coming up with something a bit more elaborate and more appropriate for the context (given the next section is on the SQL Server execution strategy). I will send a separate PR for review.

PR #545, opened 2017-10-26
Fix typo in configuring-dbcontext.md

mikeaag:


divega:
Thanks @mikeaag for the fix!

PR #548, opened 2017-10-28
Suggest using nuget package for helper method

leomwa:
Was not obvious as to why I was not getting the option.


divega:
Thanks @leomwa for the fix!

PR #551, opened 2017-11-05
Relocate text trying to improve readibility

pollirrata:


divega:
Thanks for the improvement!

PR #553, opened 2017-11-05
Update pomelo provider get start docs

yukozh:
@divega


divega:
Thanks for the update @yukozh!

PR #563, opened 2017-11-10
Document Migrations

### LINKED ISSUE #204: Migrations
Our getting started guides mentions migrations and shows how to add one, but we have a dedicated section to document the rest of what is possible with migrations.

Some specific things to cover in addition to a general overview
 - [x] Overview of how migrations works, something like the content at the start of [EF6.x Migrations in Team Environments](https://msdn.microsoft.com/en-us/library/dn481501(v=vs.113).aspx)
 - [x] how to use --startup-project (#294)
 - [x] Multiple projects and migrations (#43)
 - [x] Custom Migrations history table (#371)
 - [x] IDbContextFactory
 - [x] [Custom Migrations Operations](https://romiller.com/2013/02/27/ef6-writing-your-own-code-first-migration-operations/)
 - [x] Using IMigrator (#518)

bricelam:
Resolves #204

### DIFF
--- a/entity-framework/core/managing-schemas/migrations/index.md
+++ b/entity-framework/core/managing-schemas/migrations/index.md
@@ -1,0 +1,201 @@
+---
+title: Migrations - EF Core
+author: bricelam
+ms.author: bricelam
+ms.date: 10/30/2017
+ms.technology: entity-framework-core
+---
+Migrations
+==========
+Migrations provide a way to incrementally apply schema changes to the database to keep it in sync with your EF Core
+model while preserving existing data in the database.
+
+Creating the database
+---------------------
+After you've [defined your initial model][1], it's time to create the database. To do this, add an initial migration.
+Install the [EF Core Tools][2] and run the appropriate command.
+
+``` powershell
+Add-Migration InitialCreate
+```
+``` Console
+dotnet ef migrations add InitialCreate
+```
+
+Tree files are added to your project under the **Migrations** directory:
+
+* **00000000000000_InitialCreate.cs**--The main migrations file. Contains the operations necessary to apply the
+  migration (in `Up()`) and to revert it (in `Down()`).
+* **00000000000000_InitialCreate.Designer.cs**--The migrations metadata file. Contains information used by EF.
+* **MyContextModelSnapshot.cs**--A snapshot of your current model. Used to determine what changed when adding the next
+  migration.
+
+The timestamp in the filename helps keep them ordered chronologically so you can see the progression of changes.
+
+> [!TIP]
+> You are free to move Migrations files and change their namespace. New migrations are created as siblings of the last
+> migration.
+
+Next, apply the migration to the database to create the schema.
+
+``` powershell
+Update-Database
+```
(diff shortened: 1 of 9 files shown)


entity-framework/core/managing-schemas/migrations/index.md
```diff
@@ -0,0 +1,198 @@
...
+
+Applying migrations at runtime
+------------------------------
+Some apps may want to apply migrations at runtime during startup or first run. Do this using the `Migrate()` method.
+
+``` csharp
+myDbContext.Database.Migrate();
+```
+
+> [!WARNING]
```
ajcvickers: I think we should be more explicit about not running migrations at application startup except in very specific cases, such as testing or client apps with a local database. The advice we have generally been giving is that even if you want to use the update-database command on your production database it should only be done in a controlled environment as part of a controlled application deployment.

entity-framework/core/managing-schemas/migrations/teams.md
```diff
@@ -0,0 +1,62 @@
...
+When you merge migrations from your teammates, you may get conflicts in your model snapshot file. If both changes are
+unrelated, the merge is trivial and the two migrations can coexist. For example, you may get a merge conflict in
+the customer entity type configuration that looks like this:
+
+``` Conflict file
+<<< Mine
+b.Property<bool>("Deactivated");
+===
+b.Property<int>("LoyaltyPoints");
+>>> Theirs
```
bricelam: Wow, that didn't render right...

entity-framework/core/managing-schemas/migrations/operations.md
```diff
@@ -0,0 +1,143 @@
...
+        {
+            Name = name,
+            Password = password
+        });
+
+    return migrationBuilder;
+}
+```
+
+Each provider can now decide what SQL to generate for this operation in their `IMigrationsSqlGenerator` service. Here is
```
ajcvickers: Maybe note here something like "this approach only works if every provider knows how to generate SQL for the operation." To align with the similar statement above about .Sql

entity-framework/core/managing-schemas/migrations/teams.md
```diff
@@ -0,0 +1,58 @@
...
+
+Resolving conflicts
+-------------------
+Sometimes you encounter a true conflict when merging the model snapshot model. For example, you and your teammate may
+each have renamed the same property.
+
+    <<<<<<< Mine
+    b.Property<bool>("Username");
+    =======
+    b.Property<int>("Alias");
```
ajcvickers: Shouldn't both properties be the same type in this example, since both are just renames?

bricelam: lol, thanks. Copy-and-paste error. I'm going to turn some of the code here (in in the tools docs) into samples.

entity-framework/core/managing-schemas/migrations/index.md
bricelam: Updated.

PR #569, opened 2017-11-20
Add note that EF 2.x is not supported

gldraphael:


entity-framework/core/providers/mysql/index.md
```diff
@@ -15,7 +15,7 @@ uid: core/providers/mysql/index
 This database provider allows Entity Framework Core to be used with MySQL. The provider is maintained as part of the [MySQL project](http://dev.mysql.com).
 
 > [!WARNING]  
-> This provider is pre-release.
+> This provider is pre-release. EF 2.x is currently [not supported](https://github.com/aspnet/EntityFrameworkCore/issues/10065).
```
divega: I agree it can be useful to have some notes here on the state of this provider, but if we are going to have a link, it should be to the MySQL documentation page where the up to date information should be available, e.g. https://dev.mysql.com/doc/connector-net/en/connector-net-versions.html.

gldraphael: Perhaps we could reword the statement to something like "Please review the MySQL documentation for version compatibility information [here](https://dev.mysql.com/doc/connector-net/en/connector-net-versions.html) and [here](https://dev.mysql.com/doc/connector-net/en/connector-net-entityframework-core.html)". Partly because we can then move it out of the warning box (below the supported database engines / supported platforms headings?), and partly because this page won't have to be updated to reflect version related changes in the future.

What do you think?

divega: Thanks @gldraphael. That sounds like a good idea to me. Want to give it a try?

gldraphael: I've made the change. Feel free to make any adjustments as you see fit :)

The other provider pages could possibly benefit from links like these as well.

divega:
Thanks @gldraphael. Sorry it took me so long.

PR #576, opened 2017-11-29
Added missing word (...cannot BE null...).

jeffa00:


divega:
Thanks @jeffa00 but there is already a verb in the sentence: contain. With your proposal we would have two verbs "contain be" and that would make the sentence incorrect.

> Current: A property whose CLR type cannot contain null 
> Proposed: A property whose CLR type cannot contain be null

PR #582, opened 2017-12-04
Added a note explaining TPH/TPT/TPC status

jv42:
Many people are looking for TPT or TPC, that way it is clear in the documentation, without having to search GitHub issues.


divega:
Thanks @jv42.

PR #584, opened 2017-12-10
Fix a syntax error in console parameter

miroslavpopovic:


divega:
Thanks @miroslavpopovic.

PR #586, opened 2017-12-12
FromSql should be given as composable SQL

gfoidl:
Fixes https://github.com/aspnet/EntityFramework.Docs/issues/585


divega:
Thanks @gfoidl, I am planning to add a bit more context around this, but adding this paragraph here is a good starting point.

PR #588, opened 2017-12-20
Improvements to configuring-DbContext

divega:
Adds more context around design-time discovery.

### DIFF
--- a/entity-framework/core/miscellaneous/cli/dbcontext-creation.md
+++ b/entity-framework/core/miscellaneous/cli/dbcontext-creation.md
@@ -8,7 +8,10 @@ uid: core/miscellaneous/cli/dbcontext-creation
 ---
 Design-time DbContext Creation
 ==============================
-Some of the EF Core Tools commands (for example, the [Migrations][3] commands) require a derived `DbContext` instance to be created at design time, in order to gather details about the application's entity types and how they map to a database schema. In most cases, it is desirable that the `DbContext` thereby created is [configured][4] in a similar way to how it would be configured at run time.
+Some of the EF Core Tools commands (for example, the [Migrations][1] commands) require a derived `DbContext` instance
+to be created at design time in order to gather details about the application's entity types and how they map to a
+database schema. In most cases, it is desirable that the `DbContext` thereby created is configured in a similar
+way to how it would be [configured at run time][2].
 
 There are various ways the tools try to create the `DbContext`:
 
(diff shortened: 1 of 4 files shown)


entity-framework/core/miscellaneous/cli/dbcontext-creation.md
```diff
@@ -4,30 +4,46 @@ author: bricelam
...
+default constructor if the `DbContext` is configured using the [`OnConfiguring`][6] method.
 
 From a design-time factory
 --------------------------
-You can also tell the tools how to create your DbContext by implementing `IDesignTimeDbContextFactory`. If a class
-implementing this interface is found inside your project, the tools bypass the other ways of creating the DbContext.
-They always use the factory at design time. A factory is especially useful if you need to configure the DbContext
-differently for design time than at runtime.
+You can also tell the tools how to create your DbContext by implementing the `IDesignTimeDbContextFactory<TContext>`
+interface. If a class implementing this interface is found in the same project as the derived `DbContext`, the tools
```
divega: @bricelam is there still a restriction that the factory needs to be in the same project/assembly as the dervived `DbContext` type? I copied this from the original text in configuring-dbcontext.

entity-framework/core/miscellaneous/cli/dbcontext-creation.md
```diff
@@ -49,8 +65,18 @@ namespace MyProject
...
+DI at all, or if for some reason you prefer not to have a `BuildWebHost` method in your ASP.NET Core application's  
+`Main` class.
+
+  [1]: xref:core/managing-schemas/migrations/index
+  [2]: xref:core/miscellaneous/configuring-dbcontext
+  [3]: https://docs.microsoft.com/aspnet/core/migration/1x-to-2x/#update-main-method-in-programcs
+  [4]: xref:core/miscellaneous/configuring-dbcontext#constructor-argument
+  [5]: xref:core/miscellaneous/configuring-dbcontext#using-dbcontext-with-dependency-injection
+  [6]: xref:core/miscellaneous/configuring-dbcontext#onconfiguring
+  [7]: https://github.com/aspnet/EntityFrameworkCore/issues/8332
```
divega: @bricelam I tried to keep the style used when you created this document. I think I succeeded, but feel free to point out anything I should change. BTW, as expected it was a real pain :smile:

divega:
@bricelam @ajcvickers I believe this is ready for review now.

entity-framework/core/miscellaneous/cli/dbcontext-creation.md
bricelam: It can either be in the startup assembly or in the same assembly as the `DbContext`. Most of the time, you'll probably want it in the startup assembly since that is typically where you set up your IoC container.

divega: Updated.

Issue #585, opened 2017-12-11
Document that FromSql should be given composable SQL


ajcvickers:
See https://github.com/aspnet/EntityFrameworkCore/issues/10474#issuecomment-349236822 and potentially mention that for SQL Server there should be no terminating semicolon.

WayPoint7:
After successfully following the guidelines and implementing the examples for my own database, the API runs on the localhost. But after upoloading to Azure I just get an HTTP 200 (OK) in return and an empty JSON file.
What could be wrong?

PR #590, opened 2017-12-21
More details on raw-sql composability paragraph

divega:
Additions to the fix for #585. 

cc @gfoidl


gfoidl:
@divega for me this looks clear and concise 👍

PR #592, opened 2017-12-26
Update cascade-delete.md

dnair926:
spelling correction


divega:
Thanks @dnair926 for catching this one.

PR #593, opened 2017-12-27
Fix casing of NuGet

Martin4ndersen:


divega:
Thanks @Martin4ndersen.

Issue #353, opened 2017-02-22
Add more 3rd party tools


ErikEJ:
LinqKit for Microsoft.EntityFrameworkCore with Include(..... By: sheyenrath 

Microsoft.EntityFrameworkCore.DynamicLinq By: sheyenrath 

Microsoft.EntityFrameworkCore.AutoHistory By: RigoFunc 

Microsoft.EntityFrameworkCore.UnitOfWork By: RigoFunc

bricelam:
[Devart Entity Developer](https://www.devart.com/entitydeveloper/)

rowanmiller:
@ErikEJ were you planning to send a PR for these?

ErikEJ:
Yes

rowanmiller:
Fixed by PR #359

axelheer:
How about [NeinLinq](https://github.com/axelheer/nein-linq)? 🤓

PR #600, opened 2018-01-09
Add NeinLinq.EntityFrameworkCore to Tools & Extensions

axelheer:
I'm not sure if there's some kind of process to add tools to the list (compare #353), but filing an issue asking for sending a PR feels a bit cumbersome. I even don't  know what qualifies a tool being added (hey, it's repo has > 100 stars) here...

Just let me know, if that works for you.


divega:
Thanks @axelheer. Accepted the PR.

PR #601, opened 2018-01-12
Add docs on configuring discriminator properties

ajcvickers:


entity-framework/core/modeling/relational/inheritance.md
```diff
@@ -84,3 +84,51 @@ public class RssBlog : Blog
     public string RssUrl { get; set; }
 }
 ```
+
+## Configuring the discriminator property
+
+In the examples above, the discriminator is created as a shadow property on the base entity of the hierarchy. Since it is a property in the model, it can be configured just like other properties. For example, to set the max length when the default, by-convention discriminator is being used:
```
divega: Can you add an xref to the shadow property topic?

PR #624, opened 2018-02-18
EF6 documentation migration 

divega:
This PR migrates the contents of the EF6 documentation from https://msdn.microsoft.com/en-us/library/aa937723(v=vs.113).aspx to https://docs.microsoft.com/ef/ef6.


divega:
Note to self: address feedback from https://github.com/aspnet/EntityFramework.Docs/pull/542 as well.

PR #634, opened 2018-02-27
Add docs for System.Transactions

### LINKED ISSUE #619: Update transactions documentation with System.Transaction support details
We can probably add a section to https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/saving/transactions.md. We should call out any limitations.

AndriySvyryd:
Fixes #619

### DIFF
--- a/entity-framework/core/saving/transactions.md
+++ b/entity-framework/core/saving/transactions.md
@@ -123,3 +123,12 @@ It is possible to use ambient transactions if you need to coordinate across a la
 It is also possible to enlist in an explicit transaction.
 
 [!code-csharp[Main](../../../samples/core/Saving/Saving/Transactions/CommitableTransaction/Sample.cs?name=Transaction&highlight=1,13,26,27,28)]
+
+### Limitations of System.Transactions  
+
+1. EF Core relies on database providers to implement support for System.Transactions. Although support is quite common among ADO.NET providers for .NET Framework, the API has only been recently added to .NET Core and hence support is not be as widespread. If a provider does not implement support for System.Transactions, it is possible that calls to these APIs will be completely ignored. SqlClient for .NET Core does support it from 2.1 onwards. SqlClient for .NET Core 2.0 will throw an exception of you attempt to use the feature. 
+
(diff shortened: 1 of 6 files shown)


samples/core/Saving/Saving/Transactions/AmbientTransaction/Sample.cs
```diff
@@ -0,0 +1,74 @@
...
+                new DbContextOptionsBuilder<BloggingContext>()
+                    .UseSqlServer(connectionString)
+                    .Options))
+            {
+                context.Database.EnsureDeleted();
+                context.Database.EnsureCreated();
+            }
+
+            #region Transaction
+            using (var scope = new TransactionScope(TransactionScopeOption.Required))
```
divega: Mind if I change the isolation level? The default isn't great. See @davidbaxterbrowne's write up for reference: https://blogs.msdn.microsoft.com/dbrowne/2010/06/03/using-new-transactionscope-considered-harmful/

entity-framework/core/saving/transactions.md
```diff
@@ -111,81 +101,25 @@ public class BloggingContext : DbContext
...
-                // Commit transaction if all commands succeed, transaction will auto-rollback
-                // when disposed if either commands fails
-                transaction.Commit();
-            }
-            catch (System.Exception)
-            {
-                // TODO: Handle failure
-            }
-```
+[!code-csharp[Main](../../../samples/core/Saving/Saving/Transactions/CommitableTransaction/Sample.cs?name=Transaction&highlight=1,13,26,27,28)]
```
divega: I would like to add a note explaining that as of version 2.1, .NET Core does not support using System.Transactions to coordinate transactions distributed across multiple resource managers. Sounds good?

AndriySvyryd: There are probably some other limitations, but I can't recall now.
For instance we could mention that non-relational providers might not support this.

divega: Fair enough. We should said that it is up to the provider in general.

PR #469, opened 2017-09-03
Owned Entity page for EF Core docs

julielerman:
I combined my own simpler write up  with what was in the EF Core Whats new in 2.0 info about owned entities to go in the main portion of the EF Core Modeling documentaiton

PR #636, opened 2018-02-27
Owned entity types documentation reborn

### LINKED ISSUE #466: New feature topic: Documentation for owned types
Copied from https://github.com/aspnet/EntityFrameworkCore/issues/9620 by @FransBouma

Hi,

For LLBLGen Pro v5.3 I'm updating our EF Core support to v2.0 and as our designer supports table splitting and value types / complex types (DDD), we can support table splitting and owned types for EF Core 2.0. However there appears to be no documentation (that I could find) about owned types mapping for code first.

For table splitting there also appears to be no documentation, e.g. what the limitations are, and how to set it up. I can assume a few things, e.g. that the split off entity can be mapped the same way as the root entity but if there are limitations, e.g. fields have to be nullable in the DB, navigational properties are/aren't supported, that's unknown, as there's no documentation at all about this.

For owned type mappings there's also no documentation: as they appear to have no PK, I can't use regular entity mappings (I think, but again: no idea) so how to set it up is unclear. I know there's 1 single example in the 'what's new' page in the documentation but that maps everything with conventions which is unusable for us as we need to cover all scenarios, e.g. a field is renamed and we need to explicitly specify the mapping. Also with this, there is no description about what the limitations are, e.g. whether the owned type can or can't be null, whether the fields have to be null or don't have to be...

TIA

divega:
This is based on @julielerman's PR #469, but incorporating changes from @AndriySvyryd and my own feedback, changes in 2.1, etc.

Fixes #466.

### DIFF
--- a/entity-framework/core/modeling/owned-entities.md
+++ b/entity-framework/core/modeling/owned-entities.md
@@ -1,20 +1,26 @@
 ---
-title: EF Core | Owned Entity Types | Microsoft Docs
+title: Owned Entity Types - EF Core
 author: julielerman
-
+ms.author: divega
+ms.date: 2/26/2018
+ms.assetid: 2B0BADCE-E23E-4B28-B8EE-537883E16DF3
+ms.technology: entity-framework-core
 uid: core/modeling/owned-entities
 ---
 # Owned Entity Types
 
-This feature was added to EF Core 2.0.
+>[!NOTE]
+> This feature is new in EF Core 2.0.
 
-Owned entity type allow you to map types that do not have their own identity and are used as properties in any of your entities (such as a value object). An owned entity type shares the same CLR type with another entity type. The entity containing the defining navigation is the owner. When querying the owner the owned types will be included by default.
+EF Core allows you to model entity types that can only appear on navigation properties in another entity type. These are called _owned entity types_. The entity containing an owned entity type is its _owner_.
 
-By convention a shadow primary key will be created for the owned type and it will be mapped to the same table as the owner by using table splitting. This allows to use owned types similarly to how complex types are used in EF6. 
+## Explicit configuration
 
-The OwnsOne fluent mapping alerts the DbContext to this mapping. In this example, StreetAddress is a type with no identity property. It is used as a property of the Order type to specify the shipping address for a particular order. In the DbContext, you can use the Entity().OwnsOne() method to specify that the ShippingAddress property is an Owned Entity of the Order type.
+Owned entity types are never included by EF Core in the model by convention. You can use the `OwnsOne` method in `OnModelCreating` or annotate the type with `OwnedAttrbibute` (new in EF Core 2.1) to configure the type as an owned type.
 
-```
+In this example, StreetAddress is a type with no identity property. It is used as a property of the Order type to specify the shipping address for a particular order. In `OnModelCreating`, we use the `OwnsOne` method to specify that the ShippingAddress property is an Owned Entity of the Order type.
+
+``` csharp
 public class StreetAddress
 {
     public string Street { get; set; }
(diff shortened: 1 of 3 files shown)


entity-framework/core/modeling/owned-entities.md
```diff
@@ -0,0 +1,188 @@
...
+Owned types stored with table splitting can be used very similarly to how complex types are used in EF6.
+
+By contention, EF Core will name the database columns for the properties of the owned entity type following the pattern _EntityProperty_OwnedEntityProperty_. Therefore the StreetAddress properties will appear in the Orders table with the names ShippingAddress_Street and ShippingAddress_City.
+
+You can append the `HasColumnName` method to rename those columns. In the case where StreetAddress is a public property, the mappings would be
+
+``` csharp
+modelBuilder.Entity<Order>().OwnsOne(p => p.ShippingAddress)
+                            .Property(p=>p.Street).HasColumnName("ShipsToStreet");
+modelBuilder.Entity<Order>().OwnsOne(p => p.ShippingAddress)
```
ajcvickers: Do we not have nested closure syntax for this?

entity-framework/core/modeling/owned-entities.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Limitations
+
+Here are some limitations of owned entity types. Some of these limitations are fundamental to how owned types work, but some others are point-in-time restrictions that we would like to remove in future releases:
+
+### Current shortcomings
+- Inheritance of owned types is not supported
+- Owned types cannot be pointed at by a collection navigation property
+- Since they use table splitting by default owned types also have the following restrictions unless explicitly mapped to a different table:
+   - They cannot be owned by a derived type
```
AndriySvyryd: This is no longer true in 2.1

entity-framework/core/modeling/owned-entities.md
```diff
@@ -0,0 +1,191 @@
...
+# Owned Entity Types
+
+>[!NOTE]
+> This feature is new in EF Core 2.0.
+
+EF Core allows you to model entity types that can only ever appear on navigation properties of other entity types. These are called _owned entity types_. The entity containing an owned entity type is its _owner_.
+
+## Explicit configuration
+
+Owned entity types are never included by EF Core in the model by convention. You can use the `OwnsOne` method in `OnModelCreating` or annotate the type with `OwnedAttrbibute` (new in EF Core 2.1) to configure the type as an owned type.
```
AndriySvyryd: One can also use `Owned<>()`

entity-framework/core/modeling/owned-entities.md
```diff
@@ -0,0 +1,191 @@
...
+public class Order
+{
+    public int Id { get; set; }
+    public StreetAddress ShippingAddress { get; set; }
+}
+```
+
+## Implicit keys
+
+In EF Core 2.0 and 2.1, only reference navigation properties can point to owned types. Collections of owned types are not supported. These reference owned types always have a one-to-one relationship with the owner, therefore they don't need their own key values. In the previous example, the StreetAddress type does need to define a key property.
```
AndriySvyryd: Did you mean 'the StreetAddress type __doesn't__ need to define a key property.'?

entity-framework/core/modeling/owned-entities.md
```diff
@@ -0,0 +1,191 @@
...
+
+In order to understanding how EF Core tracks these objects, it is useful to think that a primary key is created as a [shadow property](xref:core/modeling/shadow-properties) for the owned type. The value of the key of an instance of the owned type will be the same as the value of the key of the owner instance.      
+
+## Mapping owned types with table splitting
+
+When using relational databases, by convention owned types are mapped to the same table as the owner. This requires splitting the table in two: some columns will be used to store the data of the owner, and some columns will be used to store data of the owned entity. This is a common feature known as table splitting.
+
+Owned types stored with table splitting can be used very similarly to how complex types are used in EF6.
+
+By contention, EF Core will name the database columns for the properties of the owned entity type following the pattern _EntityProperty_OwnedEntityProperty_. Therefore the StreetAddress properties will appear in the Orders table with the names ShippingAddress_Street and ShippingAddress_City.
```
AndriySvyryd: Typo in 'By __convention__'

AndriySvyryd: The pattern is more like _OwnedNavigation_OwnedEntityProperty_.

entity-framework/core/modeling/owned-entities.md
```diff
@@ -0,0 +1,191 @@
...
+modelBuilder.Entity<Order>().OwnsOne(p => p.OrderDetails, od =>
+    {
+        od.OwnsOne(c => c.BillingAddress);
+        od.OwnsOne(c => c.ShippingAddress);
+    });
+```
+
+Of course, it would be possible to achieve the same thing using `OwnedAttribute`.
+
+In addition to nested owned types, an owned type can reference a regular entity. In the following example, Country is a regular (i.e. non-owned) entity:
```
AndriySvyryd: As long as the owned type is the dependent

entity-framework/core/modeling/owned-entities.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+## Querying owned types
+
+When querying the owner the owned types will be included by default. It is not necessary to use the `Include` method, even if the owned types are stored in a separate table. Based on the model described before, the following query will pull Order, OrderDetails and the two owned StreeAddresses for all pending orders:
+
+``` csharp
+var orders = context.Orders.Where(o => o.Status == OrderStatus.Pending);
+```  
+
```
AndriySvyryd: It might be worth also showing how to handle the owned entity state:
```C#
var ownedEntry = context.Entry(order).Reference(e => e.BillingAddress).TargetEntry;
ownedEntry.State = EntityState.Unchanged;
ownedEntry.Property(e => e.Street).IsModified = true;
```

PR #641, opened 2018-02-28
Fix Read-only properties description

KrzysztofBranicki:
I think you meant "setters" in first point of Read-only properties section.


divega:
@KrzysztofBranicki I believe you are right. Cc @ajcvickers

PR #649, opened 2018-03-07
Update guidance on cross-taregting

ajcvickers:


AndriySvyryd:
Could you fix https://github.com/aspnet/EntityFramework.Docs/issues/581 as well by linking to this from https://docs.microsoft.com/en-us/ef/core/get-started/uwp/getting-started instead of cross-targeting

ajcvickers:
@AndriySvyryd This is a little more work than I want to do right now because I think we also need to actually add the steps for creating the console application, and it would be good if the steps are actually tested.

PR #650, opened 2018-03-07
Adding Jiri's provider for Firebird

divega:
Reviewing may require looking at this with a very small font or two screens.

@cincuranet can you please verify that the information is accurate? Also, is there a better link than that blog post I found?


cincuranet:
Yes, that's correct. 

At the moment I don't have any page better than the blog post (because it's not RTM yet), but The RTM should be available soon and then I will prepare some landing page (I'll do PR).

PR #652, opened 2018-03-14
Typo fix

Sojaner:
Changed the word "that" to "than" in the following sentence:
(for query types such a property is typed as `DbQuery<T>` rather that `DbSet<T>`)


divega:
Thanks @Sojaner!

PR #664, opened 2018-03-26
Update dbcontext-creation.md

marekr:
IDesignTimeDbContextFactory is located in the Microsoft.EntityFrameworkCore.Design namespace and thus the using directive is required.


divega:
Thank you @marekr!

PR #678, opened 2018-04-12
Typo fix

austindrenski:
call -> called


divega:
Thanks @austindrenski!

PR #682, opened 2018-04-17
React to rename of ToToable API to ToView for query types

divega:
Updating the conceptual topic to reflect the new API name.

Also minor markdown simplification.

I will send a separate PR to the EFCore repo to update the code snippets :unamused:

cc @anpete @ajcvickers @smitpatel


entity-framework/core/modeling/query-types.md
```diff
@@ -15,26 +15,27 @@ Query Types are read-only query result types that can be added to the EF Core mo
 
 They are conceptually similar to Entity Types in that:
 
-- They are POCO C# types that are added to the model, either in ```OnModelCreating``` using the ```ModelBuilder.Query``` method, or via a DbContext "set" property (for query types such a property is typed as ```DbQuery<T>``` rather than ```DbSet<T>```).
-- They support much of the same mapping capabilities as regular entity types. For example, inheritance mapping, navigations (see limitiations below) and, on relational stores, the ability to configure the target database schema objects via ```ToTable```, ```HasColumn``` fluent-api methods (or data annotations).
+- They are simple classes that are added to the model, either in `OnModelCreating` using the `ModelBuilder.Query` method, or via a DbContext "set" property (for query types such a property is typed as `DbQuery<T>` rather than `DbSet<T>`).
```
anpete: What does "simple class" mean? Are we moving away from talking about "POCOs"?

entity-framework/core/modeling/query-types.md
```diff
@@ -15,26 +15,27 @@ Query Types are read-only query result types that can be added to the EF Core mo
...
-- Serving as the return type for ad hoc ```FromSql()``` queries.
+- Serving as the return type for ad hoc `FromSql()` queries.
 - Mapping to queries defined in the model.
 
 > [!TIP]
-> Mapping a query type to a database view is achieved using the ```ToTable``` fluent API.
+> Mapping a query type to a database object is achieved using the `ToView` fluent API. From the perspective of EF Core, the database object specified in this method is a _view_, which simply means that it is treated as a read-only query source and cannot not be the target of update, insert or delete operations. However, this does not mean that the database object is required to be database view - It can alternatively be a database table that needs to be treated as read-only. Conversely, for entity types, EF Core assumes that a database object specified in the `ToTable` method can be treated as a _table_, meaning that it can be used as a query source but also targeted by update, delete and insert operations. In fact, you can specify the name of a database view in `ToTable` and everything should work as long as the view is configured on the database to be updatable.
```
divega: Trying to get this explanation very accurate but not too complicated. Feedback welcome @anpete @ajcvickers @smitpatel.

entity-framework/core/modeling/query-types.md
```diff
@@ -15,26 +15,27 @@ Query Types are read-only query result types that can be added to the EF Core mo
 
 They are conceptually similar to Entity Types in that:
 
-- They are POCO C# types that are added to the model, either in ```OnModelCreating``` using the ```ModelBuilder.Query``` method, or via a DbContext "set" property (for query types such a property is typed as ```DbQuery<T>``` rather than ```DbSet<T>```).
-- They support much of the same mapping capabilities as regular entity types. For example, inheritance mapping, navigations (see limitiations below) and, on relational stores, the ability to configure the target database schema objects via ```ToTable```, ```HasColumn``` fluent-api methods (or data annotations).
+- They are simple classes that are added to the model, either in `OnModelCreating` using the `ModelBuilder.Query` method, or via a DbContext "set" property (for query types such a property is typed as `DbQuery<T>` rather than `DbSet<T>`).
+- They support much of the same mapping capabilities as regular entity types. For example, inheritance mapping, navigations (see limitations below) and, on relational stores, the ability to configure the target database objects via fluent API methods (or data annotations).
```
anpete: "They support many", sounds better, I think..

entity-framework/core/modeling/query-types.md
anpete: "cannot not" 😄

anpete: "be database view"

anpete: that needs to be treated as read-only -> Consider "that will be treated as read-only"

entity-framework/core/modeling/query-types.md
divega: I am happy that our customers can take "persistence ignorance" for granted (at least to the degree that our mapping is flexible and keeps become more flexible). 

I took "POCO" out because I believe that:
- For too many customers it conveys a class with a parameterless constructor and public automatic properties, and I refuse to feed that misconception :smile: 
- There are also customers out there who missed the old debate and for which the term means nothing.

It took "C# type" because I believe .NET class is a much simpler and slightly more accurate way to refer to what it is, but .NET is also redundant so I avoided it.

I put simple there trying to preserve some of the meaning, but now that I think more about it, I am inclined to remove it. If I had to qualify "class" with anything, I would probalby use "data class". What do you think?

PR #687, opened 2018-04-20
Add EnsureCreated() details

mattferderer:
When attempting to learn from this sample, I was unaware that EnsureCreated() needed to be called. Better clarification would be of value.

### DIFF
--- a/entity-framework/core/modeling/data-seeding.md
+++ b/entity-framework/core/modeling/data-seeding.md
@@ -15,7 +15,7 @@ uid: core/modeling/data-seeding
 > [!NOTE]  
 > This feature is new in EF Core 2.1.
 
-Data seeding allows to provide initial data to populate a database. Unlike in EF6, in EF Core, seeding data is associated with an entity type as part of the model configuration. Then EF Core migrations can automatically compute what insert, update or delete operations need to be applied when upgrading the database to a new version of the model.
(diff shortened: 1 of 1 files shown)


divega:
The first paragraph already says:

> when upgrading the database to a new version of the model.

But we can be more explicit about it. I created https://github.com/aspnet/EntityFramework.Docs/issues/690 for this. 

In most cases, migrations is a much more appropriate way to achieve that than invoking EnsureCreated. 

The sample console app that hosts the code snippets happens to use EnsureCreated for simplicity (that is, it requires zero additional steps this way, and it works with in-memory in Preview 2), but we shouldn't turn this into something that can be read as a recommendation to use EnsureCreated. 

Would you like to improve your PR to reflect this?

mattferderer:
That makes sense @divega. What are your thoughts on an edit that does the following:

* Mention to use [Migrations](https://docs.microsoft.com/en-us/aspnet/core/data/ef-mvc/migrations?view=aspnetcore-2.1#introduction-to-migrations) to seed the database.
* Mention EnsureCreated for when needing to seed an in-memory database.

mattferderer:
Kept going back & forth on this debating if changes were even necessary. After reading it closer, it seemed as if I missed the obvious statement at the top to use migrations. I added a short paragraph at the end recommending them again, since that would be the next logical step after creating your data seed entities. I also made the mention earlier a link to help make it more noticeable. After re-reading your thoughts I felt it best to leave out EnsureCreated all together.

divega:
Thanks @mattferderer! The proposed change looks good. I may add something afterward mentioning that you can alternatively use EnsureCreated, e.g. if you are using an in-memory database.

PR #693, opened 2018-05-01
Fixing typo, PostgreSQL vs. PostreSQL

mungojam:


divega:
Thanks @mungojam! I guess @bricelam's Spanish speaking unconscious mind has a sweet tooth :smile:

PR #695, opened 2018-05-10
transaction rollback on exception

SimonCropp:
as per https://msdn.microsoft.com/en-us/library/zayx5s0h(v=vs.110).aspx

also reduced the indent of the using nesting since it makes it more readable with less horizontal width.  and reduces the chance of wrap/scroll on smaller devices

### DIFF
--- a/samples/core/Miscellaneous/ConnectionResiliency/Program.cs
+++ b/samples/core/Miscellaneous/ConnectionResiliency/Program.cs
@@ -31,8 +31,9 @@ namespace ConnectionResiliency
                 strategy.Execute(() =>
                 {
                     using (var context = new BloggingContext())
+                    using (var transaction = context.Database.BeginTransaction())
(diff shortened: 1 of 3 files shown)


divega:
@SimonCropp we actually made our sample code simpler than the one at https://msdn.microsoft.com/en-us/library/zayx5s0h(v=vs.110).aspx on purpose. 

If an error occurs on the database or if the transaction is disposed before it is committed, the transaction will be rolled back automatically, so it is really hard for me at the moment to come up with a scenario in which the extra lines of code could provide real value. 

But I may be missing something. Is there anything else you are trying to achieve besides alignign with existing example?

divega:
@AndriySvyryd thoughts on this?

SimonCropp:
@divega thanks for the clarification. in that case i can happily close this PR

PR #698, opened 2018-05-11
Update query types topic to clarify navigations

divega:
See https://github.com/aspnet/EntityFrameworkCore/issues/11971 for some context on what needs clarification.


divega:
cc @julielerman

PR #722, opened 2018-05-25
I can grammar!

bricelam:


entity-framework/core/managing-schemas/migrations/index.md
```diff
@@ -190,8 +190,8 @@ myDbContext.Database.Migrate();
 ```
 
 > [!WARNING]
-> Don't call `EnsureCreated()` before `Migrate()`. `EnsureCreated()` bypasses Migrations to create the schema and
-> cause `Migrate()` to fail.
+> Don't call `EnsureCreated()` before `Migrate()`. `EnsureCreated()` bypasses Migrations to create the schema which
```
divega: I can’t grammar :) but what do you think about a comma before which?

PR #753, opened 2018-06-10
Update index.md

msawczyn:
Added tool entry for Entity Framework Visual Editor (https://marketplace.visualstudio.com/items?itemName=michaelsawczyn.EFDesigner)

### DIFF
--- a/entity-framework/core/extensions/index.md
+++ b/entity-framework/core/extensions/index.md
@@ -35,6 +35,12 @@ Visual Studio 2017+ extension. You can reverse engineer of DbContext and POCO cl
 
 [GitHub wiki](https://github.com/ErikEJ/SqlCeToolbox/wiki/EF-Core-Power-Tools)
 
+### Entity Framework Visual Editor
+
+A Visual Studio 2017 extension that adds an ORM designer for visual design of Entity Framework 6, Core 2.0 and Core 2.1 classes. Code is generated using T4 templates so can be completely customized to suit any needs. Inheritance, unidirectional and bidirectional associations are all supported, as are enumerations and the ability to color-code your classes and add text blocks to explain potentially arcane parts of your design.
+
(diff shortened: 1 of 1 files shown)


divega:
@msawczyn Thanks for this PR! How do I specify that I want the code generated to be for EF Core? I am getting EF6 classes generated instead. When I try to change the "Entity Framework Version" property on the file to EF Core, I get a message dialog saying "EF Core currently only supports Table-Per-Hierarchy inheritance strategy". When I close the dialog, the property value goes back to EF6. I should clarify that the model only contains one entity (so no inheritance).

divega:
@msawczyn friendly ping... I would love to merge this but first I want to be sure I understand how to make it work with EF Core.

divega:
I tried this again today and I was able to switch to EF Core code generation before I started adding entities to the model.

msawczyn:
@divega -- I just saw this. No clue why github didn't drop me an email when you did the mention back in June, but my apologies nonetheless. I'm really not the uncommunicative fool you probably think I am \<g\>.

The message you initially got was one of those "as-designed" behaviors ... as-designed, but unclear. You had an inheritance strategy set that was something other than table-per-hierarchy, so when you tried to toggle over to EFCore, it complained that it couldn't do that with the indicated inheritance strategy.

Only the message really wasn't clear, and you're not the first one to bring it up.

So I made a change a few versions back that, since only one inheritance strategy is supported, simply changed it when you ticked the model over to EFCore. Kind of felt a bit heavy handed to do that but, hey, it's all you can do anyway, right?

When other inheritance types are supported, I'll make the change that stops that behavior, but it does seem to have made folks a bit less confused, so I guess it was the right thing to do.

Thanks for the mention, and sorry again that I didn't see your comment back when you had made it.

divega:
@msawczyn no worries :smile: Thanks again for the PR and great to learn about your tool!

PR #758, opened 2018-06-11
Update Transactions sample highlighting

AndriySvyryd:


entity-framework/core/saving/transactions.md
```diff
@@ -31,32 +31,32 @@ Not all database providers support transactions. Some providers may throw or no-
 
 <!-- [!code-csharp[Main](samples/core/Saving/Saving/Transactions/ControllingTransaction/Sample.cs?highlight=3,17,18,19)] -->
 ``` csharp
-        using (var context = new BloggingContext())
+using (var context = new BloggingContext())
```
divega: Can you just update ControllingTransaction/Sample.cs and remove this from here? :smile:

PR #759, opened 2018-06-11
Add more information on data seeding

### LINKED ISSUE #669: Data seeding docs should mention requirement to provide key values even if the key is store-generated
https://github.com/aspnet/EntityFrameworkCore/issues/11503 gives us a reason to believe that this could help.

### LINKED ISSUE #710: Document how to seed owned types
See https://github.com/aspnet/EntityFrameworkCore/issues/12004#issuecomment-389715911

### LINKED ISSUE #736: How to seed the ASP.NET Identity model, and related seeding questions
# Updated to reflect the core issue we are tracking here:
TL;DR: It is very hard to come up with really good guidance on how create or update application state on deployment. Note that this is commonly related to database schema and seed data, but it is not limited to that.

Many customers end up adding logic at application startup because that runs after deployment, but this is very problematic, especially if there is more than one instance of the application using the same database. 

We think this ideally should happen as a post-deployment step, not startup. 

We would like to do some brainstorming with a wider group with the goal of identifying patterns that we can recommend today and/or functionality that we can build into the publish components to make this work. 

# Original issue as reported:
How can the key of an entity be specified? I have an entity that should be seeded based on its name and the ID (Guid) is generated automatically. Something like this:

```
modelBuilder.Entity<Blog>().HasData(blog => blog.Name, new Blog {Name = "New Blog", Url = "http://sample.com"});
```

And for composite keys:
```
modelBuilder.Entity<Blog>().HasData(blog => new {blog.Name, blog.Url}, new Blog {Name = "New Blog", Url = "http://sample.com"});
```

---
#### Document Details

* ID: fc418ab9-0592-f610-e751-40cb501220da
* Version Independent ID: 822e8923-2647-f573-0c74-66af903776d8
* Content: [Data Seeding - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/data-seeding)
* Content Source: [entity-framework/core/modeling/data-seeding.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/data-seeding.md)
* GitHub Login: @AndriySvyryd
* Microsoft Alias: **divega**

### LINKED ISSUE #925: An Example with EnsureCreated May Help
A few suggestions:
1) It wasn't apparent to me that if I'm not using migrations, I *need* to call context.Database.EnsureCreated().  I didn't get any errors when compiling and running my code but my seed data was not appearing in the InMemory database.  It's probably helpful to note that you won't see seed data unless this call is made.
2) An example with EnsureCreated may be very helpful for those new to EF Core 2.1.  I didn't have any guidance as to where I should add this call.  I initially added this in OnModelCreating of my DbContext class which seemed to work initially but I was getting duplicate key errors later.  I later moved this to the Configure method of my Startup.cs class which seemed to work better.  A short example may go a long way.
3) An example on how to seed Identity data could also help greatly.

---
#### Document Details

* ID: fc418ab9-0592-f610-e751-40cb501220da
* Version Independent ID: 822e8923-2647-f573-0c74-66af903776d8
* Content: [Data Seeding - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/data-seeding)
* Content Source: [entity-framework/core/modeling/data-seeding.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/data-seeding.md)
* GitHub Login: @AndriySvyryd
* Microsoft Alias: **divega**

### LINKED ISSUE #1055: Where HasData fails
`HasData` relies completely on migrations, so if you already had data in that table before adding `HasData` migration next time you will try to migrate an exception will be thrown. There is no way to check if data already exists in the table you want to seed. I think that seeding using the old way (without `HasData`) inside `using (var serviceScope = webHost.Services.CreateScope())` is more advantageous because you can check if data exists before seeding, not if migration was applied.

So Imagine we have 2 migrations, first migration creates the table, some data was added to that new table in the meantime, after some time we decide we want to seed the table with some default rows but the problem is there are already some rows in the table and we can't protect ourselves against it. Just 1 problem showing that manual seeding like before EF Core 2.1 is superior to `HasData`.

---
#### Document Details

* ID: fc418ab9-0592-f610-e751-40cb501220da
* Version Independent ID: 822e8923-2647-f573-0c74-66af903776d8
* Content: [Data Seeding - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/data-seeding)
* Content Source: [entity-framework/core/modeling/data-seeding.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/data-seeding.md)
* GitHub Login: @AndriySvyryd
* Microsoft Alias: **divega**

### LINKED ISSUE #1066: When using HasData, is there a way to update/migrate and not seed?
I see that the seed data is now baked into my migrations, which is nice. But I only want to seed when integration testing via SQLite. Is there a flag or could we add one to skip over statements like `migrationBuilder.InsertData`?

---
#### Document Details

* ID: 6e28b0a8-b7e4-4c07-9e60-0df403bbee28
* Version Independent ID: 12300b31-6f7b-b75c-970d-693365e77013
* Content: [Migrations - EF Core](https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/index)
* Content Source: [entity-framework/core/managing-schemas/migrations/index.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/managing-schemas/migrations/index.md)
* GitHub Login: @bricelam
* Microsoft Alias: **bricelam**

### LINKED ISSUE #1081: How do I insert foreign key related data?
when i execute below line, 
 modelBuilder.Entity&lt;DeviceTypeFamily&gt;().HasData(new DeviceTypeFamily {Name = "ABC" });

This line will produce auto generated guid, now this shall be the input for new table. How can we do it?

---
#### Document Details

* ID: fc418ab9-0592-f610-e751-40cb501220da
* Version Independent ID: 822e8923-2647-f573-0c74-66af903776d8
* Content: [Data Seeding - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/data-seeding)
* Content Source: [entity-framework/core/modeling/data-seeding.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/data-seeding.md)
* GitHub Login: @AndriySvyryd
* Microsoft Alias: **divega**

AndriySvyryd:
Fixes #669
Fixes #710
Fixes #736
Fixes #925
Fixes #1055
Fixes #1066
Fixes #1081
Part of #882


entity-framework/core/modeling/data-seeding.md
```diff
@@ -28,3 +28,32 @@ To add entities that have a relationship the foreign key values need to be speci
...
 
 Alternatively, you can use `context.Database.EnsureCreated()` to create a new database containing the seed data, for example for a test database or when using the in-memory provider. Note that if the database already exists, `EnsureCreated()` will neither update the schema nor the seed data in the database.
+
+This type of seed data is managed by migrations and the script to update the data that's already in the database needs to be generated without connecting to the database. This imposes some restrictions:
+* The primary key value needs to be specified even if it's usually generated by the database
+* Any changes to the seed data done outside of migrations will be lost if the primary key is changed in any way
+
+Therefore this feature is most useful for static data that's not expected to change outside of migrations and doesn't depend on anything else in the database.
+
+If you need to seed temporary data for testing, complex data that depends on database state or data that uses alternate keys as identity it is recommended to add the seeding logic to the application start up method.
```
divega: Why turn this into a recommendation? It seems there are some of the same potentially concurrency issues as in applying migrations at startup.

If you need to setup seed data for your application and for some reason HasData didn't work for you, you can do it in code in your application, but it doesn't need to be at startup. It can be in some controller action available to the admin of the application instead. 

If you need to setup sample data for a test, you can do it as part of the test's setup.

AndriySvyryd: This was the smallest sample I could think of to illustrate the workaround. Do you think we should provide a sample controller action instead?

divega: Possibly. But let's discuss the whole issue with @ajcvickers. I have the impression I am missing something.

AndriySvyryd: We should mention that doing migrations and or seeding is not recommended in a running application as it could cause concurrency and permission issues.
Ideally this sort of operations should be performed during deployment in some way. And migrations seed data helps doing it is some scenarios.

AndriySvyryd:
@divega Any reason this is still blocked?

divega:
@AndriySvyryd, I still have the same concerns but it is possible than adding this with explicit guidance on all the scenarios in which you should not do it is better than not adding it. 

Let's get together with @ajcvickers and discuss it once more, when there is a chance.

AndriySvyryd:
@divega Updated

entity-framework/core/modeling/data-seeding.md
```diff
@@ -7,19 +7,74 @@ uid: core/modeling/data-seeding
...
+
+Therefore this feature is most useful for static data that's not expected to change outside of migrations and does not depend on anything else in the database, for example ZIP codes.
+
+If your scenario includes any of the following it is recommended to use custom initialization logic described in the next section:
+* Temporary data for testing
+* Data that depends on database state
+* Data that uses alternate keys as identity
+* Data that needs key values to be generated by the database
+* Data that requires custom transformation (that is not handled by [value conversions](xref:core/modeling/value-conversions)), such as some password hashing
+* Data that requires calls to external API, such as ASP.NET Core Identity roles and users creation
```
divega: Maybe the two above are just data that can be generated or computed only at runtime?

entity-framework/core/modeling/data-seeding.md
```diff
@@ -7,19 +7,74 @@ uid: core/modeling/data-seeding
...
+
+See the [full sample project](https://github.com/aspnet/EntityFramework.Docs/tree/master/samples/core/DataSeeding) for more context.
+
+Once the data has been added to the model, [migrations](xref:core/managing-schemas/migrations/index) should be used to apply the changes.
+
+> [!TIP]
+> If you need to apply migrations as part of an automated deployment you can [create a SQL script](xref:core/managing-schemas/migrations/index#generate-sql-scripts) that can be previewed before execution.
 
 Alternatively, you can use `context.Database.EnsureCreated()` to create a new database containing the seed data, for example for a test database or when using the in-memory provider. Note that if the database already exists, `EnsureCreated()` will neither update the schema nor the seed data in the database.
+
```
divega: I think we should mention here that:
- If your database provider doesn't support migrations, then EnsureCreated() is the way to go
- If your database provider supports migration, you should not use EnsureCreated() if you want to use Migrations.

entity-framework/core/modeling/data-seeding.md
```diff
@@ -7,19 +7,74 @@ uid: core/modeling/data-seeding
...
+This type of seed data is managed by migrations and the script to update the data that's already in the database needs to be generated without connecting to the database. This imposes some restrictions:
+* The primary key value needs to be specified even if it's usually generated by the database. It will be used to detect data changes between migrations.
+* Previously seeded data will be removed if the primary key is changed in any way.
+
+Therefore this feature is most useful for static data that's not expected to change outside of migrations and does not depend on anything else in the database, for example ZIP codes.
+
+If your scenario includes any of the following it is recommended to use custom initialization logic described in the next section:
+* Temporary data for testing
+* Data that depends on database state
+* Data that uses alternate keys as identity
```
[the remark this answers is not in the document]
AndriySvyryd: Same as the next one - you still need to provide the PK value in addition to the AK

entity-framework/core/modeling/data-seeding.md
```diff
@@ -7,19 +7,74 @@ uid: core/modeling/data-seeding
...
 
 Alternatively, you can use `context.Database.EnsureCreated()` to create a new database containing the seed data, for example for a test database or when using the in-memory provider. Note that if the database already exists, `EnsureCreated()` will neither update the schema nor the seed data in the database.
+
+This type of seed data is managed by migrations and the script to update the data that's already in the database needs to be generated without connecting to the database. This imposes some restrictions:
+* The primary key value needs to be specified even if it's usually generated by the database. It will be used to detect data changes between migrations.
+* Previously seeded data will be removed if the primary key is changed in any way.
+
+Therefore this feature is most useful for static data that's not expected to change outside of migrations and does not depend on anything else in the database, for example ZIP codes.
+
+If your scenario includes any of the following it is recommended to use custom initialization logic described in the next section:
```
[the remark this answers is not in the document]
AndriySvyryd: The value of `HasData` is that it makes it simpler to add and manage seed data. If you need to do any manipulation outside of it that value is lost and you might as well not use it to avoid having to thin through the interaction between both approaches.

entity-framework/core/modeling/data-seeding.md
```diff
@@ -7,19 +7,74 @@ uid: core/modeling/data-seeding
...
+
+> [!NOTE]
 > This feature is new in EF Core 2.1.
 
-Data seeding allows to provide initial data to populate a database. Unlike in EF6, in EF Core, seeding data is associated with an entity type as part of the model configuration. Then EF Core [migrations](xref:core/managing-schemas/migrations/index) can automatically compute what insert, update or delete operations need to be applied when upgrading the database to a new version of the model.
+Unlike in EF6, in EF Core, seeding data can be associated with an entity type as part of the model configuration. Then EF Core [migrations](xref:core/managing-schemas/migrations/index) can automatically compute what insert, update or delete operations need to be applied when upgrading the database to a new version of the model.
 
-As an example, you can use this to configure seed data for a `Blog` in `OnModelCreating`:
+> [!NOTE]
+> Migrations only considers model changes when determining what operation should be performed to get the seed data into the desired state. Thus any changes to the data performed outside of migrations might be lost or cause an error.
```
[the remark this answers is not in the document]
AndriySvyryd: Yes

entity-framework/core/modeling/data-seeding.md
AndriySvyryd: This list has a lot of redundancy on purpose to make it more scannable

entity-framework/core/modeling/data-seeding.md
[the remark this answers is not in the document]
divega: Swap the order? Or write one sentence that explicitly applies to both primary keys and alternate key values?

entity-framework/core/modeling/data-seeding.md
divega: It is somehow not working well for me :trollface:. Maybe it is just me, but I find lists of orthogonal items easier to understand and remember

AndriySvyryd: We can wait for user feedback on this

entity-framework/core/modeling/data-seeding.md
```diff
@@ -7,19 +7,74 @@ uid: core/modeling/data-seeding
 ---
 # Data Seeding
 
-> [!NOTE]  
+Data seeding is the process of populating a database with an initial set of data.
+
+There are two main ways this can be accomplished in EF Core:
+* Model seed data
+* Custom initialization logic
```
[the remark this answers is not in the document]
AndriySvyryd: Added

PR #762, opened 2018-06-13
Update ILazyLoader coupling level

divega:
To reflect the fact that we moved it to the Abstractions package.

### DIFF
--- a/entity-framework/core/querying/related-data.md
+++ b/entity-framework/core/querying/related-data.md
@@ -235,7 +235,7 @@ public class Post
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


entity-framework/core/querying/related-data.md
```diff
@@ -235,7 +235,7 @@ public class Post
     }
 }
 ```
```
ajcvickers: Consider: ...which is defined in the [Microsoft.EntityFrameworkCore.Abstractions](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore.Abstractions/) package. _This package contains a minimal set of types such that there is little impact in depending on it. However, t_o avoid...

PR #787, opened 2018-06-21
Fixed an "active issues" link in the Mono & Xamarin compatibility footnote

Rabadash8820:
Should be self-explanatory 😛


divega:
Thank you @Rabadash8820!

PR #812, opened 2018-07-06
fix broken EF.Functions.Like snippet

SimonCropp:
an incorrect semicolon that stops it compiling


divega:
Thank you @SimonCropp!

PR #816, opened 2018-07-09
Update tables.md

kustox:
Typo


divega:
Thanks @kustox. Good catch!

PR #823, opened 2018-07-10
Added using-blocks to manually opened connections in transaction samples

Skleni:
Some of the transaction-related samples create a new SqlConnection and call the Open() method before passing it to a DbContext. After discussing the issue on [StackOverflow](https://stackoverflow.com/questions/51239326/do-i-need-to-close-the-dbconnection-manually-when-using-ambient-transactions-wit#51239326), I think in that case the connection should also be closed by the caller.


divega:
@AndriySvyryd can you please review?

divega:
Thanks a lot @Skleni!

PR #829, opened 2018-07-11
Fixed small typo 'form' > 'from'

andreujuanc:
ORIGINAL: The optimization requires buffering the results form the subquery.
CORRECT: The optimization requires buffering the results from the subquery.

Unless it meant "to form the subquery, but I find it unlikely. That's why I went straight to PR.

Cheers


divega:
Thanks @andreujuanc!

PR #843, opened 2018-07-17
Update related-data.md

vaindil:
Fix a bunch of typos and grammar errors, as well as some code formatting. Many of the code blocks on the page show `Csharp` as the language instead of `C#`, which I'm pretty sure is because they're declared as `Csharp` instead of `csharp`.


divega:
This is great @vaindil. Thank you very much!

PR #854, opened 2018-07-22
Fix grammar "understanding" -> "understand"

TomVeltmeijer:
Original: "In order to understanding how EF Core tracks these objects, ..."
Fixed: "In order to understand how EF Core tracks these objects, ..."


divega:
Thank you @TomVeltmeijer!

PR #872, opened 2018-07-31
Fixed typo

patriksvensson:


divega:
Thank you @patriksvensson! Good catch!

PR #907, opened 2018-08-09
Update UWP tutorial for 2.1

### LINKED ISSUE #581: Improve guidance to use migrations in UWP
Currently in https://docs.microsoft.com/ef/core/get-started/uwp/getting-started we recommend to create a .NET Standard 2.0 library and then edit the .csproj file and replace `<TargetFramework>netstandard2.0</TargetFramework>` with `<TargetFrameworks>netcoreapp2.0;netstandard2.0</TargetFrameworks>`. 

But this results in a pretty bad experience in VS, especially if you do things in a different order: VS fails to add a project reference to a class library that has both targets: `<TargetFrameworks>netcoreapp2.0;netstandard2.0</TargetFrameworks>`. If you add the reference first and then add the netcoreapp2.0 target, it will kind of work but complain with a yellow bang on the reference. We don't know what else could be broken. 

Given that, I believe we should just recommend customers create a separate netcoreapp2.0 console app that targets the same netstandard2.0 library.

### LINKED ISSUE #748: Test all the walkthroughs with 2.1 release
Using VS2017 15.7.3 with latest EF components fails when using the above documentation.  Add-Migration MyFirstMigration does nothing other than rebuilding the solution with the Model.  Please address this ASAP.  This is about the 5th reboot on EF for UWP, and each time the documentation is lacking to get to a working state.

How was this missed?

`<PropertyGroup>
   <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
 </PropertyGroup>`

tdykstra:
Fixes #748 (this is the last of the get-started tutorials to be updated, and this is the one 748 was originally about)

Fixes #581


tdykstra:
@divega - Who should review this in your absence?

divega:
@tdykstra, @AndriySvyryd is probably the last one that look at this s besides me.

I looked at the changes and I like the changes, but I guess the reason it is still WIP is that the text still talks about editing csproj files and using <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks> in one of them, which shouldn’t be necessary having the extra console app.

Also, I think the name of the console app is a bit confusing. Perhaps we should switch to Blogging for the prefix on all three projects, e.g. Blogging.Model, Blogging.UWP and Blogging.Migrations.Startup, or something on those lines.

tdykstra:
@divega - Apologies for premature call to review -- leaving the .csproj edit instructions in was an oversight, and other parts were omitted because I thought changes were saved but VS Code didn't save on exit (I can't imagine why they made that the default behavior in VS Code ... I have to find out how to turn the "feature" off).  Those should now be fixed and I changed the project names per your suggestion.

entity-framework/core/get-started/uwp/getting-started.md
```diff
@@ -2,137 +2,150 @@
...
+ms.date: 08/08/2018
 ms.assetid: a0ae2f21-1eef-43c6-83ad-92275f9c0727
 uid: core/get-started/uwp/getting-started
 ---
 
 # Getting Started with EF Core on Universal Windows Platform (UWP) with a New Database
 
-> [!NOTE]
-> This tutorial uses EF Core 2.0.1 (released alongside ASP.NET Core and .NET Core SDK 2.0.3). EF Core 2.0.0 lacks some crucial bug fixes required for a good UWP experience.
+In this tutorial, you build a Universal Windows Platform (UWP) application that performs basic data access against a local SQLite database using Entity Framework.
```
divega: Using entity framework core

entity-framework/core/get-started/uwp/getting-started.md
```diff
@@ -2,137 +2,150 @@
...
 
-In this walkthrough, you will build a Universal Windows Platform (UWP) application that performs basic data access against a local SQLite database using Entity Framework.
-
-> [!IMPORTANT]
-> **Consider avoiding anonymous types in LINQ queries on UWP**. Deploying a UWP application to the app store requires your application to be compiled with .NET Native. Queries with anonymous types have worse performance on .NET Native.
```
divega: Reviewing this on my phone so, not sure I get the full picture. This paragraph seems to be effectively removed. I am actually ok with that, but I think we should have a link to a list of UWP limitations, probably a new section on platforms.md.

PR #930, opened 2018-08-22
V1300944: Converting commands to codebox to avoid localization issues

v-maudel:
Hello, @bricelam,
"Unescaped commands are generating conflicts with translated pages:
https://docs.microsoft.com/fr-fr/ef/core/miscellaneous/cli/dotnet#using-the-tools
https://docs.microsoft.com/ja-jp/ef/core/miscellaneous/cli/dotnet#using-the-tools
"
Please review and merge the proposed file change to fix to target versions. If you make related fix in another PR  then share your PR number so we can confirm and close this PR.
Many thanks in advance.

### DIFF
--- a/entity-framework/core/miscellaneous/cli/dotnet.md
+++ b/entity-framework/core/miscellaneous/cli/dotnet.md
@@ -64,10 +64,10 @@ Using the tools
 Whenever you invoke a command, there are two projects involved:
 
 The target project is where any files are added (or in some cases removed). The target project defaults to the project
-in the current directory, but can be changed using the <nobr>**--project**</nobr> option.
+in the current directory, but can be changed using the <**`--project`**</nobr> option.
 
 The startup project is the one emulated by the tools when executing your project's code. It also defaults to the project
-in the current directory, but can be changed using the **--startup-project** option.
+in the current directory, but can be changed using the **`--startup-project`** option.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


entity-framework/core/miscellaneous/cli/dotnet.md
```diff
@@ -64,10 +64,10 @@ Using the tools
 Whenever you invoke a command, there are two projects involved:
 
 The target project is where any files are added (or in some cases removed). The target project defaults to the project
-in the current directory, but can be changed using the <nobr>**--project**</nobr> option.
+in the current directory, but can be changed using the <**`--project`**</nobr> option.
```
divega: @v-maudel isn't the '<' and `</nobr>` supposed to be removed here?

divega: Or maybe `nobrb>` which was removed should be kept?

v-maudel:
Hi @divega , I shouldn't delete the nobr tag, I've reverted it. Please, let me know if this is fine now. 

regards

divega:
Thanks @v-maudel. I will let @bricelam review the rest of the changes.

bricelam:
I don't like that the formatting has to change to facilitate localization, but if it's the only way ... :shipit:

v-maudel:
Thanks @divega and @bricelam . 
Actually I don't like this formatting either but is the only solution we have so far. 

regards.

PR #937, opened 2018-08-23
Updated example of adding a package using the command 'dotnet add package' to include proper usage syntax.

ricardodsanchez:
Added '<YOURPROJECT>' to the command informing you how to add a package because the usage of the command has changed.

### DIFF
--- a/entity-framework/core/get-started/netcore/new-db-sqlite.md
+++ b/entity-framework/core/get-started/netcore/new-db-sqlite.md
@@ -34,8 +34,8 @@ To use EF Core, install the package for the database provider(s) you want to tar
 * Install Microsoft.EntityFrameworkCore.Sqlite and Microsoft.EntityFrameworkCore.Design
 
   ```Console
-  dotnet add package Microsoft.EntityFrameworkCore.Sqlite
-  dotnet add package Microsoft.EntityFrameworkCore.Design
+  dotnet add <YOURPROJECT> package Microsoft.EntityFrameworkCore.Sqlite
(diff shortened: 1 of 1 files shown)


divega:
@ricardodsanchez can you please elaborate a bit more on why this change is needed?

ricardodsanchez:
@divega Of course, the current document isn't clear about the how when you run the following command:
`dotnet add package Microsoft.EntityFrameworkCore.Sqlite` the command won't work as expected if you don't include the <project> argument... unless you run the command in the same directory as your application. To make it clear and error free, my change indicates that you need to pass the <project> argument to the command as follows: `dotnet add <PROJECT> package Microsoft.EntityFrameworkCore.Sqlite`

ricardodsanchez:
>I agree that the cd ConsoleApp.SQLite/
Yes, I experienced it myself while following this document, and while I knew I had missed the step to switch to the right directory, I can see many other people missing it, especially new developers.

The change I am proposing makes this step error free, and it will help another person to run this step successfully, even if they missed the previous step.

divega:
@ricardodsanchez I don't understand why you think this makes the step error free. For example, if you just copy and paste the instructions (which is a reasonable thing to do when following a tutorial) the step will fail because '<YOURPROJECT>' is not a real project name. 

Moving the directory of the application before you start issuing `dotnet` commands is also most common once you learn about `dotnet` commands, so it seems to me that the change you are proposing makes the tutorial slightly worse than it was. I will make a change to reduce the chance try to make sure that changing the directory is not lost.

PR #943, opened 2018-08-27
modify the code example language to improve display

johnmwright:
On the docs site, using `Csharp` results in a language of `Csharp` to be displayed, where `csharp` results in `C#`

So instead of this:
image

You see this:
image


scottaddie:
@johnmwright The language support comes from highlight.js. See the following table: https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html#language-names-and-aliases.

divega:
@scottaddie it seems the problem that @johnmwright is addressing in this PR is that those aliases are case sensitive, so `Csharp` (used in the value-conversions page) does get a match with C#, like `csharp` does. 

Perhaps it is worth adding Csharp to the alias list, if it is frequently used for some reason (maybe some editor that switches to this casing by default). Do you think that making the match case insensitive is an improvement that we want to track somewhere else? 

In the meanwhile, I will go ahead and merge this PR. Thanks @johnmwright!

PR #964, opened 2018-09-05
Making changing the current directory its own step

divega:
Because we got [a PR](https://github.com/aspnet/EntityFramework.Docs/pull/937) that seems to indicate that it is a bit hidden right now.


divega:
cc @tdykstra @Rick-Anderson @mairaw, in case this is something you would like to do in other tutorials. In hindsight, perhaps we give it for granted that people will know to do this, when perhaps it isn't that obvious to everyone. 

For example, as far as I can see, https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/console-teleprompter is a bit confusing regarding in what order you should do `dotnet new`, and change the current directory.

PR #976, opened 2018-09-12
Broken link for [.NET Core CLI with SQLite] sample

DanielSSilva:
Updated the repo link to point to the correct repo. Previous link was returning a 404.


divega:
Thank you @DanielSSilva.

PR #1003, opened 2018-09-25
EF Code first sample model should use ICollection<T>

bartlannoeye:
Modified the EF Code first sample (and thus the tutorial docs) to use ICollection<T> as the other Code first and Db first samples use as well.

### DIFF
--- a/samples/core/GetStarted/AspNetCore/EFGetStarted.AspNetCore.NewDb/Models/Model.cs
+++ b/samples/core/GetStarted/AspNetCore/EFGetStarted.AspNetCore.NewDb/Models/Model.cs
@@ -18,7 +18,7 @@ namespace EFGetStarted.AspNetCore.NewDb.Models
         public int BlogId { get; set; }
         public string Url { get; set; }
 
-        public List<Post> Posts { get; set; }
+        public ICollection<Post> Posts { get; set; }
     }
 
     public class Post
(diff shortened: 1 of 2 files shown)


ajcvickers:
@bartlannoeye `List<T>` is valid here, even though it is probably more common to use `IList<T>` or `ICollection<T>`.

@divega Thoughts?

bartlannoeye:
I you speak about XAML databinding, then one typically would use [ObservableCollection<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.observablecollection-1?view=netframework-4.7.2). Next to that, we're typically living in a multi-tier landscape where data is exposed through REST APIs and as best-practice, the models exposed should not be your EF models. So for that reason your first argument seems invalid to me.

I've always learned to use/pass the "smallest" collection that the the correct job available (without interrupting your code path, e.g. if you need List functionality, you won't pass IEnumerable and use ToList in every layer). And I noticed the difference because my colleagues use the same principle, reviewing their code and seeing ICollection while I was under the initial impression it had to be a List to work (as I didn't go through the other samples until after I noticed).

However, if both List/IList and ICollection are valid options, it might be useful to put a footnote to mention this?

divega:
> I you speak about XAML databinding, then one typically would use ObservableCollection

Indeed, that is just another valid collection type we could be using. In fact we have the same model at https://github.com/aspnet/EntityFramework.Docs/blob/master/samples/core/GetStarted/UWP/Blogging.Model/Model.cs for UWP, using List. I suppose we get away with it only because the sample doesn't implement data binding to the collection. 

> I've always learned to use/pass the "smallest" collection that the the correct job available

I am ok with using this a reason to take this PR as is. Looking at the examples you trying to change, I don't think they would make any difference. 

> if both List/IList and ICollection are valid options, it might be useful to put a footnote to mention this?

Not sure there is enough value in introducing that informaiton int the tutorial, but I created https://github.com/aspnet/EntityFramework.Docs/issues/1023 for this.

PR #1005, opened 2018-09-27
Issue #988 - Grammer correction

### LINKED ISSUE #988: Typo
"Even when using LINQ, if you are accepting user input to build expressions you need to make sure &lt;&lt;&lt;than&gt;&gt;&gt; only intended expressions can be constructed." 
Needs to become: 
"Even when using LINQ, if you are accepting user input to build expressions you need to make sure &lt;&lt;&lt;that&gt;&gt;&gt; only intended expressions can be constructed.".

---
#### Document Details

* ID: 3208b215-1bac-c110-12f4-e1a6b448fdaf
* Version Independent ID: 3f6fb56f-0a79-8861-f8c4-ea5969a9aff5
* Content: [How Queries Work - EF Core](https://docs.microsoft.com/en-us/ef/core/querying/overview)
* Content Source: [entity-framework/core/querying/overview.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/querying/overview.md)
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

keyurpd:
Tried to improve the grammar and readability.

Fixes #988

### DIFF
--- a/entity-framework/core/querying/overview.md
+++ b/entity-framework/core/querying/overview.md
@@ -1,7 +1,7 @@
 ---
 title: How Queries Work - EF Core
 author: rowanmiller
-ms.date: 10/27/2016
+ms.date: 09/26/2018
 ms.assetid: de2e34cd-659b-4cab-b5ed-7a979c6bf120
 uid: core/querying/overview
 ---
@@ -40,4 +40,4 @@ The most common operations that result in the query being sent to the database a
 * Databinding the results of a query to a UI
 
 > [!WARNING]  
-> **Always validate user input:** While EF does provide protection from SQL injection attacks, it does not do any general validation of input. Therefore if values being passed to APIs, used in LINQ queries, assigned to entity properties, etc., come from an untrusted source then appropriate validation, per your application requirements, should be performed. This includes any user input used to dynamically construct queries. Even when using LINQ, if you are accepting user input to build expressions you need to make sure than only intended expressions can be constructed.
+> **Always validate user input:** While EF does provide protection from SQL injection attacks, it does not do any general validation of input. Therefore if values that are being passed to APIs, used in LINQ queries, assigned to entity properties etc., come from an untrusted source then appropriate validation, per your application requirements, should be performed. This includes any user input used to dynamically construct queries. Even when using LINQ, if you are accepting user input to build expressions, you need to make sure that only intended expressions can be constructed.


entity-framework/core/querying/overview.md
```diff
@@ -40,4 +40,4 @@ The most common operations that result in the query being sent to the database a
...
 
 > [!WARNING]  
-> **Always validate user input:** While EF does provide protection from SQL injection attacks, it does not do any general validation of input. Therefore if values being passed to APIs, used in LINQ queries, assigned to entity properties, etc., come from an untrusted source then appropriate validation, per your application requirements, should be performed. This includes any user input used to dynamically construct queries. Even when using LINQ, if you are accepting user input to build expressions you need to make sure than only intended expressions can be constructed.
+> **Always validate user input:** While EF does provide protection from SQL injection attacks, it does not do any general validation of input. Therefore if values that are being passed to APIs, used in LINQ queries, assigned to entity properties etc., come from an untrusted source then appropriate validation, per your application requirements, should be performed. This includes any user input used to dynamically construct queries. Even when using LINQ, if you are accepting user input to build expressions, you need to make sure that only intended expressions can be constructed.
```
divega: Agreed that grammar can be improved, thanks for the PR, but looking at this, I think there is an opportunity for polishing and simplifying things even more. For examples, by removing "being" instead of adding "that are", and even getting rid of etcetera, because "passed to APIs" covers pretty much all the cases anyway. 

If we don't remove the etcetera, getting rid of the comma seems incorrect. 

And I think there should be a comma before "then".

Proposal (including some other smaller changes):

>**Always validate user input:** While EF Core does provide protection from SQL injection attacks, it does not do any general validation of input. Therefore if values passed to APIs, used in LINQ queries, or assigned to entity properties come from an untrusted source, then appropriate validation, per the application's requirements, should be performed. This includes any user input used to dynamically construct queries. Even when using LINQ, if you are accepting user input to build expressions, you need to make sure that only intended expressions can be constructed.

keyurpd: Yeah, that statement is pretty long on the first place. Maybe we should shorten it, by removing examples or breaking them out.

keyurpd: But the proposal could work for time, thanks.

divega: Thanks @keyurpd. I will merge your PR and the I will make the additional edits.

PR #1013, opened 2018-10-02
C42263: Broken hyperlink

v-makoud:
Hello, @divega,
Localization team has reported source content issue that causes issue for localization. 
Description: The link doesn't seem to work anyway. So either it is needed to update the link or place it in the box. Also there are extra asterisks, which werent reported as a bug (LOC and ENU is matching).
Please review the comment tag added and reply with explanation if fix is needed or not. If you make related fix in another PR then share your PR number so we can confirm and close this PR.
Many thanks in advance.

### DIFF
--- a/entity-framework/ef6/fundamentals/databinding/wpf.md
+++ b/entity-framework/ef6/fundamentals/databinding/wpf.md
@@ -385,10 +385,10 @@ The code declares a long-running instance of **ProductContext**. The **ProductCo
 
 -   Press the **Save** button to save the data to the database
 
-After the call to DbContext’s **SaveChanges**(), the IDs are populated with the database generated values. Because we called **Refresh**() after **SaveChanges**() the **DataGrid** controls are updated with the new values as well.
+After the call to DbContext’s **SaveChanges()**, the IDs are populated with the database generated values. Because we called **Refresh()** after **SaveChanges()** the **DataGrid** controls are updated with the new values as well.
 
 ![Main Window with IDs populated](~/ef6/media/screen2.png)
 
 ## Additional Resources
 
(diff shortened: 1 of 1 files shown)


entity-framework/ef6/fundamentals/databinding/wpf.md
```diff
@@ -247,11 +247,11 @@ Add the classes that are defined in the model as data sources for this WPF appli
 
     ![Data Sources](~/ef6/media/datasources.png)
 
--   Select the **Category **data source and drag it on the form.
+-   Select the **Category **data source and drag it on the form.<!-- extra asterisks -->
```
divega: I don't think the asterisks are extra, they are just "misaligned". They should be surrounding the word Category. as in `**Category**`.

entity-framework/ef6/fundamentals/databinding/wpf.md
```diff
@@ -247,11 +247,11 @@ Add the classes that are defined in the model as data sources for this WPF appli
...
 
 The following happened when we dragged this source:
 
--   The **categoryViewSource** resource and the** categoryDataGrid** control were added to XAML. For more information about DataViewSources, see http://bea.stollnitz.com/blog/?p=387.
+-   The **categoryViewSource** resource and the** categoryDataGrid** <!-- extra asterisks -->control were added to XAML. For more information about DataViewSources, see http://bea.stollnitz.com/blog/?p=387.<!-- link seems to be broken -->
```
divega: Again, there are no extra asterisks, just misplaced. It should be `**categoryDataGrid**`.

divega: Replace the word DataViewSources with CollectionViewSource and the link with https://docs.microsoft.com/dotnet/api/system.windows.data.collectionviewsource.

divega:
@v-makoud the commited code doesn't have any of the comments you added, but as house keeping, could you please sign the CLA requested above?

PR #1021, opened 2018-10-03
Update link to migrations doc

jahbenjah:
The previous link doesn't help too much.

### DIFF
--- a/entity-framework/core/get-started/netcore/new-db-sqlite.md
+++ b/entity-framework/core/get-started/netcore/new-db-sqlite.md
@@ -59,7 +59,7 @@ Tip: In a real application, you put each class in a separate file and put the co
 
 ## Create the database
 
-Once you have a model, you use [migrations](https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/) to create a database.
+Once you have a model, you use [migrations](xref:core/managing-schemas/migrations/index) to create a database.
 
 * Run `dotnet ef migrations add InitialCreate` to scaffold a migration and create the initial set of tables for the model.
 * Run `dotnet ef database update` to apply the new migration to the database. This command creates the database before applying migrations.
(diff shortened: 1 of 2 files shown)


entity-framework/core/get-started/netcore/new-db-sqlite.md
```diff
@@ -88,10 +88,11 @@ The *blogging.db** SQLite DB is in the project directory.
...
-- The SQLite database engine doesn't support certain schema changes that are supported by most other relational databases. For example, the `DropColumn` operation is not supported. EF Core Migrations will generate code for these operations. But if you try to apply them to a database or generate a script, EF Core throws exceptions. See [SQLite Limitations](../../providers/sqlite/limitations.md). For new development, consider dropping the database and creating a new one rather than using migrations when the model changes.
-- 
+- The SQLite database engine doesn't support certain schema changes that are supported by most other relational databases. For example, the `DropColumn` operation is not supported. EF Core Migrations will generate code for these operations. But if you try to apply them to a database or generate a script, EF Core throws exceptions. 
+- See [SQLite Limitations](../../providers/sqlite/limitations.md). For new development, consider dropping the database and creating a new one rather than using migrations when the model changes.
```
divega: This was part of the paragraph above by design.

entity-framework/core/get-started/netcore/new-db-sqlite.md
```diff
@@ -59,7 +59,7 @@ Tip: In a real application, you put each class in a separate file and put the co
 
 ## Create the database
 
-Once you have a model, you use [migrations](https://docs.microsoft.com/aspnet/core/data/ef-mvc/migrations#introduction-to-migrations) to create a database.
+Once you have a model, you use [migrations](https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/) to create a database.
```
divega: In general we prefer [xref](https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html?tabs=tabid-1%2Ctabid-a#cross-reference) links when referencing a topic in the same document set:
``` markdown
[migrations](xref:core/managing-schemas/migrations/index)
```

Otherwise we don't include "en-us" in this URL. This allows the link to be redirected to the right locale if the docs are localized.

tdykstra: The link in question was meant to be

https://docs.microsoft.com/en-us/aspnet/core/data/ef-mvc/migrations?view=aspnetcore-2.0#introduction-to-migrations

which gives a little lengthier explanation of what migrations is than the EF doc does. The EF doc only has a single brief sentence before it jumps into the how-to instructions.  It might be worth expanding the intro of the EF doc or adding to some other EF doc a better explanation of what Migrations is and how it works.  In any case, I agree that linking to the ASP.NET doc was a mistake even if the link had the correct version query string.

divega: Thanks @tdykstra. I will look into copying some of the explanation about migrations into the EF Core docs, or you can do it yourself if you want.

divega:
Thank you @jahbenjah!

PR #1096, opened 2018-11-02
Document using spatial data on EF Core

### LINKED ISSUE #1006: Spatial
Coming in EF Core 2.2...

* NTS
  * Use a factory from `GeometryServiceProvider.Instance.CreateGeometryFactory(srid)` to create instances. `new` is possible but discouraged.
  * `M` is hard (needs a different CoordinateSequence/Factory) and doesn't always work (e.g. NetTopologySuite/NetTopologySuite#156)
  * `CircularString`, `CompoundCurve`, and `CurvePolygon` not supported (see NetTopologySuite/NetTopologySuite#247)
  * `SRID` not honored. `Area`, `Distance`, and `Length` will be in degrees
* InMemory (no additional package)
  * Install NTS; add properties; done
* SQLite (package Microsoft.EntityFrameworkCore.Sqlite.NetTopologySuite)
  * Needs SpatiaLite
    * Windows: In the box via NuGet
      * No AnyCPU on .NET Framework
      * No UWP
    * MacOS: `brew install libspatialite`
    * Linux: `apt-get install libsqlite3-mod-spatialite` (on Ubuntu anyway)
  * Configure SRID and dimension on properties
* SQL Server (package Microsoft.EntityFrameworkCore.SqlServer.NetTopologySuite)
  * Defaults to `geography`, can map to `geometry`
  * The usual `geography` gotchas
    * `X` is `Long`
    * `Y` is `Lat`
    * `Polygon` is weird
      * `ExteriorRing` must be oriented counter-clockwise
      * `InteriorRings` must be oriented clockwise
  * `FullGlobe` isn't supported by NTS
    * Including geography polygons without an exterior ring
  * Need a SQL-Server-specific operation? Let us know on aspnet/EntityFrameworkCore#13421

bricelam:
Resolves #1006

### DIFF
--- a/entity-framework/core/managing-schemas/ensure-created.md
+++ b/entity-framework/core/managing-schemas/ensure-created.md
@@ -2,22 +2,22 @@
 title: Create and Drop APIs - EF Core
 author: bricelam
 ms.author: bricelam
-ms.date: 11/10/2017
+ms.date: 11/7/2018
 ---
 # Create and Drop APIs
 
-The EnsureCreated and EnsureDeleted methods provide a lightweight alternative to [Migrations](migrations/index.md) for managing the database schema. This is useful in scenarios when the data is transient and can be dropped when the schema changes. For example during prototyping, in tests, or for local caches.
+The EnsureCreated and EnsureDeleted methods provide a lightweight alternative to [Migrations](migrations/index.md) for managing the database schema. These methods are useful in scenarios when the data is transient and can be dropped when the schema changes. For example during prototyping, in tests, or for local caches.
 
-Some providers (especially non-relational ones) don't support Migrations. For these, EnsureCreated is often the easiest way to initialize the database schema.
+Some providers (especially non-relational ones) don't support Migrations. For these providers, EnsureCreated is often the easiest way to initialize the database schema.
 
 > [!WARNING]
 > EnsureCreated and Migrations don't work well together. If you're using Migrations, don't use EnsureCreated to initialize the schema.
 
-Transitioning from EnsureCreated to Migrations is not a seamless experience. The simpelest way to achieve this is to drop the database and re-create it using Migrations. If you anticipate using Migrations in the future, it's best to just start with Migrations instead of using EnsureCreated.
+Transitioning from EnsureCreated to Migrations is not a seamless experience. The simplest way to do it is to drop the database and re-create it using Migrations. If you anticipate using migrations in the future, it's best to just start with Migrations instead of using EnsureCreated.
 
 ## EnsureDeleted
 
-The EnsureDeleted method will drop the database if it exists. If you don't have the appropiate permissions, an exception is thrown.
+The EnsureDeleted method will drop the database if it exists. If you don't have the appropriate permissions, an exception is thrown.
 
 ``` csharp
 // Drop the database if it exists
--- a/entity-framework/core/modeling/spatial.md
+++ b/entity-framework/core/modeling/spatial.md
@@ -1,0 +1,306 @@
+---
+title: Spatial Data - EF Core
+author: bricelam
+ms.author: bricelam
+ms.date: 11/01/2018
+uid: core/modeling/spatial
+---
+# Spatial Data
+
+> [!NOTE]
+> This feature is new in EF Core 2.2.
+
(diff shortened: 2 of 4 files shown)


bricelam:
This is my first draft. It still needs some wordsmithing and layouting, but it has all the content I wanted to cover. Let me know if you think anything is missing.

entity-framework/core/miscellaneous/spatial.md
```diff
@@ -0,0 +1,288 @@
+---
+title: Spatial Data - EF Core
+author: bricelam
+ms.author: bricelam
+ms.date: 11/01/2018
+uid: core/miscellaneous/spatial
+---
+# Spatial Data
```
divega: [!NOTE]
This feature was introduced in EF Core 2.2.

bricelam:
I should add a comment about reverse engineering too...

entity-framework/core/miscellaneous/spatial.md
```diff
@@ -0,0 +1,295 @@
...
+    public IGeometry Border { get; set; }
+}
+```
+
+### Creating values
+
+You can use constructors to create geometry objects; however, NTS recommends using a geometry factory instead. This lets you specify a default SRID and gives you control over more advanced things like the precision model (used during calculations) and the coordinate sequence (determines which ordinates--dimensions and measures--are available).
+
+``` csharp
+var geometryFactory = NtsGeometryServices.Instance.CreateGeometryFactory(srid: 4326);
```
divega: Maybe add a note about 4326 and some wikipedia(?) link to coordinate systems?

divega: BTW, I understand we may want to stay away from trying to explain a lot of these things in our own docs but I think there is a set of customers for which we are making spatial data more accessible, and they may find links and brief explanations useful.

entity-framework/core/miscellaneous/spatial.md
```diff
@@ -0,0 +1,295 @@
...
+var currentLocation = geometryFactory.CreatePoint(-122.121512, 47.6739882);
+```
+
+### Longitude and Latitude
+
+Coordinates in NTS are in terms of X and Y values. To represent longitude and latitude, use X for longitude and Y for latitude. Note that this is **backwards** from the `latitude, longitude` format in which you typically see these values.
+
+### SRID Ignored
```
divega: "SRID Ignored in client computations"?

entity-framework/core/miscellaneous/spatial.md
```diff
@@ -0,0 +1,295 @@
...
+
+[Table("Countries", Schema = "Application"))]
+class Country
+{
+    public int CountryID { get; set; }
+
+    public string CountryName { get; set; }
+
+    // Database includes both Polygon and MultiPolygon values
+    public IGeometry Border { get; set; }
```
divega: The comment is good. But I wonder if there is worth discussing in the text when it makes sense to constrain a property type (e.g. if all the instances are going to be a Point) vs. leaving it as Geometry, after showing this example.

entity-framework/core/miscellaneous/spatial.md
```diff
@@ -0,0 +1,295 @@
...
+    * MultiPoint
+    * MultiLineString
+    * MultiPolygon
+
+> [!WARNING]
+> CircularString, CompoundCurve, and CurePolygon aren't supported by NTS.
+
+The following entity classes could be used to map to tables in the [Wide World Importers sample database](http://go.microsoft.com/fwlink/?LinkID=800630).
+
+``` csharp
```
divega: Are you planning to create a working sample project? I think in this case it could be very useful, but no problem with doing it afterwards.

entity-framework/core/miscellaneous/spatial.md
```diff
@@ -0,0 +1,295 @@
+---
+title: Spatial Data - EF Core
+author: bricelam
+ms.author: bricelam
+ms.date: 11/01/2018
+uid: core/miscellaneous/spatial
```
divega: Any reasons you think this should be under miscellaneous instead of modeling?

bricelam: It's a bit of everything--reverse engineering, modelling, querying, using NTS...

But I'm happy with it wherever.

divega: Yes, you can do all those things with spatial, but isn’t that true for most things under modeling? I have the impression that the ability to map a type or concept in the model is the defining characteristic here.

divega: Ok. Thank you for letting me know.

entity-framework/core/miscellaneous/spatial.md
bricelam: We should consider moving [this sample](https://github.com/divega/EFDemos/tree/master/spatial) into either here or the main repo.

PR #1142, opened 2018-11-29
C52739: Asterisk issue

v-makoud:
Hello, @divega,
Description of the source issue:
string in (for example) form "**Entity.**EntityState" doesn't render in bold on EN page and asterisks are shown instead. Problem is the dot before two asterisk. Possible fix for this would be moving a dot after two asterisks. 
Please review and merge the proposed file change to fix to target versions. If you make related fix in another PR  then share your PR number so we can confirm and close this PR.
Many thanks in advance.

### DIFF
--- a/entity-framework/ef6/what-is-new/upgrading-to-ef6.md
+++ b/entity-framework/ef6/what-is-new/upgrading-to-ef6.md
@@ -61,8 +61,8 @@ Types like ObjectContext that were previously in System.Data.Entity.dll have bee
 The general rule for namespace changes is that any type in System.Data.* is moved to System.Data.Entity.Core.*. In other words, just insert **Entity.Core.** after System.Data. For example:
 
 - System.Data.EntityException => System.Data.**Entity.Core**.EntityException  
-- System.Data.Objects.ObjectContext => System.Data.**Entity.Core.**.Objects.ObjectContext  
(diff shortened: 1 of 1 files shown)


entity-framework/ef6/what-is-new/upgrading-to-ef6.md
```diff
@@ -60,20 +60,20 @@ Types like ObjectContext that were previously in System.Data.Entity.dll have bee
 
 The general rule for namespace changes is that any type in System.Data.* is moved to System.Data.Entity.Core.*. In other words, just insert **Entity.Core.** after System.Data. For example:
 
-- System.Data.EntityException => System.Data.**Entity.Core.**EntityException  
-- System.Data.Objects.ObjectContext => System.Data.**Entity.Core.**Objects.ObjectContext  
-- System.Data.Objects.DataClasses.RelationshipManager => System.Data.**Entity.Core.**Objects.DataClasses.RelationshipManager  
+- System.Data.EntityException => System.Data.**Entity.Core**.EntityException  
+- System.Data.Objects.ObjectContext => System.Data.**Entity.Core.**.Objects.ObjectContext
```
divega: This and next one still seem incorrect.

v-makoud:
@divega I have deleted the extra dots, it should be correct now.

divega:
Thanks @v-makoud!

PR #1167, opened 2018-12-10
Add efg to tools and extensions page

divega:
cc @pwelter34. Let us know if there is anything you would like added or edited. I intend to merge this today.

### DIFF
--- a/entity-framework/core/extensions/index.md
+++ b/entity-framework/core/extensions/index.md
@@ -39,6 +39,13 @@ A Visual Studio 2017 extension that adds an ORM designer for visual design of En
 
 [Marketplace](https://marketplace.visualstudio.com/items?itemName=michaelsawczyn.EFDesigner)
 
+### LoreSoft's Entity Framework Core Generator
+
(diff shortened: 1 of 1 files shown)


entity-framework/core/extensions/index.md
```diff
@@ -45,6 +45,13 @@ CatFactory is a scaffolding engine for .NET Core and Entity Framework Core. The
 
 [GitHub repository](https://github.com/hherzl/CatFactory.EntityFrameworkCore)
 
+### LoreSoft's Entity Framework Core Generator
+
+Entity Framework Core Generator (efg) is a .NET Core CLI tool that can generate EF Core models from an existing database, much like `dotnet ef dbcontext scaffold`. However its different int that it also supports safe code [regeneration](https://efg.loresoft.com/en/latest/regeneration/). Regeneration is accomplished either via region replacement or by parsing mapping files. It also supports view model, validation and object mapper generation. For more information, see the tutorial and the product documentation.
```
pwelter34: Minor typo,  `However its different int that it also supports safe code`...  should be `However its different in that it also supports safe code`.

divega: Good catch! It's also the wrong version of "its" :smile:

PR #1172, opened 2018-12-12
Example Code Screenshot mismatch

diegorod:
Fixed example code to match output displayed on syncoutput.png.


divega:
Thank you @diegorod!

PR #1177, opened 2018-12-14
M33118: Sentence needs review [DO NOT MERGE]

v-makoud:
Hello, @ErikEJ,
Translator has reported possible source content issue. 
Description:
This sentence is almost incomprehensible, non-standard use of semicolon, missing words. Author should probably review (and rephrase) this sentence. Thank you.
Please also compare this sentence to a sentence written in a different article by the same author: The key thing in CatFactory it's import an existing database from a SQL Server instance, then scaffold code for specific technology (https://github.com/hherzl/CatFactory). > it seems like "export" in the sentence below needs to be updated to "import".

Please review the comment tag added and reply with explanation if fix is needed or not. If you make related fix in another PR then share your PR number so we can confirm and close this PR.
Many thanks in advance.

### DIFF
--- a/entity-framework/core/extensions/index.md
+++ b/entity-framework/core/extensions/index.md
@@ -41,7 +41,7 @@ A Visual Studio 2017 extension that adds an ORM designer for visual design of En
 
 ### CatFactory
 
-CatFactory is a scaffolding engine for .NET Core and Entity Framework Core. The concept behind CatFactory it's export an existing database from SQL Server instance, then with the representation in models for database; scaffold entities, configurations, repositories and more.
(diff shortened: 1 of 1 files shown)


hherzl:
Hi.

How about this?

> CatFactory it's a scaffolding engine for .NET Core built with C#.
> 
> The concept behind **CatFactory** is import an existing database from SQL Server instance, then scaffolding *Entity Framework Core*. **CatFactory** allows to scaffold Entities, DbContext, Configurations and Repositories; Services are on the roadmap and much more.

I look forward for your feedback.

v-maudel:
Thanks so much @divega and @hherzl 
Also, could you please change the author of the article? We first assign it to Erikej because he appears in the header of the file as author.

For record: Fixed here: https://github.com/aspnet/EntityFramework.Docs/pull/1193

v-maudel:
@v-makoud #please-close

divega:
The actual fix will be in https://github.com/aspnet/EntityFramework.Docs/pull/1217. Closing this now.

PR #1259, opened 2019-01-30
Fix grammar for including multiple related entities

pyrox18:
Changed

```
This does not mean you will get redundant joins, in most cases EF will consolidate the joins when generating SQL.
```

to

```
This does not mean you will get redundant joins; in most cases, EF will consolidate the joins when generating SQL.
```

for better clarity.


ajcvickers:
Thanks!

PR #1265, opened 2019-02-04
Removed an extra square bracket

nmur:
This may have been a hyperlink in the past (to something like [this](https://docs.microsoft.com/en-us/dotnet/core/tools/?tabs=netcore2x)).

I've just removed the bracket, but can add in the hyperlink if desired.


divega:
Thank you @nmur!

PR #1268, opened 2019-02-04
Add example of disabling tracking in raw queries

jzabroski:


divega:
@jzabroski Thanks for submitting this for consideration. 

First, I am curious about how you arrived to the conclusion that the example has to do with query types? Based on the fact that BloggingContext.Blogs is a `DbSet<Blog>`, Blog is a regular entity type, and that is why it can be tracked. This is completely orthogonal to the usage of FromSql(). 

We have been discussing lately that the way we introduced query types, even the name of the concept, and the name of the API, can cause confusion for some customers. For example, EF Core customers need to get familiar with the new APIs to be able to distinguish between entity types and query types. The name query is also very overloaded in a product like EF Core. 

In fact, we are now considering [consolidating the concepts of query types and entity types](https://github.com/aspnet/EntityFrameworkCore/issues/14194). So it is likely that in 3.0 we will stop talking about query types and we will just talk about entity types that don't have keys, and hence cannot be tracked. Would this have helped?

Second, it seems to me that the underlying motivation is to show that types returned by a query with FromSql() can be tracked or not tracked, following the same rules as queries that don't use FromSql(). If that the case, I think that is what the documentation should say.

jzabroski:
To be honest, I'm not 100% sure how I came to this conclusion. HOWEVER, I
will say that EFCore proxies can be horribly slow, so I've just gotten so
used to slapping AsNoTracking() everywhere. If I had to guess, most people
do that same hack because proxies are so awful in EFCore that it's
literally not worth using.

However, it does bring up an important point indirectly touched upon by you:

What I really want is a way for my compiler to tell me I need to add a
column to a POCO and wire it up to EF. Even better if, as in a ReSharper
warning, I can click a button to confirm the change and fix the warning.

That's it. Then I want to go drink a beer.

I don't care about query types and entity types. Query types have caused
stupid discussions on our team, which I can summarize in a follow up to
this, but I really just want to take this philosophical discussion about
good API design and discuss what's important to me.

Consider for example I can mostly do this already with F# type providers
and pretty much skip EF and just use SQLProvider. HOWEVER, it becomes a
giant mess on large projects with many migrations and team members.
Somewhere there is a middle ground that makes debugging, say, a bad code
merge on a huge project reverse integrating from a maintenance branch to a
release branch ACTUALLY FEEL PAINLESS.

We will get there, my friend Diego. And I will have a beer waiting for you.

divega:
@jzabroski I have read your feedback, and some of it maps to existing issues. 

You say that "EF Core proxies" are slow, which I think means you think EF Core change tracking is slow (there are no change tracking proxies in EF Core). This maps to issue https://github.com/aspnet/EntityFrameworkCore/issues/9422 and can be mitigated in part in some scenarios if we fix https://github.com/aspnet/EntityFrameworkCore/issues/13552 in 3.0. 

I have captured the need to explain that FromSql() is orthogonal to tracking vs. non tracking in https://github.com/aspnet/EntityFramework.Docs/issues/1271. I will try to take your PR and make a few edits to accomplish that. 

Hopefully making query types just entity types without keys will remove some conceptual overhead for customers that learn EF Core in the future. 

To be honest, I am not sure what to do with your other feedback:

> What I really want is a way for my compiler to tell me I need to add a column to a POCO and wire it up to EF. Even better if, as in a ReSharper warning, I can click a button to confirm the change and fix the warning.

It seems you want the database schema to be the source of truth for the model, so this maps to somewhat to update model from database https://github.com/aspnet/EntityFrameworkCore/issues/831, but it is hard to imagine this process being triggered in the background to become as seamless as you want it to be. 

Type providers are indeed an interesting option, unfortunately only available to F# developers. FWIW, the property bag entities feature we are doing in EF Core 3.0 should make it easier to write an erasing type provider that then uses EF Core.

jzabroski:
@divega  Wow, thanks.  Just got around to reading this and blasting through all my GitHub notifications,  you are always so thoughtful.  Hope I made your life better and not harder! :)

PR #1291, opened 2019-02-16
Add example for ASP.NET Core application

ncarandini:
I think it can be useful for many devs.


divega:
Thank you @ncarandini!

PR #1318, opened 2019-02-27
Suggested improvement to the wording of the warning for the EF Core Raw SQL Document

HockeyJustin:
See [this issue](https://github.com/aspnet/EntityFramework.Docs/issues/1295) for previous discussion and correspondence with @divega .

The purpose of this update is to clarify that; while both string interpolation and params methods provide protection against SQL Injection, Entity Framework cannot protect your application against non-parameterized dynamic sql.

I believe this update will provide clarification for those making comments on the article. It may also help ensure that people check their stored procedures as well as their EF code when doing security reviews.

### DIFF
--- a/entity-framework/core/querying/raw-sql.md
+++ b/entity-framework/core/querying/raw-sql.md
@@ -134,6 +134,6 @@ There are a few limitations to be aware of when using raw SQL queries:
 * SQL statements other than `SELECT` are recognized automatically as non-composable. As a consequence, the full results of stored procedures are always returned to the client and any LINQ operators applied after `FromSql` are evaluated in-memory.
 
 > [!WARNING]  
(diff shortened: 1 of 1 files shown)


divega:
@HockeyJustin I have made a few tweaks to the PR. Let me know how you think. I am mostly satisfied with the result, but perhaps we should separate this idea of passing input to stored procedures or statements that can execute dynamic SQL. To me it is just a trivial example of something the sentence is already covering.

HockeyJustin:
@divega I'm happy with your suggestion. I think it provides a good clarification.

One minor point. Should the following part: 

...or passing use input..

be 

...or passing user input..

?

In case you are interested, [here](https://stackoverflow.com/questions/54750151/secure-way-of-accessing-sql-stored-procedures-from-entity-framework-core) is a bit more background on why I was keen to provide some clarification. 

I think your final wording will do just that.

Thank you for your consideration.

divega:
Thanks @HockeyJustin for your contribution!

PR #1324, opened 2019-02-28
Made the property DateCreated in class BlogDetails non-nullable.

mikepetty21:
It appears the intent in the ComplexType code example was to make the DateCreated property non-nullable.  This is confirmed by this excerpt in the document: "Another interesting note is that although the DateCreated property was defined as a non-nullable DateTime in the class, the relevant database field is nullable. You must use the Required annotation if you wish to affect the database schema."

### DIFF
--- a/entity-framework/ef6/modeling/code-first/data-annotations.md
+++ b/entity-framework/ef6/modeling/code-first/data-annotations.md
@@ -207,7 +207,7 @@ It’s not uncommon to describe your domain entities across a set of classes and
 ``` csharp
     public class BlogDetails
     {
-        public DateTime? DateCreated { get; set; }
+        public DateTime DateCreated { get; set; }
 
         [MaxLength(250)]
         public string Description { get; set; }
@@ -222,7 +222,7 @@ However as a property in the Blog class, BlogDetails it will be tracked as part 
     [ComplexType]
     public class BlogDetails
     {
-        public DateTime? DateCreated { get; set; }
+        public DateTime DateCreated { get; set; }
 
         [MaxLength(250)]
         public string Description { get; set; }


divega:
@mikepetty21 Not sure how much value there is in this, but I would be ok with taking the change you propose. 

The only problem is that the paragraph you quoted seems incorrect. I am not sure what was the original intent but perhaps it is just the product of @julielerman playing with preview bits of EF when she wrote the original article? :smile:

mikepetty21:
I went ahead to implemented a working example of the Blog and BlogDetails entities. Here are my observations:

1. It appears the ComplexType data annotation does not work in EF Core version 2.2.2. When running "add-migration", I get an error message stating "The entity type 'BlogDetails' requires a primary key to be defined." The fix for this is to use the fluent API by overiding the OnModelCreated method in the DbContext and adding the statement "modelBuilder.Owned<BlogDetails>();". The ComplexType data annotation doesn't appear to be needed in this case.
2. In SQL Server, the nullability of column BlogDetail_DateCreated in table Blogs is determined by the data type of DateCreated property in the BlogDetails entity. Therefore, if it's a nullable DateTime (DateTime?), the column BlogDetail_DateCreated allows nulls in the DB. If it's a normal DateTime, the column BlogDetail_DateCreated does not allow nulls in the DB.
3. It appears the [Required] data annotation is not very useful for value types (DateTime, int, bool) since the type can be defined as a nullable type (like DateTime?). The [Required] data annotation appears to make more sense for reference types (like string or object) since they are by definition nullable.
4. It appears to me the ComplexType data annotation is not working correctly. If this is the case, maybe a fix is in order. If it is working correctly, maybe the docs should shed some light on how to use it properly. If ComplexType isn't supported in core, maybe the docs should refer the reader to using modelBuilder.Owned in the fluent API instead.
5. In the meantime, the docs should maybe just indicate the ComplexType data annotation is not working (maybe indicating version numbers) and the workaround is to use modelBuilder.Owned<T> in the fluent API.

Also, in December of 2018, victor-david noticed this inconsistency in the docs: https://github.com/aspnet/EntityFramework.Docs/issues/1160.

I know this sounds like a can of worms. Thoughts?

julielerman:
Just curious ... what original article are you referring to? I haven't worked on this one ...I don't think I've used data annotations since about EF 4.1! ;)

ajcvickers:
@mikepetty21 These are the docs for EF6, but your comments seem to be about EF Core. Please see the [EF Core documentation](https://docs.microsoft.com/en-us/ef/core/) for documentation on EF Core.

divega:
Closing as it seems this was the wrong set of documents (EF6 vs. EF Core). I will remove the paragraph that is incorrect in a separate commit.

PR #1325, opened 2019-02-28
Fixed sentence regarding InverseProperty

mikepetty21:
Removed the extra word "will" from the statement "but code first will create four will foreign key properties:" to read "but code first will create four foreign key properties:".


ajcvickers:
Thanks!

PR #1327, opened 2019-02-28
Edits to breaking changes doc

ajcvickers:
Specifically:
* Pre-announce client evaluation (for discussion--do we want to do this?)
* Start adding the preview in which these changes were introduced
* Use a common format of heading inside each change--looking for feedback on this too
* Stop using `Code` markup in headings, because it renders horribly


entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -15,28 +15,75 @@ These are breaks in either API or behavior between the 2.2.x releases and the 3.
...
+> [!IMPORTANT]
+> We are pre-announcing this break. It has not yet shipped in any 3.0 preview.
+
+#### Old behavior
+
+Prior to EF Core 3.0, any part of a query that could not be translated to SQL would be instead evaluated on the client.
+
+#### New behavior
+
+Starting with EF Core 3.0, if some part of a query cannot be translated to SQL, then an exception is thrown.
```
divega: > Starting with EF Core 3.0, we only allow the top level projection (the expression passed to the last Select LINQ operator in the query) to be evaluated on  the client.  If expressions in any other part of the query cannot be translated to SQL, then an exception is thrown.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -15,28 +15,75 @@ These are breaks in either API or behavior between the 2.2.x releases and the 3.
...
+
+#### Why
+
+The automatic client-evaluation of queries allowed many queries to be executed even if they could not be translated. However, this can result in unexpected and potentially damaging behavior that may only become apparent in production. Also, it results in very fragile behavior for queries that breaks easily from one release to the next.
```
divega: > The automatic client-evaluation of queries allowed many queries to be executed even if important parts of them could not be translated. This can result in unexpected and potentially damaging behavior that may only become apparent in production. For example, a query containing an expression in the Where LINQ operator that cannot be translated, can result in all rows from the table to be transferred from the database server, and the filter to be applied on the client. This can easily go undetected if the table contains only a few rows in development and staging, but hit hard when you move the application to production where the table contains millions of rows. Client evaluation warnings also proved too easy to ignore during development.  
Besides, automatic client-evaluation lead issues in which improving query translation for specific expressions caused breaking changes between releases.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -15,28 +15,75 @@ These are breaks in either API or behavior between the 2.2.x releases and the 3.
...
+### Queries are not evaluated on the client
+
+[Tracking Issue](https://github.com/aspnet/EntityFrameworkCore/issues/12795)
+
+> [!IMPORTANT]
+> We are pre-announcing this break. It has not yet shipped in any 3.0 preview.
+
+#### Old behavior
+
+Prior to EF Core 3.0, any part of a query that could not be translated to SQL would be instead evaluated on the client.
```
divega: > Prior to EF Core 3.0, any part of a query that could not be translated to SQL would be instead [evaluated on the client](xref:core/querying/client-eval). Client evaluation of potentially expensive expressions only triggered a warning by default.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
ajcvickers: Looks good.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
ajcvickers: I was making an attempt to be brief with the "why" parts--more like a TL;DR of the actual tracking issue. That being said, I do think especially for this one it makes sense to clearly say why we're doing it. But...with fewer words? :trollface:

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
ajcvickers: Sounds good.

ajcvickers:
Merging what we have now. Will make further edits based on feedback and in-person discussion.

PR #1335, opened 2019-03-04
Line break per sentence in 3.0 breaking changes

divega:
To optimize for future reviews.


ajcvickers:
This is hideous! Why?

divega:
It is a normal practice to make changes in markdown easier to review. Markdown treats sentences separated like this as part of the same paragraph. I have another PR coming that can use that help.

ajcvickers:
@divega Then I defer to someone else to approve this.

PR #1353, opened 2019-03-13
Update raw-sql.md

lgonzalezr:
The example doesn't really do a named parameter call. I had this bug and fixed with my change in a real product.

### DIFF
--- a/entity-framework/core/querying/raw-sql.md
+++ b/entity-framework/core/querying/raw-sql.md
@@ -65,7 +65,7 @@ You can also construct a DbParameter and supply it as a parameter value. This al
 var user = new SqlParameter("user", "johndoe");
 
 var blogs = context.Blogs
-    .FromSql("EXECUTE dbo.GetMostPopularBlogsForUser @user", user)
+    .FromSql("EXECUTE dbo.GetMostPopularBlogsForUser @user=@user", user)
     .ToList();
 ```
 


divega:
@lgonzalezr I suspect when we wrote this, we were not even thinking about stored procedures with named parameters, but just show than you can specify the name of the parameter when you create the DbParameter object and then refer to it in the SQL. Anyway, this is a good catch!

1. It would be good to find a good documentation page for SqlClient that describes when (i.e. the motivation) you use named parameters (e.g. when there are multiple optional parameters). I did a quick search but I couldn't find something appropriate. 

2. Maybe this could be a second example? I think there is value in showing that what the current example does works.

lgonzalezr:
@divega I agree with you that the example has value on its own and probably another example will fit the named parameter case better. However the above explanation that states "This allows you to use named parameters in the SQL query string." it's a little misleading. 

Anyway I just wanted to suggest a probable ambiguous explanation, or it is only me :-). The doc is excellent and I use it a lot, great job! Thanks.

divega:
> However the above explanation that states "This allows you to use named parameters in the SQL query string." it's a little misleading.

Agreed. If you don't intend to make further changes, I will merge your PR but I will then add a separate example that just shows that you can refer to the name assigned to the DbParameter in the SQL.

PR #1363, opened 2019-03-18
Update new-db.md

OOberoi:
Simplified the paragraph by eliminating "these services" which is being used twice in the same sentence.

### DIFF
--- a/entity-framework/core/get-started/aspnetcore/new-db.md
+++ b/entity-framework/core/get-started/aspnetcore/new-db.md
@@ -109,7 +109,7 @@ A production app would typically put each class in a separate file. For the sake
 
 ## Register the context with dependency injection
 
-Services (such as `BloggingContext`) are registered with [dependency injection](http://docs.asp.net/en/latest/fundamentals/dependency-injection.html) during application startup. Components that require these services (such as MVC controllers) are provided these services via constructor parameters or properties.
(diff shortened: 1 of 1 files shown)


ajcvickers:
@OOberoi Agreed the wording is clumsy, but with the edit it reads to me like the components are being provided, not the services.

entity-framework/core/get-started/aspnetcore/new-db.md
```diff
@@ -109,7 +109,7 @@ A production app would typically put each class in a separate file. For the sake
 
 ## Register the context with dependency injection
 
-Services (such as `BloggingContext`) are registered with [dependency injection](http://docs.asp.net/en/latest/fundamentals/dependency-injection.html) during application startup. Components that require these services (such as MVC controllers) are provided these services via constructor parameters or properties.
+Services (such as `BloggingContext`) are registered with [dependency injection](http://docs.asp.net/en/latest/fundamentals/dependency-injection.html) during application startup. Components that require these services (such as MVC controllers) are provided via constructor parameters or properties.
```
divega: What about this?

> Services (such as `BloggingContext`) are registered with [dependency injection](http://docs.asp.net/en/latest/fundamentals/dependency-injection.html) during application startup so that they can be provided automatically to components that consume services (such as MVC controllers) via constructor parameters and properties.

divega: Also make this the first sentence of the topic:

> To make `BloggingContext` available to MVC controllers, register it as a service.

The rest is just a generic explanation of the role of DI.

OOberoi:
Quite succinct.
If I might add..."To make BloggingContext available to MVC controllers, register it as a service in Startup.cs".

PR #1372, opened 2019-03-19
Update Limitations reference Named Mapping

SeanMartz:
Update Limitations section to reflect issue with name mapping vs. order mapping from stored procedure.


divega:
Thanks @SeanMartz.

divega:
@SeanMartz If you have an actual repro for https://github.com/aspnet/EntityFrameworkCore/issues/13489? then it would be great if you could add it to the issue. I tried this myself today but I don't see the bug. 

For now, I will revert this change, until we know more.

PR #1377, opened 2019-03-20
Update mocking.md

rschindhelm-haba:


divega:
Thanks @rschindhelm-haba.

PR #1385, opened 2019-03-22
Update DbContext configuration doc to warn about threading concerns

### LINKED ISSUE #922: Document explicitly that DbContext is not thread-safe

mjrousos:
Mention that `DbContext` cannot be used in parallel on multiple threads and warn, specifically, about async scenarios (since tasks may run in parallel) and dependency injection scenarios (since `DbContext` services are registered with scope lifetime by default).

Fixes #922

### DIFF
--- a/entity-framework/core/miscellaneous/configuring-dbcontext.md
+++ b/entity-framework/core/miscellaneous/configuring-dbcontext.md
@@ -165,26 +165,16 @@ One code pattern that could lead to parallel operations with a single `DbContext
 
 ### Avoiding DbContext threading issues with dependency injection
 
-Another code pattern that can lead to accidental use of a `DbContext` on multiple parallel threads is retrieving `DbContext` instances from dependency injection. Registering a `DbContext` using `AddDbContext<TContext>`, as described previously, will add the `DbContext` to the app's dependency injection container with a [scoped lifetime](https://docs.microsoft.com/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2#service-lifetimes). Although a scoped lifetime is generally correct for `DbContext` objects, this can lead to threading problems if a service consuming the `DbContext` performs work in parallel. There are a few options for avoiding parallel `DbContext` usage in these scenarios:
+Another code pattern that can lead to accidental use of a `DbContext` on multiple parallel threads is retrieving `DbContext` instances from dependency injection since the `AddDbContext<TContext>` method will register the `DbContext` with a [scoped lifetime](https://docs.microsoft.com/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2#service-lifetimes) by default. Although a scoped lifetime is generally correct for `DbContext` objects, this can lead to threading problems if a service performs work in parallel within a scope.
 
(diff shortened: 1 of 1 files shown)


entity-framework/core/miscellaneous/configuring-dbcontext.md
```diff
@@ -155,6 +155,36 @@ using (var context = serviceProvider.GetService<BloggingContext>())
...
 ```
+## Avoiding DbContext threading issues
+
+Entity Framework Core does not support multiple parallel operations being run on the same `DbContext`. Because of this, it's important to use separate `DbContext` instances for operations that may execute in parallel. Not doing so could result in invalid operation exceptions.
+
+### Avoiding DbContext threading issues with async methods
+
+One code pattern that could lead to parallel operations with a single `DbContext` is the use of asynchronous extension methods. These methods enable Entity Framework Core operations to be performed in a non-blocking way, but if callers do not `await` completion of one operation before starting another, a `DbContext` could encounter issues from running more than one operation in parallel. You should be sure to `await` the completion of any asynchronous operation before beginning another with the same `DbContext`.
+
+### Avoiding DbContext threading issues with dependency injection
```
divega: IMO this whole section boils down to “don’t use DbContext in parallel”, and DI usage seems to have very little incidence.

One way DI misuse could lead to concurrency is if you made the DbContext a singleton. But that can be generalized as any pattern that leads to sharing the same DbContext instance across multiple requests. But this and singleton instances seem like they can be explained without even mentioning DI.

divega:
I looked and I think this page is indeed the best place to put this content. I would like to change it a bit to look more like the EF6 page at 
https://docs.microsoft.com/ef/ef6/fundamentals/working-with-dbcontext, which has a section about lifetime that is closely related. But I will create a new issue for that.

entity-framework/core/miscellaneous/configuring-dbcontext.md
mjrousos: I agree. I included the DI discussion, in particular, since it's one we've seen a couple customers have questions about. Do you think the doc would be better if I shortened this part up?

mjrousos:
@divega, I shortened the DI discussion. Let me know if you think it's still too long or could otherwise benefit from revision.

ajcvickers:
Some thoughts:

Agree with Diego that the first section could be much shorter. Maybe:

> The `DbContext` class is not thread-safe and must not be used concurrently by multiple threads. Concurrent access will result in undefined behavior potentially resulting in application crashes or data corruption. Note that starting a second async operation without awaiting the first can result in concurrent access and so is not safe. This includes parallel execution of multiple async operations.

With regards to D.I., the key points are:
> By default in an ASP.NET application, the context is resolved as a scoped service from the request. This is safe from current access issues because only one thread will be executing the request at a given time. Care should be taken to ensure that the context instance is not shared when using dependency injection in other scenarios or when creating new threads. This is usually achieved by either registering the context as scoped and creating scopes as needed, or by registering the context as transient.

mjrousos:
Thanks, @ajcvickers. I've made the DbContext threading issues even more concise.

divega:
Thanks a lot @mjrousos. I plan to make a few minor touches before I merge this into live, but it is a good starting point.

PR #1401, opened 2019-03-31
Add raw/interpolated string break

ajcvickers:


entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -72,6 +72,46 @@ Developers can also now control exactly when EF Core and EF Core data providers
...
+Starting with EF Core 3.0, use `FromSqlRaw`, `ExecuteSqlRaw`, and `ExecuteSqlRawAsync` to create a parameterized query where the parameters are passed separately from the query string.
+For example:
+
+```C#
+context.Products.FromSqlRaw(
+    "SELECT * FROM Products WHERE Name = {0}",
+    product.Name);
+```
+
+Use `FromSqlInterpolated`, `ExecuteSqlInterpolated`, and `ExecuteSqlInterpolatedAsync` to create a parameterized query where the parameters are passed as part of the query string.
```
divega: ... where the parameters are passed as part of an interpolated query string.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -72,6 +72,46 @@ Developers can also now control exactly when EF Core and EF Core data providers
...
+For example:
+
+```C#
+context.Products.FromSqlInterpolated(
+    "SELECT * FROM Products WHERE Name = {product.Name}");
```
divega: Missing $

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -72,6 +72,46 @@ Developers can also now control exactly when EF Core and EF Core data providers
...
+```
+
+Note that both of the queries above will produce the same parameterized SQL with the same SQL parameters.
+
+**Why**
+
+Method overloads like this make it very easy to accidentally call the raw srting method when the intent was to call the interpolated string method.
```
divega: Add ", and vice versa." (or "and the other way around.")

PR #1408, opened 2019-04-05
Update sqlite.md added tip

jecaestevez:
Added tip to show show Nuget dependencies for this sample

### DIFF
--- a/entity-framework/core/miscellaneous/testing/sqlite.md
+++ b/entity-framework/core/miscellaneous/testing/sqlite.md
@@ -43,4 +43,8 @@ The simplest way to enable testing against a different database is to modify you
 
 The key to testing with this provider is the ability to tell the context to use SQLite, and control the scope of the in-memory database. The scope of the database is controlled by opening and closing the connection. The database is scoped to the duration that the connection is open. Typically you want a clean database for each test method.
 
+>[!TIP]
+> To use the `.SqliteConnection()` extension method, reference the NuGet package `Microsoft.Data.Sqlite`.
+> To use the `.UseSqlite()` extension method, reference the NuGet package `Microsoft.EntityFrameworkCore.Sqlite`.
+
 [!code-csharp[Main](../../../../samples/core/Miscellaneous/Testing/TestProject/SQLite/BlogServiceTests.cs)]
--- a/entity-framework/core/miscellaneous/testing/sqlite.md
+++ b/entity-framework/core/miscellaneous/testing/sqlite.md
@@ -44,7 +44,7 @@ The simplest way to enable testing against a different database is to modify you
 The key to testing with this provider is the ability to tell the context to use SQLite, and control the scope of the in-memory database. The scope of the database is controlled by opening and closing the connection. The database is scoped to the duration that the connection is open. Typically you want a clean database for each test method.
 
 >[!TIP]
-> To use the `.SqliteConnection()` extension method, reference the NuGet package `Microsoft.Data.Sqlite`.
+> To use `SqliteConnection()` reference the NuGet package `Microsoft.Data.Sqlite`.
 > To use the `.UseSqlite()` extension method, reference the NuGet package `Microsoft.EntityFrameworkCore.Sqlite`.
 
 [!code-csharp[Main](../../../../samples/core/Miscellaneous/Testing/TestProject/SQLite/BlogServiceTests.cs)]


entity-framework/core/miscellaneous/testing/sqlite.md
```diff
@@ -43,4 +43,8 @@ The simplest way to enable testing against a different database is to modify you
 
 The key to testing with this provider is the ability to tell the context to use SQLite, and control the scope of the in-memory database. The scope of the database is controlled by opening and closing the connection. The database is scoped to the duration that the connection is open. Typically you want a clean database for each test method.
 
+>[!TIP]
+> To use the `.SqliteConnection()` extension method, reference the NuGet package `Microsoft.Data.Sqlite`.
```
divega: There is no SqliteConnection extension method. What you see in the snippets is calling the constructor of SqliteConnection. Also, the Microsoft.Data.Sqlite package is included when you add the Microsoft.EntityFrameworkCore.Sqlite package. No need to add it explicitly.

entity-framework/core/miscellaneous/testing/sqlite.md
```diff
@@ -43,4 +43,8 @@ The simplest way to enable testing against a different database is to modify you
...
+> To use the `.UseSqlite()` extension method, reference the NuGet package `Microsoft.EntityFrameworkCore.Sqlite`.
```
divega: We haven't agreed on a pattern for including this type of additional information relevant to the code snippets shown in the page, but I am ok with trying something for this one.

I am inclined to integrate this with the tip at the top of the page, which talks about where to find the actual sample projects, the reasons being that:

- There are several other snippets earlier in the page that depend on the package being added as a dependency, so something at the top of the page helps more. 
- That tip happens to be about the code snippets already.

cc @bricelam @ajcvickers in case they have an opinion.

jecaestevez: Hi Diego ! I found useful this other tip [ for testing in memory ](https://docs.microsoft.com/en-us/ef/core/miscellaneous/testing/in-memory#writing-tests) so I thought it will be also useful have it also  in the sqllite.md.

The reason why I suggested it  was because I was following the steps to configure in-memory.md It was useful read the tip but for sqlLite.md in my PoC test project  I need to add those Nuget so I missed that tip .

entity-framework/core/miscellaneous/testing/sqlite.md
jecaestevez: I agree it's not an extension method xD Copy&paste Documentation mistake . 
I just wanted to add the tip to share with the reader quickly a necessary package

entity-framework/core/miscellaneous/testing/sqlite.md
divega: @jecaestevez any reason you closed this?

jecaestevez: Hi Diego , I'm not very familiar contributing open source project so I  just closed because after some days I had impression the PR won't be approve and I didn't want to leave the PR open producing noise so I just closed xD

Let me know if this PR make sense and I'll very happy to open it again ;-)

divega:
It does make sense. I would like to take it with just one tweak. I will add a comment to the current version. Let me know if you can reopen.

entity-framework/core/miscellaneous/testing/sqlite.md
```diff
@@ -43,4 +43,8 @@ The simplest way to enable testing against a different database is to modify you
 
 The key to testing with this provider is the ability to tell the context to use SQLite, and control the scope of the in-memory database. The scope of the database is controlled by opening and closing the connection. The database is scoped to the duration that the connection is open. Typically you want a clean database for each test method.
 
+>[!TIP]
+> To use `SqliteConnection()` reference the NuGet package `Microsoft.Data.Sqlite`.
```
divega: I would make it simply 

> To use `SqliteConnection()` and the `.UseSqlite()` extension method, reference the NuGet package `Microsoft.EntityFrameworkCore.Sqlite`.

Although SqlIteConnection is part of the ADO.NET provider, in practice you only need to add the EF Core provider package to make both work.

jecaestevez: Thanks for the explanations! I was confused because in my example I added first the Nuget package for Microsoft.Data.Sqlite and later Microsoft.EntityFrameworkCore.Sqlite soI was wrong!

Here we have the [new pull request ](https://github.com/aspnet/EntityFramework.Docs/pull/1427)

PR #1419, opened 2019-04-10
Even more breaking changes

AndriySvyryd:


entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -681,6 +681,44 @@ modelBuilder
     .HasField("_id");
 ```
 
+## Field-only property names should match the field name
```
divega: Is there an issue we fixed with this that we can point to? 

BTW, I wasn’t aware of this breaking change and I would like to discuss reverting it. Cc @ajcvickers

AndriySvyryd: There's no separate tracking issue. It's just something I found while working on @214
We allowed to call `SetField` on a field-only or shadow property, in addition to the reasoning explained above this would change its `IdentifyingMemberInfo` or shadowness. And model builder logic assumes these values are immutable after a property has been created. We could make this work properly, but it would require an exhaustive review of every place these values are used in model building for a marginally useful feature.

divega: Can we bring it to the design meeting?

AndriySvyryd: Sure

divega: @AndriySvyryd should we get this in as is, or is it worth adding a sentence saying we will add the ability to specify a field with a different name in a latter preview?

AndriySvyryd: The main breaking change remains the same, so I don't think it's worth spoiling our plans before we release the next preview.

PR #1420, opened 2019-04-10
Add .NET universal header

rowanmiller:
This will put the same header on the EF docs that is used on the [.NET website](https://dotnet.microsoft.com/) and [main .NET docs](https://docs.microsoft.com/en-us/dotnet/).

Here is an example of this setting in use in the main .NET docs https://github.com/dotnet/docs/blob/master/docfx.json#L108.


divega:
Thanks @rowanmiller!

PR #1421, opened 2019-04-10
Add documentation for table splitting

### LINKED ISSUE #497: New feature topic: table splitting
Originally from #447.

AndriySvyryd:
Fixes #497


samples/core/Modeling/TableSplitting/DetailedOrder.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace TableSplitting
+{
+    #region DetailedOrder
+    public class DetailedOrder : Order
```
divega: DetailedOrder inherits a DetailedOrder property from Order. I wonder what happens with it.

entity-framework/core/modeling/table-splitting.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Usage
+
+Saving and querying entities using table splitting is done in the same way as other entities, the only difference is that all entities sharing a row must be tracked for the insert.
+
+[!code-csharp[Usage](../../../samples/core/Modeling/TableSplitting/Program.cs?name=Usage)]
+
+## Concurrency tokens
+
+If any of the entity types sharing a table has a concurrency token then it must be included in all other entity types to avoid a stale concurrency token value when only one of the entities mapped to the same table is updated.
```
divega: I knew entities in table splitting to share the key. When I learned about sharing the concurrency token I didn’t connect the dots that we supported entities with more overlap besides the key and the concurrency token. Should we have a sentence on how to avoid shooting oneself in the foot with this? E.g. about avoiding making conflicting changes in the same unit of work?

samples/core/Modeling/TableSplitting/DetailedOrder.cs
AndriySvyryd: Good point, I'll ignore it for clarity

entity-framework/core/modeling/table-splitting.md
AndriySvyryd: I think it would become clear once you start using it. And the exception message is straightforward.

samples/core/Modeling/TableSplitting/DetailedOrder.cs
divega: If the reason for using inheritance is to avoid repeating the properties, wouldn't a common base type (i.e. OrderData or OrderBase) that only has ID and status and isn't included in the model be a bit cleaner?

AndriySvyryd: There hasn't been an opportunity to use `HasBaseType((string)null)` in samples yet and I wanted to show that it's possible

divega: Okay

PR #1422, opened 2019-04-10
Indicate where Data Annotations can be found.

### LINKED ISSUE #1297: Where is the "RequiredAttribute" defined?
Please add assembly reference to this page. I cannot find this attribute.

---
#### Document Details

* ID: dfac6ff9-1cb7-d5d4-036a-ad2f26d299f9
* Version Independent ID: 6a3de9c8-9024-c155-94aa-7f8fc1c7bd58
* Content: [Required/optional properties - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/required-optional)
* Content Source: [entity-framework/core/modeling/required-optional.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/required-optional.md)
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

lajones:
Fixes #1297.

### DIFF
--- a/entity-framework/core/modeling/included-properties.md
+++ b/entity-framework/core/modeling/included-properties.md
@@ -19,6 +19,7 @@ You can use Data Annotations to exclude a property from the model.
 
 <!-- [!code-csharp[Main](samples/core/Modeling/DataAnnotations/Samples/IgnoreProperty.cs?highlight=6)] -->
 ``` csharp
+using System.ComponentModel.DataAnnotations.Schema;
 public class Blog
 {
     public int BlogId { get; set; }
--- a/entity-framework/core/modeling/included-properties.md
+++ b/entity-framework/core/modeling/included-properties.md
@@ -17,41 +17,10 @@ By convention, public properties with a getter and a setter will be included in 
 
 You can use Data Annotations to exclude a property from the model.
 
-<!-- [!code-csharp[Main](samples/core/Modeling/DataAnnotations/Samples/IgnoreProperty.cs?highlight=6)] -->
-``` csharp
-using System.ComponentModel.DataAnnotations.Schema;
-public class Blog
-{
-    public int BlogId { get; set; }
-    public string Url { get; set; }
-
-    [NotMapped]
-    public DateTime LoadedFromDatabase { get; set; }
-}
-```
+[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/IgnoreProperty.cs?highlight=17)]
 
 ## Fluent API
 
 You can use the Fluent API to exclude a property from the model.
 
-<!-- [!code-csharp[Main](samples/core/Modeling/FluentAPI/Samples/IgnoreProperty.cs?highlight=7,8)] -->
-``` csharp
-class MyContext : DbContext
-{
-    public DbSet<Blog> Blogs { get; set; }
-
-    protected override void OnModelCreating(ModelBuilder modelBuilder)
-    {
-        modelBuilder.Entity<Blog>()
-            .Ignore(b => b.LoadedFromDatabase);
-    }
-}
-
-public class Blog
-{
-    public int BlogId { get; set; }
-    public string Url { get; set; }
-
-    public DateTime LoadedFromDatabase { get; set; }
-}
-```
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/IgnoreProperty.cs?highlight=12,13)]
--- a/entity-framework/core/modeling/relationships.md
+++ b/entity-framework/core/modeling/relationships.md
@@ -113,7 +113,7 @@ You can use the Data Annotations to configure how navigation properties on the d
 
 To configure a relationship in the Fluent API, you start by identifying the navigation properties that make up the relationship. `HasOne` or `HasMany` identifies the navigation property on the entity type you are beginning the configuration on. You then chain a call to `WithOne` or `WithMany` to identify the inverse navigation. `HasOne`/`WithOne` are used for reference navigation properties and `HasMany`/`WithMany` are used for collection navigation properties.
 
-[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/NoForeignKey.cs?highlight=14-17)]
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/NoForeignKey.cs?highlight=14-16)]
 
 ### Single Navigation Property
 
--- a/entity-framework/core/modeling/relationships.md
+++ b/entity-framework/core/modeling/relationships.md
@@ -98,7 +98,7 @@ There are two data annotations that can be used to configure relationships, `[Fo
 
 You can use the Data Annotations to configure which property should be used as the foreign key property for a given relationship. This is typically done when the foreign key property is not discovered by convention.
 
-[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/Relationships/ForeignKey.cs?name=highlight=30)]
+[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/Relationships/ForeignKey.cs?highlight=30)]
 
 > [!TIP]  
 > The `[ForeignKey]` annotation can be placed on either navigation property in the relationship. It does not need to go on the navigation property in the dependent entity class.
--- a/entity-framework/core/modeling/relationships.md
+++ b/entity-framework/core/modeling/relationships.md
@@ -98,7 +98,7 @@ There are two data annotations that can be used to configure relationships, `[Fo
 
 You can use the Data Annotations to configure which property should be used as the foreign key property for a given relationship. This is typically done when the foreign key property is not discovered by convention.
 
-[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/Relationships/ForeignKey.cs?name=Entities&highlight=17)]
+[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/Relationships/ForeignKey.cs?name=Entities&highlight=30)]
 
 > [!TIP]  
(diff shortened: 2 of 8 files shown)


entity-framework/core/modeling/included-properties.md
```diff
@@ -19,6 +19,7 @@ You can use Data Annotations to exclude a property from the model.
 
 <!-- [!code-csharp[Main](samples/core/Modeling/DataAnnotations/Samples/IgnoreProperty.cs?highlight=6)] -->
 ``` csharp
+using System.ComponentModel.DataAnnotations.Schema;
```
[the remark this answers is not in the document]
lajones: @divega Fix 2 above looks good, but I don't understand what needs to be changed about the .cs code sample. Are you referring to "regions"? In this case the sample is small - so there is no need for a region which might exclude the using statement. We can just do:
```[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/IgnoreProperty.cs?highlight=17)]```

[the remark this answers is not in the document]
Comment (divega) [reply entity-framework/core/modeling/included-properties.md]: As long as the code that ends up being shown in the page includes the using and doesn't include too much noise, that is fine.

File: entity-framework/core/modeling/relationships.md [hunk 2]
```diff
@@ -92,13 +92,13 @@ See [Cascade Delete](../saving/cascade-delete.md) for more details about the dif
...
 
-There are two data annotations that can be used to configure relationships, `[ForeignKey]` and `[InverseProperty]`.
+There are two data annotations that can be used to configure relationships, `[ForeignKey]` and `[InverseProperty]`. These are available in the `System.ComponentModel.DataAnnotations` assembly.
 
 ### [ForeignKey]
 
 You can use the Data Annotations to configure which property should be used as the foreign key property for a given relationship. This is typically done when the foreign key property is not discovered by convention.
 
-[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/Relationships/ForeignKey.cs?name=Entities&highlight=17)]
+[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/Relationships/ForeignKey.cs?name=Entities&highlight=30)]
```
divega: This doesn't highlight the right thing or show any namespace in the preview. I think it is because the reference is using a region.

entity-framework/core/modeling/relationships.md
```diff
@@ -107,29 +107,29 @@ You can use the Data Annotations to configure which property should be used as t
 
 You can use the Data Annotations to configure how navigation properties on the dependent and principal entities pair up. This is typically done when there is more than one pair of navigation properties between two entity types.
 
-[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/Relationships/InverseProperty.cs?name=Entities&highlight=20,23)]
+[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/Samples/Relationships/InverseProperty.cs?name=Entities&highlight=33,36)]
```
divega: Same for this.

entity-framework/core/modeling/relationships.md
```diff
@@ -107,29 +107,29 @@ You can use the Data Annotations to configure which property should be used as t
...
 
 ## Fluent API
 
 To configure a relationship in the Fluent API, you start by identifying the navigation properties that make up the relationship. `HasOne` or `HasMany` identifies the navigation property on the entity type you are beginning the configuration on. You then chain a call to `WithOne` or `WithMany` to identify the inverse navigation. `HasOne`/`WithOne` are used for reference navigation properties and `HasMany`/`WithMany` are used for collection navigation properties.
 
-[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/NoForeignKey.cs?name=Model&highlight=8,9,10)]
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/NoForeignKey.cs?name=Model&highlight=14-16)]
```
divega: Same, although in this case it is highlighting the wrong thing.

entity-framework/core/modeling/relationships.md
```diff
@@ -107,29 +107,29 @@ You can use the Data Annotations to configure which property should be used as t
...
 
 ### Single Navigation Property
 
 If you only have one navigation property then there are parameterless overloads of `WithOne` and `WithMany`. This indicates that there is conceptually a reference or collection on the other end of the relationship, but there is no navigation property included in the entity class.
 
-[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/OneNavigation.cs?name=Model&highlight=10)]
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/OneNavigation.cs?name=Model&highlight=14-16)]
```
divega: Idem

entity-framework/core/modeling/relationships.md
```diff
@@ -107,29 +107,29 @@ You can use the Data Annotations to configure which property should be used as t
...
 
 ### Foreign Key
 
 You can use the Fluent API to configure which property should be used as the foreign key property for a given relationship.
 
-[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/ForeignKey.cs?name=Model&highlight=11)]
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/ForeignKey.cs?name=Model&highlight=17)]
```
divega: Idem

entity-framework/core/modeling/relationships.md
```diff
@@ -107,29 +107,29 @@ You can use the Data Annotations to configure which property should be used as t
...
 
 The following code listing shows how to configure a composite foreign key.
 
-[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/CompositeForeignKey.cs?name=Model&highlight=13)]
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/CompositeForeignKey.cs?name=Model&highlight=20)]
```
divega: Idem

entity-framework/core/modeling/relationships.md
```diff
@@ -92,13 +92,13 @@ See [Cascade Delete](../saving/cascade-delete.md) for more details about the dif
 
 ## Data Annotations
 
-There are two data annotations that can be used to configure relationships, `[ForeignKey]` and `[InverseProperty]`.
+There are two data annotations that can be used to configure relationships, `[ForeignKey]` and `[InverseProperty]`. These are available in the `System.ComponentModel.DataAnnotations` assembly.
```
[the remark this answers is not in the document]
lajones: @divega the original bug specifically mentioned that they didn't know which assembly to use. Still feel the same?

entity-framework/core/modeling/relationships.md
lajones: OK. I guess I'll have to find something other than VSCode to test with - as this worked for me. What do you use?

divega: I looked at the preview site that gets published. There is a link in the checks section, and also an email when you commit. Anyway, I am pretty sure that the issue is that name=Entity in the address means to extract the Entity region, so you have to either remove that or make sure what you want to show is in the region, and that the highlight value represents the line number inside the region.

entity-framework/core/modeling/relationships.md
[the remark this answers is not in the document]
divega: Good point, that is what he asked, but I wonder if he did it out of custom. I don’t know how the assembly reference would have help him, since that is not how you acquire it when you use EF Core. Now that I think about it from that angle, the package doesn’t help much either because it gets added automatically, unless you want to build the model into a project that doesn’t reference EF Core. Maybe the namespace is all he needed?

On the other hand, we should consider pointing to the API browser for matters like this, but unfortunately when you point to something like https://docs.microsoft.com/dotnet/api/system.componentmodel.dataannotations.requiredattribute, it defaults to .NET Framework and also shows the assembly that isn’t generally how you should acquire it or look for it.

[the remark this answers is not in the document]
ajcvickers: @lajones @divega I think it's the namespace people really want; If you're not using VS/Rider, then that can be hard to find. (Both VS and Rider will prompt you to insert the `using`, so for those IDE's even the namespace is additional clutter in the docs.)

Like @divega said, the way you acquire the assembly is by installing the EF package, which brings in the Data Annotations package automatically. So it's hard to see how knowing anything about the assembly/package would help _unless_ it is as a way to find the namespace.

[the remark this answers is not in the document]
divega: Ok, then I am convinced we just need to specify the namespace when it isn’t obvious.

entity-framework/core/modeling/included-properties.md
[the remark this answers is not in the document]
lajones: @divega Done.

entity-framework/core/modeling/relationships.md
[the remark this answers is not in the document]
lajones: Updated to refer to the namespace.

entity-framework/core/modeling/relationships.md
lajones: Yep. That was the problem. Fixed

entity-framework/core/modeling/relationships.md
lajones: Fixed as above.

entity-framework/core/modeling/relationships.md
lajones: Fixed as above.

entity-framework/core/modeling/relationships.md
lajones: Fixed as above.

entity-framework/core/modeling/relationships.md
lajones: Fixed as above.

entity-framework/core/modeling/relationships.md
lajones: Fixed as above.

Issue #984, opened 2018-09-14
Suggest change to clarify access modifier required when using Lazy Loading Proxies


geoffRGWilliams:
When combining entity type constructors with lazy-loading proxies (enabled via a call to `UseLazyLoadingProxies(true)`), model constructors must be at a minimum `protected`, or exceptions will be thrown.

Suggest changing this line:
> The constructor can be public, private, or have any other accessibility.

to:
> The constructor can be public, private, or have any other accessibility. However when using Lazy Loading Proxies the constructor must be, at minimum, protected.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ecc7aedf-6301-7ec7-4f65-81749b1f54c8
* Version Independent ID: d0774ffa-c841-e6df-72de-d7f53c0d6f6f
* Content: [Entity types with constructors - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/constructors)
* Content Source: [entity-framework/core/modeling/constructors.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/constructors.md)
* Product: **entity-framework**
* GitHub Login: @ajcvickers
* Microsoft Alias: **divega**

geoffRGWilliams:
> When combining entity type constructors with lazy-loading proxies (enabled via a call to `UseLazyLoadingProxies(true)`), model constructors must be at a minimum `protected`, or exceptions will be thrown.
> 
> Suggest changing this line:
> > The constructor can be public, private, or have any other accessibility.
> to:
> > The constructor can be public, private, or have any other accessibility. However when using Lazy Loading Proxies the constructor must be, at minimum, protected.
> 
> #### Document Details
> ⚠ _Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking._
> 
> * ID: ecc7aedf-6301-7ec7-4f65-81749b1f54c8
> * Version Independent ID: d0774ffa-c841-e6df-72de-d7f53c0d6f6f
> * Content: [Entity types with constructors - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/constructors)
> * Content Source: [entity-framework/core/modeling/constructors.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/constructors.md)
> * Product: **entity-framework**
> * GitHub Login: @ajcvickers
[...]

PR #1444, opened 2019-04-23
Resolve Issue #984 constructors.md

jackapie:
Change to clarify model constructor access modifier required when using Lazy Loading Proxies #984


divega:
@ajcvickers could you take a look?

ajcvickers:
@divega @jackapie Agree with the sentiment; maybe tweak the wording a little. Something like, "However, lazy-loading proxies require that the constructor is accessible from the inheriting proxy class. Usually this means making it either public or protected."

(Clearly, protected internal is okay too. Technically it can sometimes be internal (or protected and internal--don't remember the name for that one) when using internals-visible-to, but no need to mention that.)

divega:
Thanks @jackapie for the contribution.

PR #1448, opened 2019-04-28
Update thread-safety paragraphs to include guard exception

divega:


roji:
> Entity Framework Core does not support multiple parallel operations being run on the same `DbContext` instance.

It may be worth making the distinction between concurrent usage of the DbContext API (unsupported) and concurrent resultsets being active on the DbContext instance (supported where MARS is supported). The sentence above seems like it could be read in both ways.

Ambiguity could be reduce by saying that "the DbContext API (as well as other EF Core APIs) is not thread-safe and cannot be used concurrently on the same thread". We could also add an extra note:

> Although the DbContext API isn't thread-safe, some database providers allow multiple queries to execute simultaenously on the same DbContext. The execution of these queries and the traversal of their resultsets must still be done in a thread-safe manner.

ajcvickers:
I think I agree with @roji. On the other hand, as soon as we mention threads, then some people think it doesn't apply to multiple async operations. Maybe we should be explicit about async too?

roji:
I *think* it's somewhat standard for documentation to specify whether an API (or type) is thread-safe or not, and users are expected that understand what that means for async. But it definitely wouldn't hurt to explicitly mention the multiple async operations case which people seem to run into, and even show an example.

I could find the time in the next few days to make a proposal if you want.

ajcvickers:
> I think it's somewhat standard for documentation to specify whether an API (or type) is thread-safe or not, and users are expected that understand what that means for async. 

Agreed, this is standard. But several times recently I've heard comments along the lines of, "but that won't use a separate thread because its async" so I'm wondering if it's actually helpful to be more explicit. But then I also like @divega's original proposal for being very short and to-the-point.

Maybe:
> Entity Framework Core does not support multiple parallel operations being run on the same `DbContext` instance. This includes both parallel execution of `async` queries and any explicit concurrent use from multiple threads. Therefore, always `await` `async` calls immediately, or use separate `DbContext` instances for operations that execute in parallel.

divega:
I incorporated the last proposal from @ajcvickers. I think it looks very good. 

The other thing that I saw in the thread that I agree we could try to address is the fact that interleaving enumeration of queries _should be_ ok. The mental model to convey is that initiating the execution of query and reading each individual element of the results are all considered separate _operations_, hence whether this is supported is going to depend on the capabilities of the underlying database connection. 

However, it actually [doesn't work well](https://github.com/aspnet/EntityFrameworkCore/issues/15518) at the moment and I think it would be ok to capture it in another PR.

PR #1451, opened 2019-04-30
Fix spelling in docs

jawn:


roji:
Thanks for the fix @jawn!

PR #1472, opened 2019-05-16
Increase indent in IncludeTree sample

nphmuller:
Increase indent for ThenInclude of `author.Photo`, so it matches the other samples (see `MultipleThenIncludes`)


divega:
Thanks @nphmuller!

PR #1483, opened 2019-05-23
Fixed sentence.

cincuranet:


entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -694,7 +694,7 @@ This change was made to prevent EF Core from erroneously triggering business log
 
 **Mitigations**
 
-The pre-3.0 behavior can be restored through configuration of the property access mode in the modelBuilder fluent API.
+The pre-3.0 behavior can be restored through configuration of the property access mode in the ModelBuilder's fluent API.
```
divega: Not sure this is better. The "'s" seems unnecessary. But I feel the same about "fluent API". Maybe just:

The pre-3.0 behavior can be restored through configuration of the property access mode on ModelBuilder.

cincuranet: Works for me.

Issue #1209, opened 2018-12-30
Add more examples for relationship configurations


ghost:
Please provide an example that's similar to this example: https://github.com/aspnet/EntityFrameworkCore/issues/2517.

Creating a FK without a navigation property is a useful technique in DDD.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 24b2b50e-35f8-9e0b-8ce5-840e1ea816e6
* Version Independent ID: 9e0e6e2c-48fa-76c3-3bd9-813a6df22645
* Content: [Relationships - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/relationships)
* Content Source: [entity-framework/core/modeling/relationships.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/relationships.md)
* Product: **entity-framework**
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

Odonno:
We did it today and we had a hard time to do it (even questioning ourselves if it was even possible). After a couple of hours, we succeed to make a one-to-many relationship without specifying navigation property, only with just 1 foreign key.

And indeed, the syntax is like this: 

```cs
modelBuilder.Entity<Post>()
    .HasOne<Blog>()
    .WithMany()
    .HasForeignKey(p => p.BlogId);
```

We can even specify a lambda for the FK which is really cool.

It definitely need a section in the docs.

Odonno:
Done.

https://docs.microsoft.com/fr-fr/ef/core/modeling/relationships#without-navigation-property

PR #1503, opened 2019-06-04
feat(relationships): add "without navigation property" section

Odonno:
related to #1209

### DIFF
--- a/entity-framework/core/modeling/relationships.md
+++ b/entity-framework/core/modeling/relationships.md
@@ -135,6 +135,12 @@ You can use the string overload of `HasForeignKey(...)` to configure a shadow pr
 
 [!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/ShadowForeignKey.cs#Sample)]
 
+### Without Navigation Property
+
+You don't necessarily need to provide a navigation property. You can simply provide a Foreign Key on one side of the relationship.
+
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/NoNavigation.cs?highlight=14-17)]
+
 ### Principal Key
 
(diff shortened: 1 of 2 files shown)


samples/core/Modeling/FluentAPI/Samples/Relationships/NoNavigation.cs
```diff
@@ -0,0 +1,36 @@
...
+        public string Url { get; set; }
+    }
+
+    public class Post
+    {
+        public int PostId { get; set; }
+        public string Title { get; set; }
+        public string Content { get; set; }
+
+        public int PostForeignKey { get; set; }
```
divega: The name of the FK property is a bit strange. Can we changed to be just BlogId?

entity-framework/core/modeling/relationships.md
```diff
@@ -135,6 +135,12 @@ You can use the string overload of `HasForeignKey(...)` to configure a shadow pr
 
 [!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Samples/Relationships/ShadowForeignKey.cs#Sample)]
 
+### Without Navigation Property
```
divega: @AndriySvyryd @ajcvickers can you think of any pitfalls worth mentioning here?

ajcvickers: Looks good. The only pitfall I can think of is the obvious one--there's no navigation properties to use!

samples/core/Modeling/FluentAPI/Samples/Relationships/NoNavigation.cs
Odonno: Of course.

divega:
Thanks a lot @Odonno!

PR #1509, opened 2019-06-07
Add more metadata breaking changes.

AndriySvyryd:


entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -1133,6 +1133,28 @@ This change simplifies the implementation of the aforementioned interfaces.
 
 Use the new extension methods.
 
+## Provider-specific Metadata API changes
+
+[Tracking Issue #214](https://github.com/aspnet/EntityFrameworkCore/issues/214)
+
+This change will be introduced in EF Core 3.0-preview 6.
```
divega: Nit: I suggest we go with with present tense, which seems to me more temporally neutral, so that we don't need to change it anymore, e.g.:

This change is introduced in EF Core 3.0-preview 6.

Incidentally, I think there is a guideline for docs that recommends just using present tense.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -1133,6 +1133,28 @@ This change simplifies the implementation of the aforementioned interfaces.
...
+
+**New behavior**
+
+The provider-specific extension methods will be flattened out:
+
+* `IProperty.Relational().ColumnName` -> `IProperty.GetColumnName()`
+* `IEntityType.SqlServer().IsMemoryOptimized` -> `IEntityType.GetSqlServerIsMemoryOptimized()`
+* `PropertyBuilder.UseSqlServerIdentityColumn()` -> `PropertyBuilder ForSqlServerUseIdentityColumn()`
```
divega: Is there something missing in this example?

AndriySvyryd: Yes, these are just examples. Do you think we should list every method affected?

divega: No, sorry if I wasn't clear. I just don't understand the last call: 'PropertyBuilder ForSqlServerUseIdentityColumn()'. It looks like it is missing a dot, but perhaps it is another typo, or perhaps I don't understand what it means.

AndriySvyryd: Oh, yes, it's missing a dot

divega:
Thanks @AndriySvyryd!

PR #1514, opened 2019-06-09
Update index.md

softchris:
EF NuGet  package information is currently missing 

this adds instructions on 
- what NuGet package you need to install
- how to install EF Core 

Based on current comments by readers it seems important that not only the current Nuget package is pointed out but also instructions on how to install

### DIFF
--- a/entity-framework/core/index.md
+++ b/entity-framework/core/index.md
@@ -14,6 +14,24 @@ EF Core can serve as an object-relational mapper (O/RM), enabling .NET developer
 
 EF Core supports many database engines, see [Database Providers](providers/index.md) for details.
 
+## Inatalling Entity Framework core
+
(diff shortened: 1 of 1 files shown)


entity-framework/core/index.md
```diff
@@ -14,6 +14,24 @@ EF Core can serve as an object-relational mapper (O/RM), enabling .NET developer
 
 EF Core supports many database engines, see [Database Providers](providers/index.md) for details.
 
+## Inatalling Entity Framework core
```
divega: Typo

softchris:
I think going with 2) seems like a option, adding a link to the installation page seems like a good compromise 
I agree an overview page is an overview.

divega:
Sounds good @softchris.

PR #1532, opened 2019-06-24
Added hyperlink to 'migrations' for easy access

gurmeetsinghdke:
Hyperlink from first 'migration' keyword in the document (at very top) is missing. So added hyperlink to 'migrations' for easy access.


divega:
Thanks @gurmeetsinghdke!

PR #1536, opened 2019-06-25
Add more breaking change notes

bricelam:


entity-framework/core/providers/sqlite/limitations.md
```diff
@@ -17,6 +17,25 @@ The common relational library (shared by Entity Framework relational database pr
 * Sequences
 * Computed columns
 
+## Query limitations
+
+SQLite does not have native support for the following data types. You can read and write values for them using EF Core, and query query for equality (`where e.Property == value`); however, ordering, comparison, and other operations on them will need to be performed on the client.
```
smitpatel: query query

smitpatel: comparison other than equality?
does not equal work?

bricelam: Not equals works, but that's an equality operaiton.

Comparisons don't work. They may work within certain ranges of values, but not across the full range.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -1407,3 +1407,33 @@ These methods are used by EF to determine if a database is created but empty. Th
...
+
+Starting with EF Core 3.0, it is a DevelopmentDependency package. Which means that the dependency won't flow transitively into other projects, and that you can no longer, by default, reference its assembly.
+
+**Why**
+
+This package is only intended to be used at design time. Deployed applications shouldn't reference it. Making the package a DevelopmentDependency reinforces this recommendation.
+
+**Mitigations**
+
+If you need to reference this package to override EF Core's design-time behavior, you can update update PackageReference item metadata in your project. If the package is being referenced transitively via Microsoft.EntityFrameworkCore.Tools, you will need to add an explicit PackageReference to the package to change its metadata.
```
divega: What types do you need access to in order to override EF Core's design-time behavior. Would it be enough to move some interfaces to the abstractions package? I remember vaguely that we already discussion this, but can't remember what the answer was.

entity-framework/core/providers/sqlite/limitations.md
```diff
@@ -17,6 +17,25 @@ The common relational library (shared by Entity Framework relational database pr
 * Sequences
 * Computed columns
 
+## Query limitations
```
divega: Are you going to add a breaking change entry for this too? I think that was what we wanted :smile: cc @ajcvickers

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
bricelam: IPluralizer (if you’re not using something like my extension package), C# generators if you’re doing something pretty advanced... Nothing else I can think of

entity-framework/core/providers/sqlite/limitations.md
bricelam: I was, but I think most of it throws in 2.2, so it may not be as severe as we initially thought

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
divega: That sounds like it would be reasonable to move. IDesignTimeDbContextFactory is there too?

divega: I created https://github.com/aspnet/EntityFrameworkCore/issues/16248. We can discuss what other things are required often as application dependencies in triage.

entity-framework/core/providers/sqlite/limitations.md
divega: Okay with me.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
bricelam: IDesignTimeDbContextFactory is in Core

bricelam:
Updated and added a note about SQLitePCL.raw v2.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -1407,3 +1407,55 @@ These methods are used by EF to determine if a database is created but empty. Th
...
+
+Microsoft.EntityFrameworkCore.Sqlite previously depended on version 1.1.12 of SQLitePCL.raw.
+
+**New behavior**
+
+We've update our package to depend on version 2.0.0.
+
+**Why**
+
+Version 2.0.0 of SQLitePCL.raw targets .NET Standard 2.0. It previously targeted .NET Standard 1.1 which required a large closure transitive packages to work.
```
divega:
```suggestion
Version 2.0.0 of SQLitePCL.raw targets .NET Standard 2.0. It previously targeted .NET Standard 1.1 which required a large closure of transitive packages to work.
```

entity-framework/core/providers/sqlite/limitations.md
```diff
@@ -17,6 +17,25 @@ The common relational library (shared by Entity Framework relational database pr
...
+
+SQLite doesn't natively support the following data types. EF Core can read and write values of these types, and querying for equality (`where e.Property == value`) is also support. Other operations, however, like comparison and ordering will require evaluation on the client.
```
smitpatel: > EF Core can read and write values of these types, and querying for equality (`where e.Property == value`) is also support.

bricelam: Meh. Gotta leave contributors some improper grammar to correct. 😉

PR #1537, opened 2019-06-25
Update relationships.md

kylejrp:
Updated the arrow function variable names to be more consistent with the other functions below it.

Below, the variable `t` refers to a `Tag` entity. However, it's really should be named `pt` as it's referring to a `PostTag` entity.


divega:
Thanks @kylejrp!

PR #1552, opened 2019-07-05
Update doc on migrations

niklasholm:
- Add note about using different assemblies for the database and startup.

This information would've been very useful to our team if it was mentioned here, and would've saved us a lot of time and head scratching.


divega:
Thank you @niklasholm.

PR #1573, opened 2019-07-12
On every other page 'Data Annotations' is on top

jeanpoelie:
On every other page 'Data Annotations' option is on top, and Fluent API is on the bottom, to keep this in 1 structure I moved Fluent API down 1 section.

If possible try to tabify this where the visitor can select if they use Fleunt API or Data Annotations.

### DIFF
--- a/entity-framework/core/modeling/index.md
+++ b/entity-framework/core/modeling/index.md
@@ -14,14 +14,14 @@ This article covers configuration that can be applied to a model targeting any d
 > [!TIP]  
 > You can view this article’s [sample](https://github.com/aspnet/EntityFramework.Docs/tree/master/samples) on GitHub.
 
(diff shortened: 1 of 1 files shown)


divega:
cc @ajcvickers for his opinion.

ajcvickers:
@divega Agreed that there is non much value in changing the order, but if we do, then the wording also needs to be updated. (You could make the case that fluent should come first, since it the most full-fidelity experience.)

divega:
@jeanpoelie Ok, given that @ajcvickers and I seem to be in agreement, I will close this PR. But I want to still thank you for taking the time to suggest this change, and I encourage to try submitting a different PR for some of the things that came in this thread, like converting the data annotations, fluent API and conventions titles of the individual modeling features to use tabs.

PR #1582, opened 2019-07-15
in transactions. dont swallow exception. and rollback will happen in dispose

SimonCropp:


divega:
Thanks @SimonCropp!

Unlike https://github.com/aspnet/EntityFramework.Docs/pull/1583, I don't have any concerns with this PR. Here the old EF6 samples where incorrect, as the patter used [requires wrapping the call to Rollback() into its won try/catch block](https://docs.microsoft.com/dotnet/api/system.data.sqlclient.sqltransaction.rollback), but we didn't do it, and moving to the pattern that takes advantage of auto-rollback behavior makes the code much more terse and easy to understand. Note that we moved the EF Core samples to use this pattern, but we intentionally (so far) kept the try/catch blocks.

PR #1583, opened 2019-07-15
Remove Redundant catch

SimonCropp:
unless better advice than `// TODO: Handle failure` can be given, it would be better to let the exception bubble. leaving a todo and a swallow exception is likely to get copies as is to peoples production code

### DIFF
--- a/entity-framework/ef6/saving/transactions.md
+++ b/entity-framework/ef6/saving/transactions.md
@@ -60,27 +60,20 @@ namespace TransactionsExamples
             {
                 using (var dbContextTransaction = context.Database.BeginTransaction())
                 {
-                    try
+                    context.Database.ExecuteSqlCommand(
+                        @"UPDATE Blogs SET Rating = 5" +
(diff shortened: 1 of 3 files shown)


SimonCropp:
> Maybe we should replace the to-dos with console output or something or perhaps it is ok to get rid of it. 

i would argue "console output " is worse. unless you can express some code that handles and swallows all exceptions, then i would prefer that it bubble up, and let whatever global exception handling is in place handle it.

distantcam:
> ...the try blocks are there to illustrate auto-rollback behavior...

Then maybe mention that in the comment? That is not made clear by having an empty `try/catch`. Something like this

```csharp
catch
{
    // Rollback is automatically handled, so you don't need to handle rollback here.
    // See docs - link
    // Instead, you can just log the error as you would normally.
    throw;
}
```

SimonCropp:
ok this is clearly too controversial. so i will close

Issue #999, opened 2018-09-22
Expand Fluent API section with nameof instead of fixed strings


bartlannoeye:
While the current sample is correct, is it useful to add a second code block?

&gt; Taking it one step further, we can protect our computed query from breaking in the future due to property renames by using nameof.

+ new code block.

If this is desired, I don't mind submitting a PR.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5d48c064-a896-af4b-9a32-7846e848bb9e
* Version Independent ID: 7d134bc7-7c34-a829-2a56-9ad6ea37015c
* Content: [Computed Columns - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/relational/computed-columns)
* Content Source: [entity-framework/core/modeling/relational/computed-columns.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/relational/computed-columns.md)
* Product: **entity-framework**
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

collinstevens:
@divega may I suggest closing this issue in light of #1594?

divega:
@collinstevens Agreed. Closing this issue for the following reason:

The input to HasComputedColumnSql is a SQL expression, hence LastName and FirstName are column names, and not property names.

Although the names may match if the model is mapped by convention (and hence your proposed change would work), the names may not match if you use HasColumnName or ColumnAttribute to customize the column-to-property mappings.

Changing the code snippet to use nameof() may contribute to the confusion between properties and columns.

PR #1594, opened 2019-07-22
#999 use nameof instead of fixed strings

collinstevens:


divega:
@collinstevens thank you for taking the time to send a PR for our documentation. I have a concern with your proposed change:

The input to HasComputedColumnSql is a SQL expression, hence LastName and FirstName are column names, and not property names. 

Although the names may match if the model is mapped by convention (and hence your proposed change would work), the names may not match if you use HasColumnName or ColumnAttribute to customize the column to property mappings. 

I think your proposal may contribute to the confusion between properties and columns and I prefer to leave the code snippet the way it currently is, and I think the original issue https://github.com/aspnet/EntityFramework.Docs/issues/999 is misleading.

Perhaps what is missing is clarifying that the input to HasComputedColumnSql is a SQL expression.

collinstevens:
@divega I understand, although I can see it both ways. I would like to imagine most people are using convention over configuration and the name of the property would indeed line up with the name of the SQL column. 

I saw the open issue and thought it was a quick PR to submit, I agree it could lead to confusion if the developer discovers computed columns from this doc rather than from a SQL doc perspective.

divega:
@collinstevens unfortunately we are way behind on triage for documentation issues, but if you find something that you want to fix and that has a milestone or some label assigned, that is less likely to get push back from us.

PR #1613, opened 2019-08-01
Fix dotnet ef command sample

Eilon:
The `$` was not part of the command.


Eilon:
Indeed, copy & paste is exactly why I ran into this 😄 

Filed https://github.com/aspnet/EntityFramework.Docs/issues/1614 to find & fix other cases.

Issue #1498, opened 2019-05-31
Explain how to handle Timeout on Connection Resiliency page


AndriySvyryd:
One way to mitigate it is by enabling idle connection resiliency

divega:
Consider mentioning this in the SQL Server provider documentation instead or in addition to the general connection resiliency article. In hindsight, connection resiliency *is* a SQL Server provider topic as far as I know, so perhaps it is in the wrong place?

PR #1621, opened 2019-08-04
remove ConnectRetryCount=0 from connection strings

SimonCropp:
some of the connection strings seem to have `ConnectRetryCount=0`. i cant see why this would be the case for only a subset of connections, adn i cant see any doco in the md fiels that calls out why this should be the case. so assuming it is a copy paste issue, and submitting a PR that removes them.

if it is by design happy for this to be closed.

### DIFF
--- a/samples/core/CompiledQueries/Model/AdventureWorksContext.cs
+++ b/samples/core/CompiledQueries/Model/AdventureWorksContext.cs
@@ -10,7 +10,7 @@ namespace Samples.Model
         protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
         {
             optionsBuilder.UseSqlServer(
-                "data source=(localdb)\\mssqllocaldb;initial catalog=AdventureWorks2014;integrated security=True;MultipleActiveResultSets=True;ConnectRetryCount=0");
+                "data source=(localdb)\\mssqllocaldb;initial catalog=AdventureWorks2014;integrated security=True;MultipleActiveResultSets=True");
         }
 
         protected override void OnModelCreating(ModelBuilder modelBuilder)
--- a/samples/core/DbContextPooling/Program.cs
+++ b/samples/core/DbContextPooling/Program.cs
@@ -20,7 +20,7 @@ namespace Samples
         public static long InstanceCount;
 
         public BloggingContext(DbContextOptions options)
-            : base(options) 
+            : base(options)
             => Interlocked.Increment(ref InstanceCount);
 
         public DbSet<Blog> Blogs { get; set; }
@@ -38,7 +38,7 @@ namespace Samples
     public class Startup
     {
         private const string ConnectionString
-            = @"Server=(localdb)\mssqllocaldb;Database=Demo.ContextPooling;Integrated Security=True;ConnectRetryCount=0";
+            = @"Server=(localdb)\mssqllocaldb;Database=Demo.ContextPooling;Integrated Security=True";
 
         public void ConfigureServices(IServiceCollection services)
         {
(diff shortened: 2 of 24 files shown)


divega:
The reason is explained at https://github.com/dotnet/SqlClient/issues/29. Maybe a good idea to have the recommendation explicitly in our docs, but perhaps it will become obsolete if SqlClient is improved. 

@ajcvickers thoughts on this?

SimonCropp:
@divega if thats the case, then shouldnt it be applied consistently? eg https://github.com/aspnet/EntityFramework.Docs/blob/master/samples/core/GetStarted/AspNetCore/EFGetStarted.AspNetCore.ExistingDb/Models/BloggingContext.cs#L26

divega:
@SimonCropp we could go in either direction (depending on whether the 10 seconds delay is still an issue, IIRC, Arthur mentioned recently he has seen the behavior change). And we can always try to make things more consistent, it is just work that competes with lots other improvements we can make on the docs.

ajcvickers:
@divega @SimonCropp I always use it, mostly because without it I always got the 10 second delay in doing database existence checks. I always include in code samples I post.

That being said, we found some inconsistent behavior where sometimes I would not get the delay even without this. If we can validate that the issue has been fixed and is available everywhere people will use it, then we can start showing connection strings without it. I don't think we're there yet, so I would err on the side of using it everywhere for now.

divega:
FWIW, my (handwavy) theory of how this could have been fixed in SqlClient without implementing the override we asked for in https://github.com/dotnet/SqlClient/issues/29, is that if a connection attempt is made on a server that exists but with a database that doesn't exists, the server is able to respond immediately that the database doesn't exists, and makes sense to fail fast instead of retrying. Idle connection resiliency should only retry if the communication with the server fails. .

I will try to find some time to test this assumption myself in the next few days, to see if I can identify a difference across different versions of SqlClient and SQL Server, but @SimonCropp if you want to do this yourself, feel free.

divega:
I had the chance to do some testing around this, and I confirmed that the issue still exists. Given that, I would prefer our samples for SQL Server to have `ConnectRetryCount=0` on the connection string to save 10 seconds of startup every time there is a database existence check.

I think it would be nice to document why we recommend `ConnectRetryCount=0` in most cases (#1498 covers this as well as when it is useful to configure `ConnectRetryCount` to some other value). The closest fit seems to be the connection resiliency page (there is one for each of EF Core and EF6), but that wouldn't have much reach.

On the other hand, I don't think it is a high priority to make all existing code samples consistent.

@SimonCropp let us know what you would like to do.

SimonCropp:
my preference would be for, when EF is checking for DB existence, it automatically adds that `ConnectRetryCount=0` option to the connection.

ajcvickers:
@SimonCropp Modification of the connection string has many issues; we're not going to do that.

SimonCropp:
@ajcvickers how about a warning in the log if ConnectRetryCount is missing? or a warning if an existence check takes more than 9 seconds.

ajcvickers:
@SimonCropp Possible we could warn in some way; not sure it is high-value.

divega:
@SimonCropp The main reason I agree with @ajcvickers is that we have proof that many of our customers never look at the warnings we issue.

Besides that, if we to add a warning, we should avoid an approach that somehow parses the connection string on every use. That could become a perf bottleneck in high rps  scenarios, especially using ConnectionStringBuilder (we have hit this in the past).

Another through: idle connection resiliency is not a useless feature. It makes sense to have it enabled if you are going to need it. That is why there is no substitute for it being fixed or enhanced in SqlClient. 

All that said, if you are willing to try sending a PR to detect existence checks that take too long and produce a warning for SQL Server, that is something that I believe we would consider. Care to create a product bug for this?

PR #1630, opened 2019-08-09
Add breaking changes

ajcvickers:
* .NET Standard 2.1
* No identity resolution in no-tracking queries

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
+++ b/entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
@@ -20,6 +20,7 @@ Breaks in new features introduced from one 3.0 preview to another 3.0 preview ar
 | **Breaking change**                                                                                               | **Impact** |
 |:------------------------------------------------------------------------------------------------------------------|------------|
 | [LINQ queries are no longer evaluated on the client](#linq-queries-are-no-longer-evaluated-on-the-client)         | High       |
+| [EF Core 3.0 targets .NET Standard 2.1 rather than .NET Standard 2.0](#netstandard21) | High      |
 | [The EF Core command-line tool, dotnet ef, is no longer part of the .NET Core SDK](#dotnet-ef) | High      |
 | [FromSql, ExecuteSql, and ExecuteSqlAsync have been renamed](#fromsql) | High      |
 | [Query types are consolidated with entity types](#qt) | High      |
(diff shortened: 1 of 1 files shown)


entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -219,6 +244,29 @@ Specifying `FromSql` anywhere other than on a `DbSet` had no added meaning or ad
...
+
+This change is introduced in EF Core 3.0-preview 6.
+
+**Old behavior**
+
+Before EF Core 3.0, the same entity instance would be used for every occurrence of an entity with a given type and ID. This matches the behavior of tracking queries.
+
+**New behavior**
+
+Starting with EF Core 3.0, different entity instances will be created when an entity with a given type and ID is encountered at different places in a query tree created using  `Include` statements. This matches the behavior of EF6.
```
divega: NIT: I am not sure that the mention of a query tree or `Include` helps here. IMO those are kind of orthogonal as an entity in the same "place" of the query tree is returned twice (e.g. for another "row" of the results) it will be a different instance. Also, this is not just for include but also for projected entities and collections. 

I think mentioning that if the same entity (i.e. same type and identity) is referenced multiple times in query results, then it will be a different instance each time. An example can help too, e.g. a query that returns Products and Categories will return a different Category for each product.

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -98,6 +100,29 @@ Besides this, automatic client evaluation can lead to issues in which improving
 
 If a query can't be fully translated, then either rewrite the query in a form that can be translated, or use `AsEnumerable()`, `ToList()`, or similar to explicitly bring data back to the client where it can then be further processed using LINQ-to-Objects.
 
+<a name="netstandard21"></a>
+### EF Core 3.0 targets .NET Standard 2.1 rather than .NET Standard 2.0
```
divega: I am baffled that we didn't have this yet :man_facepalming:

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
ajcvickers: Updated

PR #1633, opened 2019-08-12
Fix call to GetService<IMigrator>()

jannikbuschke:
Using EF Core 2.2 the code sample to get an IMigrator interface `DbContext.GetService<IMigrator>()` doesnt work.


divega:
Thanks @jannikbuschke!

PR #1634, opened 2019-08-12
Update index.md

grinay:
Add redis distributed cache extension for EF core

### DIFF
--- a/entity-framework/core/extensions/index.md
+++ b/entity-framework/core/extensions/index.md
@@ -54,6 +54,15 @@ Entity Framework Core Generator (efg) is a .NET Core CLI tool that can generate 
 
 ## Extensions
 
+### Microsoft.EntityFrameworkCore.DistributedCache
+An extension for caching result of queries into redis cache.
+It implement as async and sync version. 
+Support distributed lock to be able use cache between n-numbers of servers which
+serve application, and domn’t allow any concurrency race. 
+Also you can disable distributed lock if you use only one server
+
+[GitHub repository](https://github.com/grinay/Microsoft.EntityFrameworkCore.DistributedCache/)
+
 ### Microsoft.EntityFrameworkCore.AutoHistory
 
 A plugin library that enables automatically recording the data changes performed by EF Core into a history table.


entity-framework/core/extensions/index.md
```diff
@@ -54,6 +54,15 @@ Entity Framework Core Generator (efg) is a .NET Core CLI tool that can generate
 
 ## Extensions
 
+### Microsoft.EntityFrameworkCore.DistributedCache
+An extension for caching result of queries into redis cache.
```
ErikEJ: Please fix the broken English (or let me know if you need help with that)

grinay: Hello. I did. Please correct me if I did any mistake. Thanks.

entity-framework/core/extensions/index.md
```diff
@@ -54,6 +54,14 @@ Entity Framework Core Generator (efg) is a .NET Core CLI tool that can generate
 
 ## Extensions
 
+### Microsoft.EntityFrameworkCore.DistributedCache
+An extension for caching results of queries into the Redis cache.
+This library implements asynchronous and synchronous execution.
+It supports distributed lock to be able to use cache between n-numbers of servers which runs the application and doesn’t allow any race condition for cache update.
```
divega: I believe this is easier to read, but I hope I am not changing the meaning of the sentence. @grinay what do you think?
```suggestion
It supports distributed lock to be able to use cache across multiple server which run the application and avoid race conditions on cache update.
```

entity-framework/core/extensions/index.md
```diff
@@ -54,6 +54,14 @@ Entity Framework Core Generator (efg) is a .NET Core CLI tool that can generate
 
 ## Extensions
 
+### Microsoft.EntityFrameworkCore.DistributedCache
```
divega: This isn't feedback on the PR, but on the name of the package:

- It shouldn't really start with "Microsoft." or contain "Microsoft"in the name unless the package is from Microsoft. For example, we recommend for example, provider writer, to use "Company.EntityFrameworkCore.Database"  for the provider packages. I know there are some in this page that don't follow this rule. That is because it wasn't enforced before. 
- I noticed that in the actual package you use a prefix "Extention". Is that the name of the organization that creates it or a just "Extension" misspelled?

entity-framework/core/extensions/index.md
grinay: That's great. Thanks for correcting my English.

PR #1647, opened 2019-08-17
HasForeignKey bug

noctwolf:


noctwolf:
```C#
modelBuilder.Entity<Order>()
                .ToTable("Orders")
                .HasOne(o => o.DetailedOrder).WithOne()
                .HasForeignKey<Order>(o => o.Id);
```
add
···C#
entity.Property(f => f.Id).HasColumnName("IdFoo");
···
change to
```C#
            modelBuilder.Entity<Order>(entity =>
            {
                entity.ToTable("Orders")
                    .HasOne(o => o.DetailedOrder).WithOne()
                    .HasForeignKey<Order>(o => o.Id);

                entity.Property(f => f.Id).HasColumnName("IdFoo");
            });
···
throw exception
···C#
System.InvalidOperationException
  HResult=0x80131509
  Message=The properties 'DetailedOrder.Id', 'Order.Id' are configured to use 'Identity' value generator and are mapped to the same table 'Orders'. Only one column per table can be configured as 'Identity'. Call 'ValueGeneratedNever' for properties that should not use 'Identity'.
  Source=Microsoft.EntityFrameworkCore.SqlServer
···

**Comment by noctwolf** [comment]
```C#
.HasForeignKey<Order>(o => o.Id);
```
change to
```C#
.HasForeignKey<DetailedOrder>(o => o.Id);
```
work

divega:
Thanks @noctwolf. I have merged your proposed change because I think the model that gets generated with it is closer to the original intent: DetailedOrder is the dependent and Order is the principal. 

I would like to clarify that although the model generated by the sample as it currently stands is a bit awkward, it works. The PK property of the dependent also acts as the FK into the principal, and because it is a table splitting (i.e. both entities are mapped to share the same table) example, both entity keys actually map to the same key column on the table. If you configure an FK explicitly to map to a different column, then you are drifting off that scenario.

PR #1651, opened 2019-08-19
Update features.md

OOberoi:
Added [Owned] attribute to OrderDetails class such that it is treated as an owned entity.
Added missing punctuation to the Entities section.

I feel, the entities section would be better served if it is slightly refactored (in my humble opinion).


divega:
Thanks @OOberoi! Feel free to suggest how you would refactor it, preferable in a new issue or PR.

OOberoi:
Thank you @divega. I’ll be happy to!

PR #1666, opened 2019-08-22
C113614: Changing italic asterisks by italic underscore

v-maudel:
Description: Reporting source issues as a result of nested asterisks (i.e. \*\*\* instead of \*\*\_) that are breaking code in localized sites.


v-maudel:
Hi @divega, can you please help us to merge this PR?
Thanks in advance.

divega:
@v-maudel, please let me know if this way of fixing doesn't help with localization.

Issue #1132, opened 2018-11-21
Raw SQL documentation doesn't answer "what is FromSql()" and isn't very clear


shiraNadjari:
i have seen the the raw sql queries articale in microsoft and couldn't understand what Blogs is.
is it the database itself? beacause i tryed and it didnt work.  

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7e1289d5-8364-18a4-3445-6bba05c6854b
* Version Independent ID: 2680e647-f64a-7991-f821-3b5e2a845960
* Content: [Raw SQL Queries - EF Core](https://docs.microsoft.com/en-us/ef/core/querying/raw-sql#including-related-data)
* Content Source: [entity-framework/core/querying/raw-sql.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/querying/raw-sql.md)
* Product: **entity-framework**
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

divega:
The FromSql API allows you to query for entity or query type instances using your own SELECT statements or calls to stored procedures. 

It is somewhat explained in the text of the article, but I get from your post that it isn't clear or explicit enough. 

I also believe the article is difficult to understand due to the Limitations section being the first section.

divega:
@weitzhandler the right API to invoke a stored procedure (in this case it is not a function) that doesn't return a result set is dbContext.Database.ExecuteSqlCommand().

ajcvickers:
Duplicate of #665

PR #1673, opened 2019-08-26
Document new FromSql{Raw,Interpolated} methods

### LINKED ISSUE #1667: Why the example subject to SQL Injection attack?
One of the examples has FromSql($"SELECT * FROM dbo.SearchBlogs({searchTerm})")
There is a warning not to do this
Why give an example and then warn not to use it?

---
#### Document Details

* ID: 7e1289d5-8364-18a4-3445-6bba05c6854b
* Version Independent ID: 2680e647-f64a-7991-f821-3b5e2a845960
* Content: [Raw SQL Queries - EF Core](https://docs.microsoft.com/en-us/ef/core/querying/raw-sql#feedback)
* Content Source: [entity-framework/core/querying/raw-sql.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/querying/raw-sql.md)
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

roji:
Closes #1667

### DIFF
--- a/entity-framework/core/querying/raw-sql.md
+++ b/entity-framework/core/querying/raw-sql.md
@@ -7,19 +7,19 @@ uid: core/querying/raw-sql
 ---
 # Raw SQL Queries
 
-Entity Framework Core allows you to drop down to raw SQL queries when working with a relational database. This can be useful if the query you want to perform can't be expressed using LINQ, or if using a LINQ query is resulting in inefficient SQL queries. Raw SQL queries can return entity types or, starting with EF Core 2.1, [keyless entity types](xref:core/modeling/keyless-entity-types) that are part of your model.
+Entity Framework Core allows you to drop down to raw SQL queries when working with a relational database. This can be useful if the query you want to perform can't be expressed using LINQ, or if using a LINQ query is resulting in an inefficient SQL query. Raw SQL queries can return regular entity types or [keyless entity types](xref:core/modeling/keyless-entity-types) that are part of your model.
 
 > [!TIP]  
-> You can view this article's [sample](https://github.com/aspnet/EntityFramework.Docs/tree/master/samples/core/Querying) on GitHub.
+> You can view this article's [sample](https://github.com/aspnet/EntityFramework.Docs/tree/master/samples/core/Querying/Querying/RawSQL/Sample.cs) on GitHub.
 
 ## Basic raw SQL queries
 
-You can use the *FromSql* extension method to begin a LINQ query based on a raw SQL query.
+You can use the `FromSqlRaw` extension method to begin a LINQ query based on a raw SQL query.
 
 <!-- [!code-csharp[Main](samples/core/Querying/RawSQL/Sample.cs)] -->
 ``` csharp
 var blogs = context.Blogs
-    .FromSql("SELECT * FROM dbo.Blogs")
+    .FromSqlRaw("SELECT * FROM dbo.Blogs")
     .ToList();
 ```
 
@@ -28,44 +28,53 @@ Raw SQL queries can be used to execute a stored procedure.
 <!-- [!code-csharp[Main](samples/core/Querying/RawSQL/Sample.cs)] -->
 ``` csharp
 var blogs = context.Blogs
-    .FromSql("EXECUTE dbo.GetMostPopularBlogs")
+    .FromSqlRaw("EXECUTE dbo.GetMostPopularBlogs")
     .ToList();
 ```
 
 ## Passing parameters
 
-As with any API that accepts SQL, it is important to parameterize any user input to protect against a SQL injection attack. You can include parameter placeholders in the SQL query string and then supply parameter values as additional arguments. Any parameter values you supply will automatically be converted to a `DbParameter`.
+> [!WARNING]
+> **Always use parameterization for raw SQL queries**
+>
+> When introducing any user-provided values into a raw SQL query, care must be taken to avoid SQL injection attacks. In addition to validating that such values don't contain invalid characters, always use parameterization which sends the values separate from the SQL text.
+>
+> In particular, never pass a concatenated or interpolated string (`$""`) with unvalidated user-provided values into `FromSqlRaw` or `ExecuteSqlRaw`. The `FromSqlInterpolated` and `ExecuteSqlInterpolated` methods allow using string interpolation syntax in a way that protects against SQL injection attacks.
 
-The following example passes a single parameter to a stored procedure. While this may look like `String.Format` syntax, the supplied value is wrapped in a parameter and the generated parameter name inserted where the `{0}` placeholder was specified.
-
-<!-- [!code-csharp[Main](samples/core/Querying/RawSQL/Sample.cs)] -->
-``` csharp
-var user = "johndoe";
-
-var blogs = context.Blogs
-    .FromSql("EXECUTE dbo.GetMostPopularBlogsForUser {0}", user)
-    .ToList();
-```
-
-This is the same query but using string interpolation syntax, which is supported in EF Core 2.0 and above:
+The following example passes a single parameter to a stored procedure by including a parameter placeholder in the SQL query string and providing an additional argument. While this may look like `String.Format` syntax, the supplied value is wrapped in a `DbParameter` and the generated parameter name inserted where the `{0}` placeholder was specified.
 
 <!-- [!code-csharp[Main](samples/core/Querying/RawSQL/Sample.cs)] -->
 ``` csharp
 var user = "johndoe";
 
 var blogs = context.Blogs
-    .FromSql($"EXECUTE dbo.GetMostPopularBlogsForUser {user}")
+    .FromSqlRaw("EXECUTE dbo.GetMostPopularBlogsForUser {0}", user)
     .ToList();
 ```
 
-You can also construct a DbParameter and supply it as a parameter value:
+As an alternative to `FromSqlRaw`, you can use `FromSqlInterpolated` which allows the safe use of string interpolation. As with the previous example, the value is converted to a `DbParameter` and is therefore not vulnerable to SQL injection:
+
+> [!NOTE]
+> Prior to version 3.0, `FromSqlRaw` and `FromSqlInterpolated` were two overloads named `FromSql`. See the [previous versions section](#previous-versions) for more details.
+
+
+<!-- [!code-csharp[Main](samples/core/Querying/RawSQL/Sample.cs)] -->
+``` csharp
+var user = "johndoe";
+
+var blogs = context.Blogs
+    .FromSqlInterpolated($"EXECUTE dbo.GetMostPopularBlogsForUser {user}")
+    .ToList();
+```
+
+You can also construct a DbParameter and supply it as a parameter value. Since a regular SQL parameter placeholder is used, rather than a string placeholder, `FromSqlRaw` can be safely used:
 
 <!-- [!code-csharp[Main](samples/core/Querying/RawSQL/Sample.cs)] -->
 ``` csharp
 var user = new SqlParameter("user", "johndoe");
 
 var blogs = context.Blogs
-    .FromSql("EXECUTE dbo.GetMostPopularBlogsForUser @user", user)
+    .FromSqlRaw("EXECUTE dbo.GetMostPopularBlogsForUser @user", user)
     .ToList();
 ```
 
@@ -76,7 +85,7 @@ This allows you to use named parameters in the SQL query string, which is useful
 var user = new SqlParameter("user", "johndoe");
 
 var blogs = context.Blogs
-    .FromSql("EXECUTE dbo.GetMostPopularBlogs @filterByUser=@user", user)
+    .FromSqlRaw("EXECUTE dbo.GetMostPopularBlogs @filterByUser=@user", user)
     .ToList();
 ```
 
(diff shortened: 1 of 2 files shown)


entity-framework/core/querying/raw-sql.md
```diff
@@ -28,13 +28,16 @@ Raw SQL queries can be used to execute a stored procedure.
...
+As with any API that accepts SQL, it is important to parameterize any user input to protect against a SQL injection attack. `FromSqlRaw` performs simple string interpolation, so any parameters will be included as-is and must be manually sanitized; not doing so may expose your application to major security risks. To safely pass parameters, `FromSqlInterpolated` allows you to include placeholders in the SQL query string and then supply parameter values as additional arguments. Any parameter values you supply will automatically be converted to a `DbParameter` for safe execution. Using parameters with `FromSqlRaw` can still useful for dynamically building your SQL query, but should be used with care and never when simple parameterization is needed.
+
+> [!WARNING]
+> Prior to version 3.0, `FromSqlRaw` and `FromSqlInterpolated` were two overloads named `FromSql`. These two overloads have been obsoleted for safety reasons.
```
divega: More than including a warning here, can we describe the original `FromSql` under a "using previous versions" subtitle and have a warning explaining briefly in which cases it was ambiguous/hard to predict what things would get parameterized?

entity-framework/core/querying/raw-sql.md
```diff
@@ -43,29 +46,29 @@ The following example passes a single parameter to a stored procedure. While thi
 var user = "johndoe";
 
 var blogs = context.Blogs
-    .FromSql("EXECUTE dbo.GetMostPopularBlogsForUser {0}", user)
+    .FromSqlInterpolated("EXECUTE dbo.GetMostPopularBlogsForUser {0}", user)
```
divega: Here the correct method is FromSqlRaw.

entity-framework/core/querying/raw-sql.md
```diff
@@ -43,29 +46,29 @@ The following example passes a single parameter to a stored procedure. While thi
...
     .ToList();
 ```
 
-This is the same query but using string interpolation syntax, which is supported in EF Core 2.0 and above:
+This is the same query but using named interpolation syntax:
```
divega: I am not familiar with the term "named interpolation syntax". Can you provide some references? Otherwise, I suggest:
```suggestion
This is the same query but using string interpolation syntax and the `FromSqlInterpolated` method:
```

entity-framework/core/querying/raw-sql.md
```diff
@@ -43,29 +46,29 @@ The following example passes a single parameter to a stored procedure. While thi
...
 var user = "johndoe";
 
 var blogs = context.Blogs
-    .FromSql($"EXECUTE dbo.GetMostPopularBlogsForUser {user}")
+    .FromSqlInterpolated($"EXECUTE dbo.GetMostPopularBlogsForUser {user}")
     .ToList();
 ```
 
-You can also construct a DbParameter and supply it as a parameter value:
+You can also construct a DbParameter and supply it as a parameter value. Since a regular SQL parameter placeholder is used, rather than a string placeholder, `FromSqlRaw` can be safely used:
```
divega: I don't get why this would be considered a reason it is safe to use `FromSqlRaw`.

entity-framework/core/querying/raw-sql.md
```diff
@@ -137,14 +140,9 @@ There are a few limitations to be aware of when using raw SQL queries:
...
-
-> [!WARNING]  
-> **Always use parameterization for raw SQL queries:** In addition to validating user input, always use parameterization for any values used in a raw SQL query/command. APIs that accept a raw SQL string such as `FromSql` and `ExecuteSqlCommand` allow values to be easily passed as parameters. Overloads of `FromSql` and `ExecuteSqlCommand` that accept FormattableString also allow using string interpolation syntax in a way that helps protect against SQL injection attacks. 
-> 
-> If you are using string concatenation or interpolation to dynamically build any part of the query string, or passing user input to statements or stored procedures that can execute those inputs as dynamic SQL, then you are responsible for validating any input to protect against SQL injection attacks.
+* SQL statements other than `SELECT` are recognized automatically as non-composable. As a consequence, the full results of stored procedures are always returned to the client and any LINQ operators applied after `FromSql` methods are evaluated in-memory.
```
divega: AFAIR, the plan was that in 3.0 we were going to remove any heuristics to try to distinguish composable from non-composable queries. The latter should lead to incorrect SQL now. Did this plan change? 

cc @smitpatel

divega: The warning here is an important security note for EF Core. Why remove it? I think we should consider moving it to the top of the page instead, after making any adjustments necessary to it.

samples/core/Querying/Querying/RawSQL/Sample.cs
```diff
@@ -27,7 +27,7 @@ public static void Run()
                 var user = "johndoe";
 
                 var blogs = context.Blogs
-                    .FromSql("EXECUTE dbo.GetMostPopularBlogsForUser {0}", user)
+                    .FromSqlInterpolated("EXECUTE dbo.GetMostPopularBlogsForUser {0}", user)
```
divega: This call needs to change if we want to turn it into an example for FromSqlInterpolated.

divega:
@roji #1132 may also be easy to address.

entity-framework/core/querying/raw-sql.md
```diff
@@ -137,14 +140,9 @@ There are a few limitations to be aware of when using raw SQL queries:
 
 * The SQL query cannot contain related data. However, in many cases you can compose on top of the query using the `Include` operator to return related data (see [Including related data](#including-related-data)).
 
-* `SELECT` statements passed to this method should generally be composable: If EF Core needs to evaluate additional query operators on the server (for example, to translate LINQ operators applied after `FromSql`), the supplied SQL will be treated as a subquery. This means that the SQL passed should not contain any characters or options that are not valid on a subquery, such as:
+* `SELECT` statements passed to this method should generally be composable: If EF Core needs to evaluate additional query operators on the server (for example, to translate LINQ operators applied after `FromSql` methods), the supplied SQL will be treated as a subquery. This means that the SQL passed should not contain any characters or options that are not valid on a subquery, such as:
```
smitpatel: Line 137: The SQL query must return data for all properties of the entity or query type.

no query types

entity-framework/core/querying/raw-sql.md
smitpatel: The plan did not changed. The statement is wrong.

entity-framework/core/querying/raw-sql.md
```diff
@@ -28,13 +28,16 @@ Raw SQL queries can be used to execute a stored procedure.
...
+    .FromSqlRaw("EXECUTE dbo.GetMostPopularBlogs")
     .ToList();
 ```
 
 ## Passing parameters
 
-As with any API that accepts SQL, it is important to parameterize any user input to protect against a SQL injection attack. You can include parameter placeholders in the SQL query string and then supply parameter values as additional arguments. Any parameter values you supply will automatically be converted to a `DbParameter`.
+As with any API that accepts SQL, it is important to parameterize any user input to protect against a SQL injection attack. `FromSqlRaw` performs simple string interpolation, so any parameters will be included as-is and must be manually sanitized; not doing so may expose your application to major security risks. To safely pass parameters, `FromSqlInterpolated` allows you to include placeholders in the SQL query string and then supply parameter values as additional arguments. Any parameter values you supply will automatically be converted to a `DbParameter` for safe execution. Using parameters with `FromSqlRaw` can still useful for dynamically building your SQL query, but should be used with care and never when simple parameterization is needed.
```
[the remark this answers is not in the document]
roji: > For me, FromSqlRaw performs not interpolation at all, but still allows you to "use parameters".

I agree that given our naming choice, talking about interpolation in the context of `FromSqlRaw` is not a good idea. However, note that the term "string interpolation" [is used in general](https://en.wikipedia.org/wiki/String_interpolation) to refer to any case where a string template is converted to a literal by integrating "interpolatable expressions" into it. [The C# language reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) refers to $-strings as interpolated strings, which can be *converted* to FormattableString (but don't have to be). FWIW when we were discussing naming I pointed out that calling the API FromSqlInterpolated isn't ideal, and something like FromSqlParameterized would be more accurate. But that's all in the past - I'll change the text to better fit our naming (and remove the word "parameter" from any context that doesn't include FormattableString).

entity-framework/core/querying/raw-sql.md
roji: Will do.

entity-framework/core/querying/raw-sql.md
roji: Sorry. Have replaced the composability comment with one specifically addressing SQL Server stored procedures and client evaluation.

roji: > The warning here is an important security note for EF Core. Why remove it? I think we should consider moving it to the top of the page instead, after making any adjustments necessary to it.

Have brought the warning back, updated it and moved it to the top of the "Passing Parameters" section.

entity-framework/core/querying/raw-sql.md
```diff
@@ -131,20 +140,19 @@ var blogs = context.Blogs
...
 
-> [!WARNING]  
-> **Always use parameterization for raw SQL queries:** In addition to validating user input, always use parameterization for any values used in a raw SQL query/command. APIs that accept a raw SQL string such as `FromSql` and `ExecuteSqlCommand` allow values to be easily passed as parameters. Overloads of `FromSql` and `ExecuteSqlCommand` that accept FormattableString also allow using string interpolation syntax in a way that helps protect against SQL injection attacks. 
-> 
-> If you are using string concatenation or interpolation to dynamically build any part of the query string, or passing user input to statements or stored procedures that can execute those inputs as dynamic SQL, then you are responsible for validating any input to protect against SQL injection attacks.
+EF Core version 2.2 and earlier had two overloads named `FromSql` which behaved in the same way as the newer `FromSqlRaw` and `FromSqlInterpolated`. This made it very easy to accidentally call the raw string method when the intent was to call the interpolated string method, and the other way around. This could result in queries not being parameterized when they should have been.
```
roji: I've included this relatively short note. @divega do you think this is sufficient or would you like to see examples of possible mistakes? I'm not sure the history/reasoning is relevant here.

roji:
Thanks for the reviews @divega and @smitpatel, I realized that my mental model of FromSql wasn't very accurate. Have rewritten the change, hopefully things are better now.

entity-framework/core/querying/raw-sql.md
```diff
@@ -10,16 +10,16 @@ uid: core/querying/raw-sql
 Entity Framework Core allows you to drop down to raw SQL queries when working with a relational database. This can be useful if the query you want to perform can't be expressed using LINQ, or if using a LINQ query is resulting in inefficient SQL queries. Raw SQL queries can return entity types or, starting with EF Core 2.1, [query types](xref:core/modeling/query-types) that are part of your model.
```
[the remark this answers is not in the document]
roji: Will hold off until the query types section itself is modified to keyless entity type (link has to be updated).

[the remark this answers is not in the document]
divega: The PR that renames query types to keyless entity types is out.
```suggestion
Entity Framework Core allows you to drop down to raw SQL queries when working with a relational database. This can be useful if the query you want to perform can't be expressed using LINQ, or if using a LINQ query is resulting in inefficient SQL queries. Raw SQL queries can be used to return any entities that are part of your model, including regular entities and [keyless entitties](xref:core/modeling/query-types).
```

entity-framework/core/querying/raw-sql.md
divega: I personally like it like this. Thanks.

entity-framework/core/querying/raw-sql.md
```diff
@@ -91,38 +100,38 @@ The following example uses a raw SQL query that selects from a Table-Valued Func
 var searchTerm = ".NET";
```
smitpatel: Should also show generated SQL since composition.

entity-framework/core/querying/raw-sql.md
```diff
@@ -91,38 +100,38 @@ The following example uses a raw SQL query that selects from a Table-Valued Func
...
-    .FromSql($"SELECT * FROM dbo.SearchBlogs({searchTerm})")
+    .FromSqlInterpolated($"SELECT * FROM dbo.SearchBlogs({searchTerm})")
     .AsNoTracking()
     .ToList();
 ```
 
 ## Including related data
 
-The `Include()` method can be used to include related data, just like with any other LINQ query:
+The `Include` method can be used to include related data, just like with any other LINQ query:
```
smitpatel: Should also specify that doing Include requires composition to be enabled.

entity-framework/core/querying/raw-sql.md
```diff
@@ -7,19 +7,19 @@ uid: core/querying/raw-sql
 ---
 # Raw SQL Queries
 
-Entity Framework Core allows you to drop down to raw SQL queries when working with a relational database. This can be useful if the query you want to perform can't be expressed using LINQ, or if using a LINQ query is resulting in inefficient SQL queries. Raw SQL queries can return entity types or, starting with EF Core 2.1, [query types](xref:core/modeling/query-types) that are part of your model.
+Entity Framework Core allows you to drop down to raw SQL queries when working with a relational database. This can be useful if the query you want to perform can't be expressed using LINQ, or if using a LINQ query is resulting in inefficient SQL queries. Raw SQL queries can be used to return any entities that are part of your model, including regular entities and [keyless entitties](xref:core/modeling/query-types).
```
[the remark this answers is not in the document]
roji: > I understand entity as the instance and entitytype as the object's type.

I also understand it this way, but I think it makes sense to specify the type here, as we're discussing the type of the type (regular vs. keyless). A bit like how we say that a C# method return value or reference types (and not necessarily value or reference values).

roji:
Have taken comments into account and merged. There's definitely still stuff to be improved here, we can always do it later.

entity-framework/core/querying/raw-sql.md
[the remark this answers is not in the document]
divega: The page also got renamed.

entity-framework/core/querying/raw-sql.md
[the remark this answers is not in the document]
divega: Also renamed page here.

PR #1710, opened 2019-09-09
Silence some author-missing suggestions

roji:


AndriySvyryd:
Depends on answer in https://github.com/aspnet/EntityFramework.Docs/issues/1262, but don't put `rowanmiller` or `anpete` there

PR #1713, opened 2019-09-09
C113612: Adding two spaces at the end of list items to avoid issues in live pages

v-maudel:
Description: missing spaces at the end of the reported strings created create content misalignment. Kindly consider adding 2 spaces as shown below.

### DIFF
--- a/entity-framework/ef6/fundamentals/async.md
+++ b/entity-framework/ef6/fundamentals/async.md
@@ -215,14 +215,14 @@ For a comprehensive list of available extension methods in the System.Data.Entit
 
 Now that the code is asynchronous, we can observe a different execution flow when we run the program:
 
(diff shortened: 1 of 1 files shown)


v-maudel:
Hi @divega, can you please help us by merging this PR?
Thanks in advance.

divega:
@v-maudel I am not sure why the two spaces help. In markdown, two spaces at the end of a line is a way to introduce a new line. But I will merge this anyway because it isn't incorrect. Just to keep in mind next time a similar issue arises that maybe what you need is a new line.

v-maudel:
Thanks @divega , actually, the two spaces are for a breakline, we didn't insert a real breakline to not change the style of the document but we can suggest it for further fixes.

PR #1714, opened 2019-09-10
Remove redundant dir from Querying samples path

roji:
Am not sure if it's critical to add 301 redirections for sample file moves. If so, can anyone point me in the right direction for doing this?


bricelam:
Redirects are just for articles, not samples or assets

PR #1727, opened 2019-09-12
Update EF6 release notes

bricelam:


entity-framework/toc.md
```diff
@@ -140,7 +140,6 @@
 ## [Entity Framework 6](ef6/index.md)
 
 ### [What's New](ef6/what-is-new/index.md)
-#### [Roadmap](ef6/what-is-new/roadmap.md)
```
divega: Wondering if the roadmap page would be the right place to describe our policy for future releases, e.g. the fact that we are not investing in new features, people building new applications should consider EF Core, it is still supported, and we will consider PRs, etc. I think many customers will look for a page called the roadmap. 

Also, if the roadmap page is removed, we need to add a redirection to something anyway.

bricelam: I added a redirect to the whats-new page. There is already a blurb there about the focus being on EF Core

PR #1739, opened 2019-09-17
Reorganize table of contents

### LINKED ISSUE #1517: Reorganize ToC for new pattern with expandable nodes
Some users had difficulties navigating the documentation with topics that where expandable and at the same time had content. For that reason, changes have been made across all documentation so that any content has been moved as the first item under expandable topics. 

Because the change was automatic, the results are less than ideal. For example, the names of the topics is duplicated. In some cases, the content isn't very useful:

image

A few ideas to start a discussion:

- Should we come up with a new way to name the node that was created pointing to the contents of the root? 
- In many cases that page is currently almost empty. 
   - Should we try to fill it with introductory contents? 
   - Should we duplicate the child ToC items inside it? 
   - Remove it?

**Note: I am not even sure there is anything we can do about this other than removing the index pages that are currently empty in which we cannot add much value. It seems that the source code for our ToC wasn't changed. It is implemented as a change on how it is parsed.**

divega:
Fixes #1517, which is about ToC nodes that have child nodes and content being duplicated. 

Followed different approaches depending of the state of the content:

1. Made index.md the top child, usually with an "Overview" name in the ToC
2. Removed index pages that didn't add any value 
3. Merged a few pages to improve organization:
   - EF Core roadmap is now part of the What's new overview page
   - EF Core 3.0 index page with the EF Core 3.0 features page
   - Porting to EF Core validating requirements contents are now part of the main porting page
4. Added adequate redirects for any removed pages
5. Removed unnecessary breadcrumb TOC
6. Improved some titles
7. Fixed an incorrect redirect


divega:
@aspnet/efteam it would be great if someone can review. This includes several improvements I would like to have for launch day tomorrow.

entity-framework/core/what-is-new/ef-core-3.0/index.md
```diff
@@ -1,14 +1,199 @@
...
+
+[Owned]
+public class OrderDetails
+{
+    public int Id { get; set; }
+    public string ShippingAddress { get; set; }
+}
+```
+
+## EF 6.3 on .NET Core
```
smitpatel: Should this be here in what is new in EFCore 3.0 page?

divega: I think it complements the information about EF Core 3.0 features for the reader. We can add something to the paragraph to convey that, "yeah, we know this is not really an EF Core 3.0 feature"

divega: Made it smaller, with just a link to the EF 6.3 what's new page.

PR #1756, opened 2019-09-20
Refresh/reorganize samples

### LINKED ISSUE #1074: Consider changing samples to use .NET Core 3.0 by default
Changes like https://github.com/aspnet/EntityFramework.Docs/pull/1073 include a lot of churn on project files that would go away. 

Also it is just a really nicer default.

### LINKED ISSUE #1755: Update samples projects
* Some samples use old-style csproj :astonished:
* Update reference to Microsoft.Data.SqlClient
* Various other stuff

roji:
* Replace all old-style csproj with new SDK csproj, remove packages.config etc. (@AndriySvyryd looks like I did #1074 by mistake)
* Reference Microsoft.Data.SqlClient instead of System.Data.SqlClient
* Use new EF Core APIs (e.g. raw SQL)
* Change testing sample to use xunit instead of MSTest
* Rearrange directory structure in some cases

Closes #1755
Closes #1074


roji:
Ready for review, no warnings. Will handle the rest of logging separately (#1704)

samples/core/Miscellaneous/Logging/Logging/BloggingContextWithFiltering.cs
```diff
@@ -1,19 +1,28 @@
...
+                            && level == LogLevel.Information)
+                        .AddConsole();
+                });
+
+            // = new LoggerFactory(new[]
+            // {
+            //     new ConsoleLoggerProvider((category, level)
+            //         => category == DbLoggerCategory.Database.Command.Name
+            //            && level == LogLevel.Information, true)
+            // });
```
AndriySvyryd: Remove old code

samples/core/Miscellaneous/Testing/TestProject/TestProject.csproj
```diff
@@ -1,185 +1,21 @@
...
-    <ProjectTypeGuids>{3AC096D0-A1C2-E12C-1390-A8335801FDAB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
-    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">10.0</VisualStudioVersion>
-    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
-    <ReferencePath>$(ProgramFiles)\Common Files\microsoft shared\VSTT\$(VisualStudioVersion)\UITestExtensionPackages</ReferencePath>
-    <IsCodedUITest>False</IsCodedUITest>
-    <TestProjectType>UnitTest</TestProjectType>
-    <NuGetPackageImportStamp>
-    </NuGetPackageImportStamp>
-    <TargetFrameworkProfile />
+    <TargetFramework>netcoreapp3.0</TargetFramework>
```
AndriySvyryd: @bricelam Should we make the default TFM a variable as well?

bricelam: If it were up to me, I'd just keep it at the lowest compatible framework and only update it as needed.

divega: > @bricelam Should we make the default TFM a variable as well?

If we make the EF Core version a variable, I don't see why not. 

That said I think using variable here is a bit weird. On one hand I see that it makes the samples easier to update in the future, but:
- Each sample is no longer a self-contained simple app
- Code in the sample will often be EF Core version specific anyway

> If it were up to me, I'd just keep it at the lowest compatible framework and only update it as needed.

So in this case all samples should be EF Core 3.0, the minimum compatible will be netcoreapp3.0 (or netstandard2.1), correct?

divega:
> Assuming everything builds now :smile:

... and all the snippets still align!

samples/core/Miscellaneous/Testing/TestProject/TestProject.csproj
roji: I had the same hesitation when extracting nuget dependency versions to a solution Directory.Build.props - should each csproj stand on its own for ease-of-use etc. I'll revert that and have every csproj contain its own versions as before (3.0), it's not too much trouble to update.

roji:
> Assuming everything builds now smile

Yep it does :)

> ... and all the snippets still align!

Also quickly went through the modified conceptual doc pages (hopefully not too quickly) and everything seems OK. There are very few actual source changes in the samples.

samples/core/Miscellaneous/Testing/TestProject/TestProject.csproj
roji: Note: will update from 3.0.0-rc1 to 3.0.0 RTM after release (#1760)

Issue #1747, opened 2019-09-18
Unable to use DbFunction in 3.0


cwoolum:
I am trying to use `DatePart` in EfCore 3.0 but it seems to fail due to the `dbo` schema being prepended to the function name.

### Steps to reproduce

``` C#
public class AnalyticsDbContext : DbContext
{
    public AnalyticsDbContext(DbContextOptions options) : base(options)
    {
    }

    public DbSet<AggregatedMetric> AggregatedMetrics { get; set; }

    [DbFunction("DatePart", Schema = "")]
    public static int? DatePart(string datePartArg, DateTime? date) => throw new Exception();

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        base.OnModelCreating(modelBuilder);   
    }
}
```

I always get the error `{"Cannot find either column \"dbo\" or the user-defined function or aggregate \"dbo.DatePart\", or the name is ambiguous."}` even though my definition shows an empty schema.

I would expect that the `DbFunctionAttribute` would be respected and the schema would be blank. I halve also tried to use the fluent API to register with the same problem

```C#
modelBuilder
    .HasDbFunction(typeof(DateExtensions).GetMethod(nameof(DateExtensions.DatePart)))
    .HasName("DatePart")
    .HasSchema("");
```

### Further technical details

EF Core version: 3.0.0-rc1
Database provider: (e.g. Microsoft.EntityFrameworkCore.SqlServer)
Target framework: (e.g. .NET Core 3.0)
Operating system:
IDE: (e.g. Visual Studio 2019 16.3)

smitpatel:
@cwoolum - In EF Core 3.0, we removed assumption that empty schema means built-in function. So null or empty string as schema means it's default schema which would be dbo. in order to configure to use built in function it requires you to map custom translation like follows.
```C#
            modelBuilder
                .HasDbFunction(typeof(MyContext).GetMethod(nameof(MyContext.DatePart)))
                .HasTranslation(args => SqlFunctionExpression.Create("DatePart", args, typeof(int?), null));
```
Note that above would also fail because datepartarg will be printed as string which is not valid in SqlServer. The custom translation also need to take care of it but currently blocked by https://github.com/aspnet/EntityFrameworkCore/issues/17918

Overall for this scenario see https://github.com/aspnet/EntityFrameworkCore/issues/17268 for various ways we plan to improve this experience.

smitpatel:
[...]

PR #1759, opened 2019-09-20
Add breaking change for DbFunction.Schema empty string not a built-in…

smitpatel:
… function anymore

Part of #1747

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
+++ b/entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
@@ -64,6 +64,7 @@ Breaks from one 3.0 preview to another 3.0 preview aren't documented here.
 | [SQLitePCL.raw updated to version 2.0.0](#SQLitePCL) | Low      |
 | [NetTopologySuite updated to version 2.0.0](#NetTopologySuite) | Low      |
 | [Multiple ambiguous self-referencing relationships must be configured](#mersa) | Low      |
+| [DbFunction.Schema being null or empty string configures it to be in model's default schema](#udf-empty-string) | Low      |
 
 ### LINQ queries are no longer evaluated on the client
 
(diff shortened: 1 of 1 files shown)


entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -1711,3 +1712,38 @@ modelBuilder
...
+<a name="udf-empty-string"></a>
+### DbFunction.Schema being null or empty string configures it to be in model's default schema
+
+[Tracking Issue #12757](https://github.com/aspnet/EntityFrameworkCore/issues/12757)
+
+This change is introduced in EF Core 3.0-preview 7.
+
+**Old behavior**
+
+A DbFunction with value of schema being empty string configured the function to be in no schema and treated it as built-in function. For example following code will map `DatePart` CLR function to `DATEPART` built-in function on SqlServer.
```
AndriySvyryd:
```suggestion
A DbFunction configured with schema as an empty string was treated as built-in function without a schema. For example following code will map `DatePart` CLR function to `DATEPART` built-in function on SqlServer.
```

divega: A bit hard for me to read. My version can probably be improved further.

```suggestion
Configuring a `DbFunction` with a `Schema` equal to an empty string used to result in the function being mapped with no schema, which allowed SQL Server to treat it as a built-in function. For example the following code will map `DatePart` CLR function to `DATEPART` built-in function on SqlServer.
```

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -1711,3 +1712,38 @@ modelBuilder
...
+
+```C#
+[DbFunction("DATEPART", Schema = "")]
+public static int? DatePart(string datePartArg, DateTime? date) => throw new Exception();
+
+```
+
+**New behavior**
+
+All DbFunction mappings are considered to be mapped to user defined functions. Hence empty string value would put the function inside default schema for the model. It can be default schema configured explicitly via fluent API `modelBuilder.HasDefaultSchema()` or `dbo`.
```
AndriySvyryd:
```suggestion
All DbFunction mappings are considered to be mapped to user defined functions. Hence empty string value would put the function inside the default schema for the model. Which could be the schema configured explicitly via fluent API `modelBuilder.HasDefaultSchema()` or `dbo` otherwise.
```

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -1711,3 +1712,38 @@ modelBuilder
...
+
+**Why**
+
+Earlier schema being empty was a way to treat that function is built-in but that logic is only applicable for SqlServer where built-in function does not belong to any schema.
```
AndriySvyryd:
```suggestion
Previously schema being empty was a way to treat that function is built-in but that logic is only applicable for SqlServer where built-in functions do not belong to any schema.
```

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -1711,3 +1712,38 @@ modelBuilder
...
+
+**Mitigations**
+
+Configure DbFunction's translation manually to map it to built-in function.
```
AndriySvyryd:
```suggestion
Configure DbFunction's translation manually to map it to a built-in function.
```

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
divega:
```suggestion
The logic that allowed the empty schema to indicate a built-in function wasn't consistent or discoverable and was only applicable to SQL Server.
 ```

divega: I am kind of finding this argument a bit hard to buy myself. Is the new behavior more useful and applicable across databases. What happens in PostgreSQL now?

smitpatel: @roji to answer in detail.

It makes it more useful for other providers. PostgreSQL, even built-in functions have schema. And user defined functions can be used with schema if they are in default schema. Determination of built-in or not based on schema being empty string is arbitrary.

PR #1766, opened 2019-09-23
I thought Keyless Entity Types were added in 3.0??

SimonCropp:


smitpatel:
@SimonCropp - Thank you.

divega:
@smitpatel, @SimonCropp I am not sure this is better than what we had before. Query types existed since 2.1. Keyless entities is mostly a rename of the same concept to make the API easier to use and understand.

smitpatel:
We should rephrase the statement then. Earlier was also somewhat ambiguous.

divega:
Done.

PR #1779, opened 2019-09-26
Correct await foreach pattern

### LINKED ISSUE #1778: Asynchronous streams missing a namespace?
The documentation for the **async streams** with EF Core 3.0 shows a simple LINQ query to use await foreach. I tried that and it's not working as shown (at least a namespace is missing).

https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/what-is-new/ef-core-3.0/index.md#asynchronous-streams

The DbSet<T> implements IAsyncEnumerable, but the IQueryable extension methods don't.
Does a specific namespace need to be opened? I tried it with *Microsoft.EntityFrameworkCore*, but here other extension methods are not used. 
The correct namespace should be documented, or if there is a different way what needs to be done to make it work.
 
Thanks,
Christian

smitpatel:
Resolves #1778

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-3.0/index.md
+++ b/entity-framework/core/what-is-new/ef-core-3.0/index.md
@@ -77,7 +77,7 @@ var orders =
   where o.Status == OrderStatus.Pending
   select o;
 
-await foreach(var o in orders)
+await foreach(var o in orders.AsAsyncEnumerable())
 {
   Process(o);
 } 


entity-framework/core/what-is-new/ef-core-3.0/index.md
```diff
@@ -77,7 +77,7 @@ var orders =
   where o.Status == OrderStatus.Pending
   select o;
 
-await foreach(var o in orders)
+await foreach(var o in orders.AsAsyncEnumerable())
```
bricelam: Why isn't this pattern matched by the compiler?

smitpatel: Orders is of type `IQueryable<Order>` IQueryable implements IEnumerable (which compiler matches for foreach). IQueryable does not implement IAsyncEnumerable. :

bricelam: But there’s clearly an extension method that can handle it. The compiler uses extension methods in plenty of other places...

smitpatel: lol. File an issue with Roslyn?

divega: Not a crazy idea... Actually, it should match if there was a GetAsyncEnumerator<T\> method that matches!

divega: I tried this and I confirmed that it doesn't work :cry:. So I was about to file an issue for it, but i found that @jcouv already filed it at https://github.com/dotnet/roslyn/issues/32289.

I am going to add a comment, but feel free to upvote.

PR #1793, opened 2019-09-30
Correct errors from CATS report

smitpatel:
- Missing alerts

Also bunch of warnings from authoringpack in VS.

### DIFF
--- a/entity-framework/core/get-started/index.md
+++ b/entity-framework/core/get-started/index.md
@@ -18,11 +18,11 @@ You can follow the tutorial by using Visual Studio on Windows, or by using the .
 
 Install the following software:
 
-# [.NET Core CLI](#tab/netcore-cli)
+## [.NET Core CLI](#tab/netcore-cli)
 
 * [.NET Core 3.0 SDK](https://www.microsoft.com/net/download/core).
 
-# [Visual Studio](#tab/visual-studio)
+## [Visual Studio](#tab/visual-studio)
 
 * [Visual Studio 2019 version 16.3 or later](https://www.visualstudio.com/downloads/) with this  workload:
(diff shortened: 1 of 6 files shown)


entity-framework/core/get-started/index.md
```diff
@@ -18,11 +18,11 @@ You can follow the tutorial by using Visual Studio on Windows, or by using the .
 
 Install the following software:
 
-# [.NET Core CLI](#tab/netcore-cli)
+## [.NET Core CLI](#tab/netcore-cli)
```
roji: Wasn't aware, thanks!

entity-framework/core/modeling/relational/indexes.md
```diff
@@ -1,11 +1,11 @@
 ---
-title: Indexes - EF Core
+title: Indexes (Relational Database) - EF Core
 author: rowanmiller
 ms.date: 10/27/2016
 ms.assetid: 4581e7ba-5e7f-452c-9937-0aaf790ba10a
 uid: core/modeling/relational/indexes
 ---
-# Indexes
+# Indexes (Relational Database)
```
roji: Am going to unify the two index pages but OK for now.

entity-framework/core/get-started/index.md
smitpatel: This also shows warning. H2 with same title multiple times. But multiple H1 is more severe error. I will also test out not put it as heading at some point.

roji: The issue here is probably that the tab syntax is a docfx extension, this is just how it works... It's weird for multiple H2 with the same title to flag an issue even without tabs, maybe we can suppress certain issues.

PR #1798, opened 2019-10-01
Minor changes

jahbenjah:
1.  EF Core 3.0 is a .NET Standard 2.1 but no all version of EF Core.
2. in-memory and SQL Server providers are not included in current version of ASP.NET Core
3. Change version 2.2.2 to 3.0.0
4. Some paragraphs include information specific to  ASP.NET Core 2.x .


smitpatel:
I don't think there is value in adding explicit version everywhere. The page should be most relevant for latest version. And for previous version add a note section in the doc or simple line if small. If big enough then add section at the end of page with previous versions.

cc: @aspnet/efteam

jahbenjah:
As a user of EF Core,  I think than the documentation must be relevant for the specific version that I use not only the latest. Is possible to have a way to select doc's version just like ASP.NET Core documentation?
image

bricelam:
There is a lot of overhead with maintaining multiple versions of the docs. There are also only a few versions supported at one time. For example, versions 1.0, 1.1, and 2.0 are not supported and version 2.2 goes out of support in December. This only leaves versions 2.1 and 3.0 as really relevant.

We decided that it's more pragmatic to keep the docs focused on the latest version and call out version specific content in aside text.

ajcvickers:
@bricelam @smitpatel Was the decision here not to take this change as-is? Should we close it?

smitpatel:
We need to update the page after 0s and 1s PR but this PR does not address those and cannot be merged as is.

ajcvickers:
@jahbenjah Thanks for the proposal, but we're not going to merge this as it's taking things in a direction we don't want to go.

PR #1799, opened 2019-10-01
Update Query documentation

smitpatel:
Resolves#615
Resolves#1636
Resolves#1648
Resolves#1699
Resolves#1777


.openpublishing.redirection.json
```diff
@@ -214,6 +214,17 @@
             "source_path": "entity-framework/efcore-and-ef6/porting/ensure-requirements.md",
             "redirect_url": "/ef/efcore-and-ef6/porting/index",
             "redirect_document_id": false
+        },
+        {
+            "source_path": "entity-framework/core/querying/basic.md",
+            "redirect_url": "/ef/core/querying/index",
+            "redirect_document_id": false
+        }
+        ,
```
AndriySvyryd: Extra newline

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
+- LINQ operator being used has no server side translation.
+
+In such cases, user can explicitly opt into client evaluation by calling methods like `AsEnumerable` (or `AsAsyncEnumerable` for async).
+
+[!code-csharp[Main](../../../samples/core/Querying/ClientEval/Sample.cs#ExplicitClientEval)]
+
+## Potential memory leak in client evaluation
+
+Since query translation and compilation is expensive, by default, EF Core caches compiled query plan. With client evaluation allowed in the top projection, the cached delegate can use client code. In most cases, EF Core generates parameters for client parts of the tree so that their values can be replaced when reusing the query plan. But with certain constants in expression tree, it cannot be converted to parameters. If the client code contains such constants then those objects cannot be garbage collected. If such object contains DbContext or other services inside it then it could cause memory usage of the app to grow over time. In order to avoid potential memory leak, EF Core throws exception whenever it encounter constants of type which cannot be mapped in current database providers. Common causes and their solutions are as follows:
```
AndriySvyryd:
```suggestion
Since query translation and compilation is expensive, by default, EF Core caches the compiled query plan. With client evaluation allowed in the top projection, the cached delegate can use client code. In most cases, EF Core generates parameters for client parts of the tree so that their values can be replaced when reusing the query plan. But certain constants in expression tree cannot be converted to parameters. If the client code contains such constants then those objects cannot be garbage collected as they would still be referenced from the cached query plan. If such object contains DbContext or other services inside it then it could cause memory usage of the app to grow over time. In order to avoid potential memory leak, EF Core throws an exception whenever it encounters constants of type which cannot be mapped using the current database provider. Common causes and their solutions are as follows:
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,120 @@
+---
+title: Complex Query Operators - EF Core
+author: smitpatel
+ms.date: 09/30/2019
+ms.assetid: 2e187a2a-4072-4198-9040-aaad68e424fd
+uid: core/querying/complex-query-operators
+---
+# Complex Query Operators
+
+Language Integrated Query (LINQ) contains many complex operators which combines multiple data sources or does complex processing. Not all LINQ operators have suitable translation in server side. Sometimes, query written in one form can be translated but in different form doing different computation cannot be translated. This page describes some of the complex operators and their supported variations. In future we may recognized more pattern and add their translation. Important point to remember is database also varies in what they support so a particular query which is translated in SqlServer may not work for Sqlite database.
```
AndriySvyryd:
```suggestion
Language Integrated Query (LINQ) contains many complex operators which combines multiple data sources or does complex processing. Not all LINQ operators have suitable translations on the server side. Sometimes, query written in one form can be translated but in different form it can't be even if the result is same. This page describes some of the complex operators and their supported variations. In future we may recognize more patterns and add their translations. Important point to remember is that database capabilities also vary, so a particular query which is translated on SqlServer may not work for a Sqlite database.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,120 @@
...
+
+```SQL
+SELECT [p].[PersonId], [p].[Name], [p].[PhotoId], [p0].[PersonPhotoId], [p0].[Caption], [p0].[Photo]
+FROM [PersonPhoto] AS [p0]
+INNER JOIN [Person] AS [p] ON [p0].[PersonPhotoId] = [p].[PhotoId]
+```
+
+## GroupJoin
+
+LINQ GroupJoin operator allows you to connect two data sources in same way as Join but the different here is, it creates a group of values from inner source which matches key with element from outer source. Performing query like following generates result of `Blog` & `IEnumerable<Post>`. Since databases (especially relational databases) do not support representing a collection of client side objects, GroupJoin cannot be translated to server in many cases. It requires to get all the data from server to perform GroupJoin without special selector. But if the selector is limiting data being selected then fetching all data from server may cause performance issues. Hence EF Core does not translate GroupJoin.
```
AndriySvyryd:
```suggestion
LINQ GroupJoin operator allows you to connect two data sources in same way as Join but the difference here is that it creates a group of values from inner source which matches the key with an element from the outer source. Performing a query like the following generates result of `Blog` & `IEnumerable<Post>`. Since databases (especially relational databases) do not support representing a collection of client side objects, GroupJoin cannot be translated to server in many cases. It requires to get all the data from server to perform GroupJoin without a special selector. But if the selector is limiting data being selected then fetching all data from server may cause performance issues. Hence EF Core does not translate GroupJoin.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,120 @@
...
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#GroupJoin)]
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#GroupJoinComposed)]
+
+## SelectMany
+
+LINQ SelectMany operator allows you to enumerate over collection selector for each element in outer source and generate tuples of value from each data source. In a way it is a join but without any condition so every record is connected with other. Depending on how collection selector is related to outer data source, SelectMany can translate to various different queries on server side.
```
AndriySvyryd:
```suggestion
LINQ SelectMany operator allows you to enumerate over collection selector for each element in outer source and generate tuples of value from each data source. In a way it is a join but without any condition so every record is connected with another. Depending on how collection selector is related to outer data source, SelectMany can translate to various different queries on server side.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Collection selector does not reference outer
+
+When collection selector is not referencing anything from outer source, it is cartesian product of both data sources. In translates to `CROSS JOIN` in relational databases.
```
AndriySvyryd:
```suggestion
When collection selector is not referencing anything from the outer source, it is a cartesian product of both data sources. In translates to `CROSS JOIN` in relational databases.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,120 @@
...
+
+```SQL
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], [p].[PostId], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Rating], [p].[Title]
+FROM [Blogs] AS [b]
+CROSS JOIN [Posts] AS [p]
+```
+
+### Collection selector references outer in predicate
+
+When collection selector has a where predicate which is using element from outer source, normally this arises when using collection navigation on outer source as collection selector, then such where predicate can be used as join condition and it can translate to join on databases. If the collection is empty for outer element, then it would not generate any result for that outer element. But if `DefaultIfEmpty` is applied on collection selector then outer element will be connected with default value of inner element. Due to this distinction, this kind of queries translate to `INNER JOIN` in the absence of `DefaultIfEmpty` and `LEFT JOIN` when `DefaultIfEmpty` is applied.
```
AndriySvyryd:
```suggestion
When collection selector has a where predicate which is using an element from the outer source (normally this arises when using collection navigation on outer source as collection selector) then such where predicate can be used as the join condition and it can be translated to a join on the server side. If the collection is empty for the outer element, then it would not generate any result for that outer element. But if `DefaultIfEmpty` is applied on collection selector then outer element will be connected with the default value of inner element. Due to this distinction, this kind of queries translate to `INNER JOIN` in the absence of `DefaultIfEmpty` and `LEFT JOIN` when `DefaultIfEmpty` is applied.
```

entity-framework/core/querying/complex-query-operators.md
bricelam: Casing should be *SQLite* unless you're referring to our packages or APIs.

entity-framework/core/querying/client-eval.md
bricelam: I wonder if we should avoid the word *constant* here.

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,120 @@
...
+
+> [!TIP]
+> You can view this article's [sample](https://github.com/aspnet/EntityFramework.Docs/tree/master/samples/core/Querying) on GitHub.
+
+## Join
+
+LINQ Join operator allows you to connect two data sources based on the key selector on each source and generates a tuple of values when key matches. This naturally translates to `INNER JOIN` on relational databases. While LINQ Join has outer & inner key selectors, databases take a condition as join condition and EF Core will generate join condition by comparing outer key selector to inner key selector through equality operation. Further, if key selectors are anonymous types, EF Core will generate join condition to perform equality component wise.
```
bricelam: Would *left* and *right* be clearer than *outer* and *inner*?

bricelam: *source* might also be confusing. Maybe *set* is more idiomatic for EF users

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,120 @@
...
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#SelectManyConvertedToCrossJoin)]
+
+```SQL
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], [p].[PostId], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Rating], [p].[Title]
+FROM [Blogs] AS [b]
+CROSS JOIN [Posts] AS [p]
+```
+
+### Collection selector references outer in predicate
```
bricelam: Another word that might be too technical: *predicate*

entity-framework/core/querying/how-query-works.md
```diff
@@ -3,7 +3,7 @@ title: How Queries Work - EF Core
 author: rowanmiller
 ms.date: 09/26/2018
 ms.assetid: de2e34cd-659b-4cab-b5ed-7a979c6bf120
-uid: core/querying/overview
+uid: core/querying/how-query-works
```
bricelam: I don't think you can change these. Was the old one redirected?

entity-framework/core/querying/client-eval.md
smitpatel: What should we call it?

entity-framework/core/querying/complex-query-operators.md
smitpatel: where? filter?

entity-framework/core/querying/how-query-works.md
smitpatel: yes. I added redirects.

entity-framework/core/querying/complex-query-operators.md
bricelam: where clause?

entity-framework/core/querying/async.md
```diff
@@ -1,14 +1,14 @@
 ---
 title: Asynchronous Queries - EF Core
-author: rowanmiller
-ms.date: 01/24/2017
+author: smitpatel
+ms.date: 10/01/2019
```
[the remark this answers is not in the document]
smitpatel: Removed

entity-framework/core/querying/client-eval.md
smitpatel: ideas?

entity-framework/core/querying/complex-query-operators.md
smitpatel: Open to ideas. APIs call them outer & inner source.

entity-framework/core/querying/client-eval.md
bricelam: Not sure, but I think to most users would think of the `const` keyword.
*references to fields and properties* gets a little closer, but probably isn't right

smitpatel: reference to "this" is the only one which causes issue.

smitpatel:
Updated.

entity-framework/core/querying/async.md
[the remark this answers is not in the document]
smitpatel: [Warning] Missing attribute: ms.date. A freshness date is required for this content set. Add a date in format MM/DD/YYYY.

need to add it. :X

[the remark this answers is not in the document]
AndriySvyryd: Also I remember us discussing that it should represent the date when there was a content change, not just formatting or spelling

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+```SQL
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], [p].[PostId], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Rating], [p].[Title]
+FROM [Blogs] AS [b]
+CROSS JOIN [Posts] AS [p]
+```
+
+### Collection selector references outer in where clause
+
+When collection selector has a where predicate, which is referencing the outer element, then EF Core translates it to join on database and uses the predicate as join condition. Normally it arises when using collection navigation on outer source as collection selector. If the collection is empty for outer element, then it wouldn't generate any result for that outer element. But if `DefaultIfEmpty` is applied on collection selector then outer element will be connected with default value of inner element. Because of this distinction, this kind of queries translates to `INNER JOIN` in the absence of `DefaultIfEmpty` and `LEFT JOIN` when `DefaultIfEmpty` is applied.
```
roji:
```suggestion
When the collection selector has a Where predicate which references the outer element, EF Core translates it to a database join and uses the predicate as the join condition. Normally this occurs when using a collection navigation on an outer source as the collection selector. If the collection is empty for the outer element, then no results would be generated for that outer element. But if `DefaultIfEmpty` is applied on the collection selector, the outer element will be connected with the default value of the inner element type. Because of this distinction, this kind of queries translates to `INNER JOIN` in the absence of `DefaultIfEmpty` and `LEFT JOIN` when `DefaultIfEmpty` is applied.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
+---
+title: Complex Query Operators - EF Core
+author: smitpatel
+ms.date: 10/03/2019
+ms.assetid: 2e187a2a-4072-4198-9040-aaad68e424fd
+uid: core/querying/complex-query-operators
+---
+# Complex Query Operators
```
roji: We probably need a note that this page applies only to relational database.

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+INNER JOIN [Posts] AS [p] ON [b].[BlogId] = [p].[BlogId]
+
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], [p].[PostId], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Rating], [p].[Title]
+FROM [Blogs] AS [b]
+LEFT JOIN [Posts] AS [p] ON [b].[BlogId] = [p].[BlogId]
+```
+
+### Collection selector references outer in non-predicate
+
+When collection selector has reference to outer, which isn't in predicate (as the above case), it can't be converted to join. That's why we need to evaluate collection selector for each outer element. It translates to `APPLY` operation in relational databases. If the collection is empty for outer element, then it wouldn't generate any result for that outer element. But if `DefaultIfEmpty` is applied on collection selector then outer element will be connected with default value of inner element. Because of this distinction, this kind of queries translates to `CROSS APPLY` in the absence of `DefaultIfEmpty` and `OUTER APPLY` when `DefaultIfEmpty` is applied. Sqlite doesn't support `APPLY` operator.
```
roji: This is quite SQL Server-specific.

```suggestion
When the collection selector references the outer element, but from a non-predicate operator (as was the  case above), the query can't be translated to a regular database join, since we need to evaluate collection selector for each outer element. However, the query does translate to an `APPLY` or a lateral join operation in relational databases.

As before, if the collection is empty for the outer element, then it wouldn't generate any result for that outer element. But if `DefaultIfEmpty` is applied on collection selector then outer element will be connected with the default value of inner element type. Because of this distinction, on SQL Server this kind of query translates to `CROSS APPLY` in the absence of `DefaultIfEmpty` and `OUTER APPLY` when `DefaultIfEmpty` is applied. 

Sqlite doesn't support `APPLY` operator, and therefore this kind of query cannot be translated when targeting it.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+CROSS APPLY [Posts] AS [p]
+
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], ([b].[Url] + N'=>') + [p].[Title] AS [p]
+FROM [Blogs] AS [b]
+OUTER APPLY [Posts] AS [p]
+```
+
+## GroupBy
+
+LINQ GroupBy operator creates result of type `IGrouping<TKey, TElement>` where `TKey` & `TElement` could be arbitrary types. Further, `IGrouping` implements `IEnumerable<TElement>`, which means you can compose over using any LINQ operator after grouping. Since no database structure can represent IGrouping, Group By operator has no translation in most cases. When an aggregate operator is applied to each group to which gives back a scalar, it can be translated to SQL `GROUP BY` in relational databases. SQL `GROUP BY` is restrictive too. It requires you to group by only scalar values  projection can contain only grouping key columns or aggregate applied over any column. EF Core identifies this subset and translates it to server. Following example shows a query, which has aggregate operator applied to it and generated SQL.
```
roji:
```suggestion
The LINQ GroupBy operator returns results of type `IGrouping<TKey, TElement>`, where `TKey` and `TElement` could be arbitrary types. Further, `IGrouping` implements `IEnumerable<TElement>`, which means it can be composed over using any LINQ operator after grouping. Since no database structure can represent `IGrouping`, the GroupBy operator cannot be translated in most cases. As an exception, when an aggregate operator is applied to each group (e.g. MAX, COUNT), the query can be translated to  SQL `GROUP BY`. 

Note that SQL `GROUP BY` is restrictive too, requiring you to group only by scalar values; it also requires the projection to only contain grouping key columns or applied aggregate operators. EF Core identifies this subset and translates it to server, as in the following example:
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#GroupBy)]
+
+```SQL
+SELECT [p].[AuthorId] AS [Key], COUNT(*) AS [Count]
+FROM [Posts] AS [p]
+GROUP BY [p].[AuthorId]
+```
+
+EF Core also translates when aggregate operator appears in Where filter by using `HAVING` clause in SQL and aggregate operator appears in Order By (or other ordering operators). The query before applying Group By operator can be any complex query using various different operators as long as it can be translated to server. Further, once you apply aggregate operator on grouping query to remove groupings from resulting source, you can compose it like any other query.
```
roji:
```suggestion
EF Core can also translate queries where an aggregate operator appears in a Where filter (by generating an SQL `HAVING` clause), or in an OrderBy clause (or other ordering operators). The query part before the GroupBy operator can be any complex query as long as it can be translated to server. Further, once you apply aggregate operator on a grouping query as shown above, you can compose it like any other query.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+The aggregate operators EF Core supports are following
+
+- Average
+- Count
+- LongCount
+- Max
+- Min
+- Sum
+
+## Left Join
```
roji: This repeats the above GroupJoin section no?

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
-    }
+[!code-csharp[Main](../../../samples/core/Querying/ClientEval/Sample.cs#ClientProjection)]
 
-    return url;
-}
-```
+[!code-csharp[Main](../../../samples/core/Querying/ClientEval/Sample.cs#ClientMethod)]
 
-## Client evaluation performance issues
+## Unsupported client evaluation
```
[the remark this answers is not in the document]
smitpatel: I don't agree with this reasoning. A client code worked in projection failed in where. Users would have question why that happens.

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
-    .Where(blog => StandardizeUrl(blog.Url).Contains("dotnet"))
-    .ToList();
-```
+[!code-csharp[Main](../../../samples/core/Querying/ClientEval/Sample.cs#ClientWhere)]
+
+## Explicit client evaluation
+
+User may need to force into client evaluation explicitly in certain cases like following
+
+- Data is small so that doing evaluation on client doesn't cause huge performance penalty.
```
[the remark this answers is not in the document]
smitpatel: You can disagree with it but our users don't. Our users are indeed using `AsEnumerable` to cause client eval and it is part of our exception message too.

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
+- LINQ operator being used has no server-side translation.
+
+In such cases, user can explicitly opt into client evaluation by calling methods like `AsEnumerable` (or `AsAsyncEnumerable` for async).
+
+[!code-csharp[Main](../../../samples/core/Querying/ClientEval/Sample.cs#ExplicitClientEval)]
+
+## Potential memory leak in client evaluation
+
+Since query translation and compilation are expensive, EF Core caches the compiled query plan. The cached delegate may use client code while doing client evaluation of top-level projection. EF Core generates parameters for client parts of the tree, to reuse the query plan by replacing parameter values. But certain constants in expression tree can be converted to parameters. If the cached delegate contains such constants, then those objects can't be garbage collected as they're still being referenced. If such object contains DbContext or other services inside it, then it could cause memory usage of the app to grow over time. It's generally sign of memory leak. EF Core throws exception whenever it comes across constants of type that can't be mapped using current database providers. Common causes and their solutions are as follows:
```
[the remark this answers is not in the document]
smitpatel: I don't think users are ever going to understand it anyway. :trollface: 

Important point is 3 solutions which describes what to do when we throw exception.

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#Join)]
+
+```SQL
+SELECT [p].[PersonId], [p].[Name], [p].[PhotoId], [p0].[PersonPhotoId], [p0].[Caption], [p0].[Photo]
+FROM [PersonPhoto] AS [p0]
+INNER JOIN [Person] AS [p] ON [p0].[PersonPhotoId] = [p].[PhotoId]
+```
+
+## GroupJoin
```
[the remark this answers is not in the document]
smitpatel: No variant of GroupJoin is translatable. Period.
The 2 queries are to demonstrate different amount of data needed to be fetch from server.
There is very little confusion about distinction between join/groupjoin syntax in customers. Most customers so far have known difference. The other query is not `from.... where...` it is actual group join and then composition afterwards. So I don't see any clarification required.

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+## SelectMany
+
+LINQ SelectMany operator allows you to enumerate over collection selector for each outer element and generate tuples of value from each data source. In a way, it's a join but without any condition so every outer element is connected with element from collection source. Depending on how collection selector is related to outer data source, SelectMany can translate to various different queries on server side.
+
+### Collection selector doesn't reference outer
+
+When collection selector isn't referencing anything from the outer source, the result is a cartesian product of both data sources. It translates to `CROSS JOIN` in relational databases.
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#SelectManyConvertedToCrossJoin)]
```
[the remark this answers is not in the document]
smitpatel: Decided not to do. LINQ has both syntax for various thing and LINQ docs can describe that. Our docs would be really long with very low content relating to us.

entity-framework/core/querying/complex-query-operators.md
smitpatel: That is incorrect.

entity-framework/core/querying/complex-query-operators.md
smitpatel: Lateral join and apply operations are different. It is fine the way it is. It is Oracle-specific.

entity-framework/core/querying/complex-query-operators.md
smitpatel: Nope.

entity-framework/core/querying/async.md
```diff
@@ -1,21 +1,21 @@
...
+author: smitpatel
+ms.date: 10/03/2019
 ms.assetid: b6429b14-cba0-4af4-878f-b829777c89cb
 uid: core/querying/async
 ---
 
 # Asynchronous Queries
 
-Asynchronous queries avoid blocking a thread while the query is executed in the database. This can be useful to avoid freezing the UI of a thick-client application. Asynchronous operations can also increase throughput in a web application, where the thread can be freed up to service other requests while the database operation completes. For more information, see [Asynchronous Programming in C#](https://docs.microsoft.com/dotnet/csharp/async).
+Asynchronous queries avoid blocking a thread while the query is executed in the database. It avoids freezing the UI of a thick-client application. Async operations can also increase throughput in a web application, where async query would free up the thread to service other requests while the database operation completes. For more information, see [Asynchronous Programming in C#](/dotnet/csharp/async).
```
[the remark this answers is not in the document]
smitpatel: Acrolinx score went down by 2.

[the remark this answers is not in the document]
smitpatel: I changed it something more simplistic and got perfect score of 100.

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
+ms.date: 10/03/2019
 ms.assetid: 8b6697cc-7067-4dc2-8007-85d80503d123
 uid: core/querying/client-eval
 ---
 # Client vs. Server Evaluation
 
-Entity Framework Core supports parts of the query being evaluated on the client and parts of it being pushed to the database. It is up to the database provider to determine which parts of the query will be evaluated in the database.
+To evaluate query on server as much as possible, Entity Framework Core converts parts of the query to parameters, which it can evaluate on client side. Rest of the query with generated parameters is given to database providers to determine what should be equivalent database query to do evaluation on the server. EF Core supports partial client evaluation in top-level projection (essentially, the last call to `Select()`). If parts of query in top-level projection can't be translated to server, EF Core will fetch required data from server and evaluate it on client. If EF Core detects expression, which can't be translated to server in any other place, then it throws runtime exception. See [how query works](xref:core/querying/how-query-works) to understand how EF Core determines what can't be translated to server.
```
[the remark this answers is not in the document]
smitpatel: I don't see much value in specifying performance. This is doc about EF behavior. We can mention but not describe.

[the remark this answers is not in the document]
smitpatel: Also there are queries where client eval would be more performant. :trollface:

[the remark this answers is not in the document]
smitpatel: Change it little bit to simplify.

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
+
+- **Using instance method**: When using instance methods in client projection, the expression tree contains constant of the instance. If your method doesn't use any data from instance, consider making the method static. If you need instance data in the method body, then pass specific data as an argument to the method.
+- **Passing constant argument to method**: This case arises generally by using `this` in argument to client method. Consider splitting argument in to multiple scalar arguments, which can be mapped by database provider.
+- **Other constants**: If constant is seen in any other case, then you can evaluate if constant is needed in processing. If it's necessary to have the constant or if you can't use solution to above cases, then create a local variable to store the value and use local variable in the query. EF Core will convert it to parameter.
 
-## Client evaluation logging
+## Previous versions
 
-By default, EF Core will log a warning when client evaluation is performed. See [Logging](../miscellaneous/logging.md) for more information on viewing logging output. 
+The following section applies to EF Core versions before 3.0.
```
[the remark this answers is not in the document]
smitpatel: We need to determine the formatting on how to describe behavior in previous versions. I am keeping this as is for now.

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
 
-## Optional behavior: throw an exception for client evaluation
+EF Core supported client evaluation in any part query that's why queries similar to one posted in [Unsupported client evaluation](#unsupported-client-evaluation) section worked correctly. But it caused performance issue without users noticing. To notify users that query was being evaluated on client, EF Core logged client evaluation warning. For more information on viewing logging output, see [Logging](xref:core/miscellaneous/logging).
 
-You can change the behavior when client evaluation occurs to either throw or do nothing. This is done when setting up the options for your context - typically in `DbContext.OnConfiguring`, or in `Startup.cs` if you are using ASP.NET Core.
+Optionally EF Core allowed to change default behavior to throw exception or do nothing when doing client evaluation except for projection. The exception throwing would make it similar to the behavior in 3.0. To change the behavior, you need to configure warnings while setting up the options for your context - typically in `DbContext.OnConfiguring`, or in `Startup.cs` if you're using ASP.NET Core.
```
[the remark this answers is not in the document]
smitpatel: There is also ignoring warnings rather than logging which above paragraph does not cover. ConfigureWarnings has Throw/Log/Ignore options.

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
-```
+> [!TIP]
+> You can view this article's [sample](https://github.com/aspnet/EntityFramework.Docs/tree/master/samples/core/Querying) on GitHub.
 
-<!-- [!code-csharp[Main](samples/core/Querying/ClientEval/Sample.cs)] -->
-``` csharp
-public static string StandardizeUrl(string url)
-{
-    url = url.ToLower();
+In the following example, a helper method is used to standardize URLs for blogs, which are returned from a SQL Server database. Since the SQL Server provider has no insight into how this method is implemented, it isn't possible to translate it into SQL. All other aspects of the query are evaluated in the database, but passing the returned `URL` through this method is done on the client.
```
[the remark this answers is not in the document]
smitpatel: Title makes it clear that it is in top-level projection.

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
+---
+title: Complex Query Operators - EF Core
+author: smitpatel
+ms.date: 10/03/2019
+ms.assetid: 2e187a2a-4072-4198-9040-aaad68e424fd
+uid: core/querying/complex-query-operators
+---
+# Complex Query Operators
+
+Language Integrated Query (LINQ) contains many complex operators, which combine multiple data sources or does complex processing. Not all LINQ operators have suitable translation in the server side. Sometimes, query in one form translates to server but if written in different form doesn't translate even if the result is same. This page describes some of the complex operators and their supported variations. In future, we may recognize more pattern and add their translations. Important point to remember is that database also varies in what they support. A particular query, which is translated in SqlServer, mayn't work for SQLite database.
```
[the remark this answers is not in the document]
smitpatel: It changes meaning of what we want to say.

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+> [!TIP]
+> You can view this article's [sample](https://github.com/aspnet/EntityFramework.Docs/tree/master/samples/core/Querying) on GitHub.
+
+## Join
+
+LINQ Join operator allows you to connect two data sources based on the key selector on each source and generates a tuple of values when key matches. It naturally translates to `INNER JOIN` on relational databases. While LINQ Join has outer & inner key selectors, databases take a condition as join condition. So EF Core will generate join condition by comparing outer key selector to inner key selector through equality operation. Further, if key selectors are anonymous types, EF Core will generate join condition to do equality component wise.
```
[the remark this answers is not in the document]
smitpatel: Not `The` before LINQ.

entity-framework/core/querying/complex-query-operators.md
roji: They're not really that different. PostgreSQL implements the APPLY behavior with a LATERAL JOIN, there's just an additional ON TRUE there.

entity-framework/core/querying/client-eval.md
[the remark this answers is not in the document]
roji: > I don't see much value in specifying performance

??

Performance is one of the two (extremely) major reasons we don't have client evaluation anymore... It seems critical for the understanding of what client vs. server evaluation is?

entity-framework/core/querying/async.md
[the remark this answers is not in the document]
roji: @smitpatel I don't know what Acrolinx is or how it scores, but "where async query" is missing an indefinite article, and using "avoid" twice sentence after sentence is problematic stylistically.

entity-framework/core/querying/client-eval.md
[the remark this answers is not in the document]
roji: This needs simple code samples instead of a wall of text.

entity-framework/core/querying/complex-query-operators.md
smitpatel: Acrolinx disallows uses of this.

roji:
@smitpatel FYI am going to stop reviewing the responses to my PR, as 95% seem to be simple rejections with no explanation.

entity-framework/core/querying/complex-query-operators.md
roji: Care to elaborate?

entity-framework/core/querying/complex-query-operators.md
roji: That is not a valid response to a review comment. This fragment is clearly missing articles.

entity-framework/core/querying/complex-query-operators.md
smitpatel: Again Lateral joins are apply operations are different. Postgre fits apply operation on lateral join with additional condition. I have added statement that query may not translate always from EF docs perspective, we generate Apply operation and that is accurate doc content.

entity-framework/core/querying/complex-query-operators.md
smitpatel: Acrolinx doesn't like long statements.

entity-framework/core/querying/complex-query-operators.md
smitpatel: Avoid usage of can when we do it.

entity-framework/core/querying/client-eval.md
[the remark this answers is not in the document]
roji: I'm talking about your logic. The fact that something doesn't cause a huge perf penalty is not, in itself, a justification to use it. You use it because you have no choice or because there's some other benefit.

entity-framework/core/querying/complex-query-operators.md
smitpatel: Acrolinx also dislikes parenthesis.

entity-framework/core/querying/complex-query-operators.md
smitpatel: Acrolinx believe that Note that does not add any value and unncessary.

smitpatel:
@roji - I incorporated the suggestions just not as is because acrolinx complained about it. I have taken ideas from the suggestion to update the text to simplify it. The outdated items above and if you read the docs, I have updated the content. By no means I feel obligated to take all the suggestions as is.

cc: @ajcvickers @bricelam

entity-framework/core/querying/complex-query-operators.md
roji: The fact is that what this section describes (Collection selector references outer in non-predicate) is translated in PostgreSQL via lateral join, which is part of the SQL standard; PostgreSQL users reading this will have no idea about APPLY. All I'm asking is to mention lateral join alongside APPLY so that they understand, because that's what is used for translation.

entity-framework/core/querying/async.md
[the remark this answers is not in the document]
smitpatel: This is outdated diff.

entity-framework/core/querying/client-eval.md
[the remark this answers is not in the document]
smitpatel: I don't see value in specifying this
> This is important for performance: the server performs the work, possibly using indexes for efficiency, and only sends the results back to the client, minimizing network bandwidth needs.

Anyone using databases understands that. The intent of this page is not to empasize that in my mind.

entity-framework/core/querying/complex-query-operators.md
smitpatel: That is misleading. Relational generates Apply operation. EF Core relational layer does not generate lateral joins. Writing lateral join easily confuses user asking, oh why did not translate this query via lateral join. Sorry but we don't support lateral join.

roji: I think I'll let this PR go now.

entity-framework/core/querying/client-eval.md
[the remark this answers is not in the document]
smitpatel: Again that is not true. It is simple cost vs benefit analysis. If I have table of 40 rows and a client filter function, I can also convert the client function to a database function and use UDF to map it. I can get best performance at the expense of spending ours to generate UDF when I can just do client eval because it is just cheaper and does not cause huge penalty.

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+```SQL
+SELECT [p].[PersonId], [p].[Name], [p].[PhotoId], [p0].[PersonPhotoId], [p0].[Caption], [p0].[Photo]
+FROM [PersonPhoto] AS [p0]
+INNER JOIN [Person] AS [p] ON [p0].[PersonPhotoId] = [p].[PhotoId]
+```
+
+## GroupJoin
+
+LINQ GroupJoin operator allows you to connect two data sources in similar to Join but it creates a group of inner values for matching outer elements. Executing query like following generates result of `Blog` & `IEnumerable<Post>`. Since databases (especially relational databases) don't support representing a collection of client-side objects, GroupJoin can't be translated to server in many cases. It requires to get all the data from server to do GroupJoin without special selector (first query below). But if the selector is limiting data being selected then fetching all data from server may cause performance issues (second query below). That's why EF Core doesn't translate GroupJoin.
```
bricelam:
```suggestion
LINQ's GroupJoin operator allows you to connect two data sources similar to Join, but it creates a group of inner values for matching outer elements. Executing a query like the following generates a result of `Blog` & `IEnumerable<Post>`. Since databases--especially relational databases--don't have a way to represent collections of client-side objects, GroupJoin can't be translated to the server in many cases. It requires you to get all of the data from the server to do GroupJoin without a special selector (first query below). But if the selector is limiting data being selected then fetching all of the data from the server may cause performance issues (second query below). That's why EF Core doesn't translate GroupJoin.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#GroupJoin)]
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#GroupJoinComposed)]
+
+## SelectMany
+
+LINQ SelectMany operator allows you to enumerate over a collection selector for each outer element and generate tuples of value from each data source. In a way, it's a join but without any condition so every outer element is connected with element from collection source. Depending on how the collection selector is related to the outer data source, SelectMany can translate to various different queries on the server side.
```
bricelam:
```suggestion
LINQ's SelectMany operator allows you to enumerate over a collection selector for each outer element and generate tuples of values from each data source. In a way, it's a join but without any condition so every outer element is connected with an element from the collection source. Depending on how the collection selector is related to the outer data source, SelectMany can translate into various different queries on the server side.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+```SQL
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], [p].[PostId], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Rating], [p].[Title]
+FROM [Blogs] AS [b]
+CROSS JOIN [Posts] AS [p]
+```
+
+### Collection selector references outer in where clause
+
+When the collection selector has a where clause, which references the outer element, then EF Core translates it to a database join and uses the predicate as the join condition. Normally it arises when using collection navigation on the outer element as the collection selector. If the collection is empty for outer element, then no results would be generated for that outer element. But if `DefaultIfEmpty` is applied on the collection selector then the outer element will be connected with default value of the inner element. Because of this distinction, this kind of queries translates to `INNER JOIN` in the absence of `DefaultIfEmpty` and `LEFT JOIN` when `DefaultIfEmpty` is applied.
```
bricelam:
```suggestion
When the collection selector has a where clause, which references the outer element, then EF Core translates it to a database join and uses the predicate as the join condition. Normally this arises when using collection navigation on the outer element as the collection selector. If the collection is empty for an outer element, then no results would be generated for that outer element. But if `DefaultIfEmpty` is applied on the collection selector then the outer element will be connected with a default value of the inner element. Because of this distinction, this kind of queries translates to `INNER JOIN` in the absence of `DefaultIfEmpty` and `LEFT JOIN` when `DefaultIfEmpty` is applied.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#SelectManyConvertedToCrossJoin)]
+
+```SQL
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], [p].[PostId], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Rating], [p].[Title]
+FROM [Blogs] AS [b]
+CROSS JOIN [Posts] AS [p]
+```
+
+### Collection selector references outer in where clause
```
bricelam:
```suggestion
### Collection selector references outer in a where clause
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], [p].[PostId], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Rating], [p].[Title]
+FROM [Blogs] AS [b]
+INNER JOIN [Posts] AS [p] ON [b].[BlogId] = [p].[BlogId]
+
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], [p].[PostId], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Rating], [p].[Title]
+FROM [Blogs] AS [b]
+LEFT JOIN [Posts] AS [p] ON [b].[BlogId] = [p].[BlogId]
+```
+
+### Collection selector references outer in non-where case
```
bricelam:
```suggestion
### Collection selector references outer in a non-where case
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+When the collection selector has reference to the outer element, which isn't in where clause as the case above, it doesn't translate to database join. That's why we need to evaluate the collection selector for each outer element. It translates to `APPLY` operation in many relational databases. If the collection is empty for outer element, then no results would be generated for that outer element. But if `DefaultIfEmpty` is applied on the collection selector then the outer element will be connected with default value of the inner element. Because of this distinction, this kind of queries translates to `CROSS APPLY` in the absence of `DefaultIfEmpty` and `OUTER APPLY` when `DefaultIfEmpty` is applied. Certain databases like SQLite don't support `APPLY` operator so this kind of query may not be translated.
```
bricelam:
```suggestion
When the collection selector references the outer element, which isn't in a where clause (as the case above), it doesn't translate to a database join. That's why we need to evaluate the collection selector for each outer element. It translates to `APPLY` operations in many relational databases. If the collection is empty for an outer element, then no results would be generated for that outer element. But if `DefaultIfEmpty` is applied on the collection selector then the outer element will be connected with a default value of the inner element. Because of this distinction, these kinds of queries translate to `CROSS APPLY` in the absence of `DefaultIfEmpty` and `OUTER APPLY` when `DefaultIfEmpty` is applied. Certain databases like SQLite don't support `APPLY` operators so this kind of query may not be translated.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+CROSS APPLY [Posts] AS [p]
+
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], ([b].[Url] + N'=>') + [p].[Title] AS [p]
+FROM [Blogs] AS [b]
+OUTER APPLY [Posts] AS [p]
+```
+
+## GroupBy
+
+LINQ GroupBy operator creates result of type `IGrouping<TKey, TElement>` where `TKey` and `TElement` could be arbitrary types. Further, `IGrouping` implements `IEnumerable<TElement>`, which means you can compose over using any LINQ operator after grouping. Since no database structure can represent `IGrouping`, Group By operator has no translation in most cases. When an aggregate operator is applied to each group, which gives back a scalar, it can be translated to SQL `GROUP BY` in relational databases. SQL `GROUP BY` is restrictive too. It requires you to group by only scalar values  projection can contain only grouping key columns or aggregate applied over any column. EF Core identifies this subset and translates it to server, as in the following example:
```
bricelam:
```suggestion
LINQ GroupBy operators create a result of type `IGrouping<TKey, TElement>` where `TKey` and `TElement` could be any arbitrary type. Furthermore, `IGrouping` implements `IEnumerable<TElement>`, which means you can compose over it using any LINQ operator after the grouping. Since no database structure can represent an `IGrouping`, Group By operator has no translation in most cases. When an aggregate operator is applied to each group, which returns a scalar, it can be translated to SQL `GROUP BY` in relational databases. SQL's `GROUP BY` is restrictive too. It requires you to group only by scalar values. The projection can only contain grouping key columns or any aggregate applied over a column. EF Core identifies this pattern and translates it to the server, as in the following example:
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#GroupBy)]
+
+```SQL
+SELECT [p].[AuthorId] AS [Key], COUNT(*) AS [Count]
+FROM [Posts] AS [p]
+GROUP BY [p].[AuthorId]
+```
+
+EF Core also translates queries where an aggregate operator on the grouping appears in Where or OrderBy (or other ordering) LINQ operators. It uses `HAVING` clause in SQL for where clause. The query part before applying GroupBy operator can be any complex query as long as it can be translated to server. Further, once you apply aggregate operators on a grouping query to remove groupings from resulting source, you can compose it like any other query.
```
bricelam:
```suggestion
EF Core also translates queries where an aggregate operator on the grouping appears in a Where or OrderBy (or other ordering) LINQ operator. It uses `HAVING` clauses in SQL for the where clause. The part of the query before applying the GroupBy operator can be any complex query as long as it can be translated to server. Furthermore, once you apply aggregate operators on a grouping query to remove groupings from the resulting source, you can compose on top of it like any other query.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+```SQL
+SELECT [p].[AuthorId] AS [Key], COUNT(*) AS [Count]
+FROM [Posts] AS [p]
+GROUP BY [p].[AuthorId]
+HAVING COUNT(*) > 0
+ORDER BY [p].[AuthorId]
+```
+
+The aggregate operators EF Core supports are following
```
bricelam:
```suggestion
The aggregate operators EF Core supports are as follows
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+- Average
+- Count
+- LongCount
+- Max
+- Min
+- Sum
+
+## Left Join
+
+While Left Join isn't a LINQ operator, relational database has concept of Left Join which is frequently used in queries. A particular pattern in LINQ query gives same result as `LEFT JOIN` on server. EF Core identifies such pattern and generate equivalent `LEFT JOIN` on server side. The pattern involves creating GroupJoin between both the data sources and then flattening out the grouping by using SelectMany operator with DefaultIfEmpty on the grouping source to match null when inner doesn't have related element. Following example shows how that pattern looks like what it generates.
```
bricelam:
```suggestion
While Left Join isn't a LINQ operator, relational databases have the concept of a Left Join which is frequently used in queries. A particular pattern in LINQ queries gives the same result as a `LEFT JOIN` on the server. EF Core identifies such patterns and generates the equivalent `LEFT JOIN` on the server side. The pattern involves creating a GroupJoin between both the data sources and then flattening out the grouping by using the SelectMany operator with DefaultIfEmpty on the grouping source to match null when the inner doesn't have a related element. The following example shows what that pattern looks like and what it generates.
```

entity-framework/core/querying/complex-query-operators.md
```diff
@@ -0,0 +1,128 @@
...
+
+[!code-csharp[Main](../../../samples/core/Querying/ComplexQuery/Sample.cs#LeftJoin)]
+
+```SQL
+SELECT [b].[BlogId], [b].[OwnerId], [b].[Rating], [b].[Url], [p].[PostId], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Rating], [p].[Title]
+FROM [Blogs] AS [b]
+LEFT JOIN [Posts] AS [p] ON [b].[BlogId] = [p].[BlogId]
+```
+
+Above pattern creates complex structure in expression tree. Because of that EF Core requires you to flatten out grouping results out of GroupJoin operator in immediate step. Even if the GroupJoin-DefaultIfEmpty-SelectMany is used but in different pattern, we may not identify as Left Join.
```
bricelam:
```suggestion
The above pattern creates a complex structure in the expression tree. Because of that, EF Core requires you to flatten out the grouping results of the GroupJoin operator in a step immediately following the operator. Even if the GroupJoin-DefaultIfEmpty-SelectMany is used but in a different pattern, we may not identify it as a Left Join.
```

entity-framework/core/querying/index.md
```diff
@@ -1,10 +1,32 @@
...
 
-Entity Framework Core uses Language Integrated Query (LINQ) to query data from the database. LINQ allows you to use C# (or your .NET language of choice) to write strongly typed queries based on your derived context and entity classes. A representation of the LINQ query is passed to the database provider, to be translated in database-specific query language (for example, SQL for a relational database). For more detailed information on how a query is processed, see [How Query Works](overview.md).
+Entity Framework Core uses Language Integrated Query (LINQ) to query data from the database. LINQ allows you to use C# (or your .NET language of choice) to write strongly typed queries. It uses your derived context and entity classes to reference database objects. EF Core passes a representation of the LINQ query to the database provider. Database providers in turn translate it to database-specific query language (for example, SQL for a relational database).
+
+> [!TIP]
+> You can view this article's [sample](https://github.com/aspnet/EntityFramework.Docs/tree/master/samples/core/Querying) on GitHub.
+
+Following shows a few examples to achieve common tasks with Entity Framework Core.
```
bricelam:
```suggestion
The following shows a few examples of how to achieve common tasks with Entity Framework Core.
```

entity-framework/core/querying/async.md
```diff
@@ -1,21 +1,21 @@
...
+> EF Core doesn't support multiple parallel operations being run on the same context instance. You should always wait for an operation to complete before beginning the next operation. This is typically done by using the `await` keyword on each async operation.
 
-Entity Framework Core provides a set of asynchronous extension methods that can be used as an alternative to the LINQ methods that cause a query to be executed and results returned. Examples include `ToListAsync()`, `ToArrayAsync()`, `SingleAsync()`, etc. There are not async versions of LINQ operators such as `Where(...)`, `OrderBy(...)`, etc. because these methods only build up the LINQ expression tree and do not cause the query to be executed in the database.
+Entity Framework Core provides a set of async extension methods similar to the LINQ methods, which execute a query and return results. Examples include `ToListAsync()`, `ToArrayAsync()`, `SingleAsync()`. There are no async versions of some LINQ operators such as `Where(...)`, `OrderBy(...)`, because these methods only build up the LINQ expression tree and don't cause the query to be executed in the database.
```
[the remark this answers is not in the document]
smitpatel: There are also other methods. How should we describe that?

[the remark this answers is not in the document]
smitpatel: Converted to
> some LINQ operators such as `Where(...)` or `OrderBy(...)`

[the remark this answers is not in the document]
smitpatel: Still 100 😉

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
 uid: core/querying/client-eval
 ---
 # Client vs. Server Evaluation
 
-Entity Framework Core supports parts of the query being evaluated on the client and parts of it being pushed to the database. It is up to the database provider to determine which parts of the query will be evaluated in the database.
+As a general rule, Entity Framework Core attempts to evaluate query on server as much as possible. EF Core converts parts of the query to parameters, which it can evaluate on client side. Rest of the query with generated parameters is given to database providers to determine what should be equivalent database query to do evaluation on the server. EF Core supports partial client evaluation in top-level projection (essentially, the last call to `Select()`). If parts of query in top-level projection can't be translated to server, EF Core will fetch required data from server and evaluate it on client. If EF Core detects expression, which can't be translated to server in any other place, then it throws runtime exception. See [how query works](xref:core/querying/how-query-works) to understand how EF Core determines what can't be translated to server.
```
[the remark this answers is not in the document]
smitpatel: also found another missing `the`

entity-framework/core/querying/client-eval.md
```diff
@@ -1,69 +1,62 @@
...
 
-> [!TIP]  
+> [!NOTE]
+> Prior to version 3.0, Entity Framework Core supported client evaluation anywhere in the query. For more information, see the [previous versions section](#previous-versions).
+
+## Client evaluation in top-level projection
```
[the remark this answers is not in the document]
smitpatel: Went with the since there is only one top-level projection. It's not incorrect just bit ambiguous.

entity-framework/core/querying/complex-query-operators.md
smitpatel: The page is not necessarily put here for relational. We can add more operators here as we go. Like Cosmos can translate bunch of SelectMany in awesome way. The main purpose in not putting relational here is to reuse this page as needed for other providers in future. Further, all the operators talk about the relational translation so it is evident that it applies to relational only.

roji: Re Cosmos translations specifically, we may want to consider talking about that in the Cosmos provider docs - not sure it's a good idea to pack all the provider-specific info in one page. It will not be relevant for most readers of this page. Otherwise the page as it is seems 100% relational to me, but I don't have super strong feelings on this.

roji: Yeah, there is one point on this long and elaborate page that is also common to non-relational providers.

As I said if you feel very strongly about not including a note on this, I definitely won't die on that hill.

PR #1803, opened 2019-10-02
Breaking change note on duplicated results in Include

### LINKED ISSUE #1769: Breaking changes note on new single-query behavior and cartesian explosion
Now that we generate a single SQL query for a LINQ query, users will start running into the cartesian explosion problem when joining multiple tables (e.g. https://github.com/aspnet/EntityFrameworkCore/issues/18022). ~We should add a section in the [loading related data page](https://docs.microsoft.com/en-us/ef/core/querying/related-data) with a code sample.~

Add a breaking change note.

roji:
Closes #1769

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
+++ b/entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
@@ -23,6 +23,7 @@ Breaks from one 3.0 preview to another 3.0 preview aren't documented here.
 | [Query types are consolidated with entity types](#qt) | High      |
 | [Entity Framework Core is no longer part of the ASP.NET Core shared framework](#no-longer) | Medium      |
 | [Cascade deletions now happen immediately by default](#cascade) | Medium      |
+| [Eager loading of related entities now happens in a single query](#eager-loading-single-query) | Medium      |
 | [DeleteBehavior.Restrict has cleaner semantics](#deletebehavior) | Medium      |
 | [Configuration API for owned type relationships has changed](#config) | Medium      |
 | [Each property uses independent in-memory integer key generation](#each) | Medium      |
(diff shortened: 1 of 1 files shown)


entity-framework/core/querying/related-data.md
```diff
@@ -47,8 +47,39 @@ You may want to include multiple related entities for one of the entities that i
...
+```
+
+Because of the way relational databases work, a customer with several addresses and/or orders will be duplicated in the database resultset; address fields will be duplicated as well, based on the number of orders and order details the same customer has, and so on. This may not be significant if the number of `Include` operators is low or if the database contains little data, but as additional collection navigations are included duplication quickly grows.
+
+The solution is to split this query into several smaller queries using [explicit loading](#explicit-loading), where each query includes a smaller number of collection navigations (possibly just one):
+
+``` csharp
+var baseQuery = db.Customers.Where(c => c.CustomerName == "John");
+var customers = baseQuery.ToList();  // Loads customers without any related entities
+baseQuery.SelectMany(c => c.Addresses).Load();  // Loads only the addresses
```
smitpatel: This is not explicit loading.

roji: How is it not [explicit loading](https://docs.microsoft.com/en-us/ef/core/querying/related-data#explicit-loading):

> You can also explicitly load a navigation property by executing a separate query that returns the related entities.

roji:
> Remove changes to related data section

Are you sure you want to spend your time writing this, is what I wrote insufficient? In any case it would have been better to change assignments in #1769 or communicate better to prevent the double work.

entity-framework/core/querying/related-data.md
smitpatel: Explicit load APIs are defined on NavigationEntry. This is not navigation entry. This is pure IQueryable. Load function here is whole lot different than explicit loading.

roji: OK, I guess it's terminological but the functional idea and outcome are the same: load related entities explicitly after loading the principal - that's even how our docs say it. Whether it's done on NavigationEntry or on the IQueryable doesn't seem very important (at least to users). But I don't care that much either way.

smitpatel:
#1769  was about adding breaking changes section only. If you are tracking both with it, you can assign it to me.

What is written in related data has a lot of inaccuracy. I am already working on query docs currently so related data section will also get updated.

entity-framework/core/querying/related-data.md
smitpatel: Outcome are altogether different. Explicit load API builds query with where predicate on key and can load only 1 entity's navigation at a time. You cannot even load reference navigation for items in the collection with that. At best using explicit loading for collection include gives you N + 1 queries and not 2 queries.

roji:
> #1769 was about adding breaking changes section only

Well it does have "We should add a section in the loading related data page with a code sample." but it's not important. I'll remove the related data section from this PR.

smitpatel:
Triage decision was about breaking change only. (At least that was my understanding). And related data section written is not correct.

roji:
Removed related data section.

AndriySvyryd:
Might be worth linking to the comment with the workaround: https://github.com/aspnet/EntityFrameworkCore/issues/18022#issuecomment-537219137

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -385,6 +386,29 @@ For example:
...
+
+Starting with 3.0, EF Core generates a single query with JOINs on relational databases.
+
+**Why**
+
+Issuing multiple queries to implement a single LINQ query caused numerous issues, including negative performance as multiple database roundtrips were necessary, and data coherency issues as each query could observe a different state of the database.
+
+**Mitigations**
+
+While technically this is not a breaking change, it could have a considerable effect on application performance when a single query contains a large number of `Include` operator on collection navigations. [See this comment](https://github.com/aspnet/EntityFrameworkCore/issues/18022#issuecomment-535102526) for more information and for rewriting queries in a more efficient way.
```
smitpatel: As pointed out by @AndriySvyryd - This comment is incorrect link. :frowning_face:

PR #1807, opened 2019-10-03
Add breaking change note for FromSql with sproc

smitpatel:


entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -207,6 +208,35 @@ This could result in queries not being parameterized when they should have been.
...
 Switch to use the new method names.
 
+<a name="fromsqlsproc"></a>
+### FromSql method when used with stored procedure cannot be composed
+
+[Tracking Issue #15392](https://github.com/aspnet/EntityFrameworkCore/issues/15392)
+
+**Old behavior**
+
+Before EF Core 3.0, FromSql method tried to detect if the passed SQL can be composed upon. It did client evaluation when the SQL was non-composable like stored procedure. Following query worked by running stored procedure on server and doing FirstOrDefault on client side.
```
AndriySvyryd:
```suggestion
Before EF Core 3.0, FromSql method tried to detect if the passed SQL can be composed upon. It did client evaluation when the SQL was non-composable like a stored procedure. The following query worked by running the stored procedure on the server and doing FirstOrDefault on the client side.
```

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -207,6 +208,35 @@ This could result in queries not being parameterized when they should have been.
...
+
+```C#
+context.Products.FromSqlRaw("[dbo].[Ten Most Expensive Products]").FirstOrDefault();
+```
+
+**New behavior**
+
+Starting with EF Core 3.0, EF Core will not try to parse the SQL. So if you are composing after FromSqlRaw/FromSqlInterpolated, then EF Core will compose the SQL by causing sub query. So if you are using stored procedure with composition then you will get exception for invalid SQL syntax.
```
AndriySvyryd:
```suggestion
Starting with EF Core 3.0, EF Core will not try to parse the SQL. So if you are composing after FromSqlRaw/FromSqlInterpolated, then EF Core will compose the SQL by causing sub query. So if you are using a stored procedure with composition then you will get an exception for invalid SQL syntax.
```

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -207,6 +208,35 @@ This could result in queries not being parameterized when they should have been.
...
+
+**Why**
+
+Since EF Core 3.0 does not support client evaluation parsing the SQL, which was error prone, to determine if it is composable provided little value. The query would fail either way.
```
AndriySvyryd:
```suggestion
EF Core 3.0 does not support automatic client evaluation, since it was error prone as explained [here](#linq-queries-are-no-longer-evaluated-on-the-client).
```

entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md
```diff
@@ -207,6 +208,35 @@ This could result in queries not being parameterized when they should have been.
...
+
+**Mitigation**
+
+If you are using stored procedure in FromSqlRaw/FromSqlInterpolated, you know that it cannot be composed upon, you can put `AsEnumerable`/`AsAsyncEnumerable` right after FromSql method call to avoid any composition to be done on server side.
```
AndriySvyryd:
```suggestion
If you are using a stored procedure in FromSqlRaw/FromSqlInterpolated, you know that it cannot be composed upon, so you can add __AsEnumerable/AsAsyncEnumerable__ right after the FromSql method call to avoid any composition on server side.
```

PR #1816, opened 2019-10-04
Fix for syntax error in code sample in local-data.md

mojogeorge:
Fix for syntax error in code sample for title "Using Local to add and remove entities from the context  "


smitpatel:
@mojogeorge - Thank you.

PR #1820, opened 2019-10-05
 Use IANA example.com domain

vanillajonathan:
Use the preferred IANA example.com domain for examples.


smitpatel:
@vanillajonathan - Thank you.

PR #1827, opened 2019-10-08
A title should be: this is article title - this is a brand/product name.

smitpatel:
CATS report


smitpatel:
@bricelam - Only docs people can answer that.

PR #1842, opened 2019-10-13
added FileContextCore as Database Provider

morrisjdev:
Hi.
I created a small database provider that stores the data of a Context in files (json, xml, csv etc.).
I think it is interesting for development.
Is it ok for you to include it in the official docs?


ajcvickers:
@morrisjdev Thanks!

PR #1849, opened 2019-10-14
Added Teradata.EntityFrameworkCore provider

MaxG117:
1. Added Teradata.EntityFrameworkCore.
~~2. The database providers weren't sorted in any logical order, so I sorted the list alphabetically by NuGet Package name, with Microsoft.EntityFrameworkCore.* packages listed first due to being part of the EF Core Project.~~


bricelam:
My ~~two~~ **three** cents:
1. Alphabetical seems useless, if I'm looking for a specific provider, I'm going to Ctrl+F for the database engine I'm looking for.
2. Most popular is probably more useful, but there's also value in keeping first-party providers first since they're always up-to-date
3. When is a provider too out-of-date to be on this list? EF Core 1.1, 1.0, are out of support and providers written against them won't work in newer versions.

MaxG117:
1. Feel free to suggest an order or arrange in any way you see fit. I considered ordering by the database engine, but some providers support multiple engines.
2. I agree and considered this as well. It seems that not all first-party providers reserved their NuGet ID prefix, so verifying the first-party may require additional steps.
3. Agreed -- perhaps the table should have another column to indicate the supported version of EFCore, and then we could use that as the first sort column.

MaxG117:
Speaking of "out of support", EF Core 2.2 will be out of support in December, but it contains breaking changes and new features that aren't available in the EF Core 2.1 that is LTS and will be supported for 2 more years. This does not make sense -- 2.2 should either not have had any breaking changes and new features, or it should have been made LTS (along with 2.1).

ajcvickers:
@bricelam I think we discussed this back when we were doing docs planning. However, I don't remember (and can't find) what we decided. Here are my thoughts:
* Providers that work with the latest release should be given preference over providers that don't
* We have the highest confidence that this is true for the providers we ship, so I think it makes sense for those to be listed first
* For other providers in the general case we don't have any real way of knowing how well they work, or whether they work at all with the latest release
  * However, we do know than any 2.x provider won't work with 3.x unless it has been specifically, updated.
* Therefore, I think it makes sense to have providers list include the latest release that the provider author asserts that the provider works with.
* Providers that have been asserted to work with, for example, 3.0, then go above those not asserted to work 3.0.
* Within these sub-lists we can order any way--alphabetical is fine.

Does that make sense? Am I making this too complicated?

ajcvickers:
@MaxG117 2.2 does not contain any intentional breaking changes from 2.1. That being said, I agree that 2.2 not being LTS is not ideal. Hopefully #18141 will make it possible for more people to move to 3.1 before 2.2 goes out of support.

MaxG117:
@ajcvickers 

> I agree that 2.2 not being LTS is not ideal. 
That's good to hear. I targeted provider development to 2.2 because it had new features without realizing it would go out of support _before_ 2.1. As a slower-moving database provider developer, I'm now inclined to skip all the non-LTS releases entirely. This means that in the future I'm likely to skip 5.0 and 7.0, and I'm probably not the only one. Something for y'all to keep in mind.

> Hopefully #18141 will make it possible for more people to move to 3.1 before 2.2 goes out of support.

More, maybe. I'm more concerned about the changes around Query.

MaxG117:
@ajcvickers @bricelam 

So do you want me to reorder and resubmit/update the PR or are you going to do this? My more immediate goal was to get the Teradata provider listed, I just didn't know where to insert it.

ajcvickers:
@MaxG117 Feel free to submit a PR that adds it to the end of the list without doing any ordering changes. I filed #1899 for the ordering.

MaxG117:
@ajcvickers I updated this PR to add my provider to the end of the list without reordering.

ajcvickers:
@MaxG117 I'm just coming back to this now; sorry for the long wait. I have merged #1864 from @ErikEJ . Now I'm going to do some cleanup, then hopefully we can rebase this on top of the results.

PR #1873, opened 2019-10-22
Refined documentation by adding and removing few character

gurmeetsinghdke:
At some places dot was used instead of colons- fixed those.
At one line (in the end) a bar character ('|') was present, removed that.


ajcvickers:
Thanks!

PR #1877, opened 2019-10-24
Update index.md

OOberoi:
Simplified the paragraph with minor corrections especially from readability point of view.


OOberoi:
Thank you Smit!

OOberoi:
@smitpatel @ajcvickers - Noticed that the change hasn't made it to the production yet!! Do I need to create another PR?

PR #1892, opened 2019-11-04
Update ef-core-2.0.md

vanillajonathan:


ajcvickers:
@vanillajonathan Are the changes here only to add backticks on type/member names? If so, this isn't a change we are planning to make because unfortunately the reading experience for backticked words is too jarring to use everywhere.

vanillajonathan:
@ajcvickers Well the backticks should probably be present on the first mention of a property, method, class or interface. Then on subsequent mentions there is less need for it to be marked up in backticks if that is considering too jarring.

The commit does other things too, such as consistent capitalization of headers, hyphenation between words, and proper 4-space indention for code.

ajcvickers:
@vanillajonathan The intention is to only use backticks when we want to explicitly call attention to something being a type/member names. I don't think this is the case for most of the changes here.

vanillajonathan:
@ajcvickers I reverted the backticks that I added.

ajcvickers:
Thanks!

PR #1903, opened 2019-11-06
Replace all double quotes (”) by a normal double quote character (").

StefH:
Replace all double quotes (”) by a normal double quote character (").
This makes sure that the commands can be copy-pasted into the command prompt.

Related to https://github.com/aspnet/EntityFramework.Docs/issues/1901


ajcvickers:
Thanks!

PR #1907, opened 2019-11-07
Clarify how primary keys interact with value generation

### LINKED ISSUE #1046: Primary key types
It would be nice to have a discussion of the recommended type choice(s) for primary keys. The docs here use string but what about int, uint, long, and so on? Are there types that are not allowed as primary keys?

---
#### Document Details

* ID: d0a084df-15c4-ab25-26ca-83c8bdf6f026
* Version Independent ID: f1fba5b2-eb2e-da0a-fce0-d492855c1777
* Content: [Keys (primary) - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/keys)
* Content Source: [entity-framework/core/modeling/keys.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/keys.md)
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

AndriySvyryd:
Fixes #1046


entity-framework/core/modeling/generated-properties.md
```diff
@@ -48,7 +49,7 @@ Like `value generated on add`, if you specify a value for the property on a newl
 
 ## Conventions
 
-By convention, non-composite primary keys of type short, int, long, or Guid will be setup to have values generated on add. All other properties will be setup with no value generation.
+By default, non-composite primary keys of type short, int, long, or Guid will be setup to have values generated on add. All other properties will be setup with no value generation.
```
ajcvickers: Is it still true that you don't get value generation if the property is part of a composite? I vaguely remember that we (possibly unintentionally) changed this.

AndriySvyryd: Yes, we changed this back for non-owned types. And I don't think we need to talk about owned types here.

PR #1954, opened 2019-12-02
Update parameter name for clarity

rmarskell:
Updated the `HasOne` parameter name to be consistent with `HasForeignKey` so as not to cause confusion. Resolves https://github.com/aspnet/EntityFramework.Docs/issues/1626


ajcvickers:
Thanks!

PR #1995, opened 2019-12-17
Updates to provider/extensions docs

### LINKED ISSUE #1899: Order the providers/extensions
From https://github.com/aspnet/EntityFramework.Docs/pull/1849#issuecomment-542876404

* Providers that work with the latest release should be given preference over providers that don't
* We have the highest confidence that this is true for the providers we ship, so I think it makes sense for those to be listed first
* For other providers in the general case we don't have any real way of knowing how well they work, or whether they work at all with the latest release
  * However, we do know than any 2.x provider won't work with 3.x unless it has been specifically, updated.
* Therefore, I think it makes sense to have providers list include the latest release that the provider author asserts that the provider works with.
* Providers that have been asserted to work with, for example, 3.0, then go above those not asserted to work 3.0.
* Within these sub-lists we can order any way--alphabetical is fine.

ajcvickers:
Fixes #1899

Most significantly, adding the version that the provider was built against and ordering by that version.


entity-framework/core/providers/index.md
```diff
@@ -15,32 +14,35 @@ Entity Framework Core can access many different databases through plug-in librar
...
-| [Teradata.EntityFrameworkCore](https://www.nuget.org/packages/Teradata.EntityFrameworkCore/)                         | Teradata Database 16.10 onwards | [Teradata](https://downloads.teradata.com/download/connectivity/net-data-provider-for-teradata) | Prerelease               | [website](https://www.nuget.org/packages/Teradata.EntityFrameworkCore/)                                                                                                                            |
+> [!IMPORTANT]  
+> EF Core providers typically work across minor versions, but not across major versions. For example, a provider released for EF Core 2.1 should work with EF Core 2.2, but will not work with EF Core 3.0. 
+
+| NuGet Package                                                                                                        | Supported database engines | Maintainer / Vendor                                                           | Notes / Requirements | Built for | Useful links                                                                                                                                                                                       |
```
roji: Maybe "Built for Version", or maybe just "Version" - built for is a bit vague.

Issue #1669, opened 2019-08-24
Merge relational database modeling into general modeling


roji:
As discussed offline, we should consider removing the separate "relational database modeling" subsection and merge its contents into general modeling. 99% of EF Core users are relational (so information is split across two places), and the split between relational and non-relational isn't very meaningful or interesting to the vast majority of users.

Where we think it's really important to highlight that something is relational, we can call it out.

AndriySvyryd:
Related: https://github.com/aspnet/EntityFramework.Docs/issues/862

roji:
PRs are out merging all relational model building pages into non-relational.

Issue #1398, opened 2019-03-27
Update guidance on using default values


ajcvickers:
See discussion here: https://github.com/aspnet/EntityFrameworkCore/issues/15070#issuecomment-477220843

* For hand-written code document:
  * Use client side default values. This ensures the defaults are consistent in the domain model.
  * Show how store-generated values work in the "normal" case--e.g. with integers
  * Show how to use option B for when both the normal way and client-side defaults are not appropriate, also mentioning C
  * Show how to get the default value into the migration if it needs to be there

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1f32ff95-761b-279a-f5a9-7fe859121356
* Version Independent ID: 63cc902d-8261-8185-d0a0-c48c627234b1
* Content: [Default Values - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/relational/default-values)
* Content Source: [entity-framework/core/modeling/relational/default-values.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/relational/default-values.md)
* Product: **entity-framework**
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

Sergey-Nosov-GH:
Should setting default values in the parameter-less constructor of the underlying class also be an option?

For example:
```cs
class MyContext : DbContext
{
    public DbSet<Blog> Blogs { get; set; }
}

public class Blog
{
    public Blog()
    {
        Rating = 3;
    }
    public int BlogId { get; set; }
    public string Url { get; set; }
    public int Rating { get; set; }
}
```

ajcvickers:
@Sergey-Nosov-GH Can you give some details on why this is preferable to:
```C#
public class Blog
{
    public int BlogId { get; set; }
    public string Url { get; set; }
    public int Rating { get; set; } = 3;
}
```

Sergey-Nosov-GH:
@ajcvickers those are functionally equivalent, so your example would be fine. Nether is currently in the docs, and that is confusing. I came to this discussion from the following url:

[https://docs.microsoft.com/en-us/ef/core/modeling/relational/default-values](url)

As of right now, it has no mentions of defaults set on the underlying classes; and it makes the programmer wonder, if EF overrides such defaults, goes along with them, or what happens. Thus, more guidance would be helpful.

Thank you.

ajcvickers:
[...]

PR #2003, opened 2019-12-19
Merge default values and computed columns into generated properties

roji:
Plus some cleanup.

Part of #1669

PS I plan to work some more on this page, specifically to merge the introductory "patterns" with the configuration section, and to try to make things a bit lighter/terser. But that's for another PR.


entity-framework/core/modeling/generated-properties.md
```diff
@@ -47,48 +47,73 @@ Like `value generated on add`, if you specify a value for the property on a newl
 >
 > [!code-sql[Main](../../../samples/core/Modeling/FluentAPI/ValueGeneratedOnAddOrUpdate.sql)]
 
-## Conventions
+## Value generated on add
+
+By convention, non-composite primary keys of type short, int, long, or Guid are be set up to have values generated on add, and your database provider typically takes care of the necessary configuration. For example, an numeric primary key in SQL Server is automatically set up to be an IDENTITY column.
```
ajcvickers: It's weird to see "add" written like this. Consider "to have values generated for entities that will be inserted".

entity-framework/core/modeling/generated-properties.md
```diff
@@ -47,48 +47,73 @@ Like `value generated on add`, if you specify a value for the property on a newl
...
 
-By default, non-composite primary keys of type short, int, long, or Guid will be setup to have values generated on add. All other properties will be setup with no value generation.
+You can configure any property to be value generated on add as follows:
```
ajcvickers: Likewise, weird use of "add".

entity-framework/core/modeling/generated-properties.md
```diff
@@ -47,48 +47,73 @@ Like `value generated on add`, if you specify a value for the property on a newl
...
 
-> [!WARNING]  
+> [!WARNING]
 > This just lets EF know that values are generated for added entities, it does not guarantee that EF will setup the actual mechanism to generate values. See [Value generated on add](#value-generated-on-add) section for more details.
 
-### Value generated on add or update (Data Annotations)
+### Default values
 
-[!code-csharp[Main](../../../samples/core/Modeling/DataAnnotations/ValueGeneratedOnAddOrUpdate.cs#Sample)]
+On relational databases, a column can be configured with a default value; if a row is inserted without a value for that column, the default value will be used.
```
ajcvickers: This will likely cause confusion. May be okay to leave for now and fix as part of #1398

entity-framework/core/modeling/generated-properties.md
roji: Agree. Changed to the slightly shorter "to have values generated for inserted entities".

entity-framework/core/modeling/generated-properties.md
roji: OK. I plan to do some work on this page as well, specifically to merge the first "conceptual" pass over the value generation patterns, and the second "configuration" pass. I don't think it makes sense to talk about "value generated on add" twice in the same page, once conceptually and once concretely - so I'll try to find a good way to merge them.

We can also look improving the default column guidance as part of #1398, yeah.

PR #2005, opened 2019-12-20
Improve backing field modeling page

roji:
* Reorganize content based on topics and not convention/fluent API/data annotations, as done in #1669.
* Correct the documentation of the default property access mode (change from PreferFieldDuringConstruction to PreferField in 3.0), and add a note on the behavior change.
* Various other cleanups and improvements.

### DIFF
--- a/entity-framework/core/modeling/backing-field.md
+++ b/entity-framework/core/modeling/backing-field.md
@@ -41,7 +41,7 @@ See the [PropertyAccessMode enum](https://docs.microsoft.com/dotnet/api/microsof
 
 ## Field-only properties
 
-You can also create a conceptual property in your model that does not have a corresponding CLR property in the entity class, but instead uses a field to store the data in the entity. This is different from [Shadow Properties](shadow-properties.md), where the data is stored in the change tracker, rather than in the entity's CLR type. Field-only properties are typically used when the entity class uses methods to get/set values, instead of properties.
+You can also create a conceptual property in your model that does not have a corresponding CLR property in the entity class, but instead uses a field to store the data in the entity. This is different from [Shadow Properties](shadow-properties.md), where the data is stored in the change tracker, rather than in the entity's CLR type. Field-only properties are commonly used when the entity class uses methods instead of properties to get/set values, or in cases where fields shouldn't be exposed at all in the domain model (e.g. primary keys).
 
 You can configure a field-only property by providing a name in the `Property(...)` API:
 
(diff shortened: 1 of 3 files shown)


entity-framework/core/modeling/backing-field.md
```diff
@@ -7,49 +7,49 @@ uid: core/modeling/backing-field
...
+You can also configure backing fields explicitly, e.g. if the field name doesn't correspond to the above conventions:
 
-## Fluent API
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/BackingField.cs?name=BackingField&highlight=5)]
 
-You can use the Fluent API to configure a backing field for a property.
+## Field and property access
 
-[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/BackingField.cs#Sample)]
+By default, EF will always read and write to the backing field - assuming one has been properly configured - and will never use the property. However, EF also supports other access patterns. For example, the following sample instructs EF to write to the backing field only while materializing, and to use the property in all other cases:
```
ajcvickers: I think this is misleading. By default, EF will always read and write to the backing field so long as the backing field has been discovered or configured as described above. The property will be used if the backing field is not known.

entity-framework/core/modeling/backing-field.md
```diff
@@ -7,49 +7,49 @@ uid: core/modeling/backing-field
...
+See the [PropertyAccessMode enum](https://docs.microsoft.com/dotnet/api/microsoft.entityframeworkcore.propertyaccessmode) for the complete set of supported options.
 
-[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/BackingFieldAccessMode.cs#Sample)]
+> [!NOTE]
+> With EF Core 3.0, the default property access mode changed from `PreferFieldDuringConstruction` to `PreferField`.
 
-### Fields without a property
+## Field-only properties
 
-You can also create a conceptual property in your model that does not have a corresponding CLR property in the entity class, but instead uses a field to store the data in the entity. This is different from [Shadow Properties](shadow-properties.md), where the data is stored in the change tracker. This would typically be used if the entity class uses methods to get/set values.
```
ajcvickers: I would not necessarily call this the "typical" case, even though it is a common case. Another common case is for fields that should not be exposed at all in the domain model, such as primary keys.

entity-framework/core/modeling/backing-field.md
roji: Do you think that "assuming one has been properly configured" isn't sufficient here? Maybe "assuming one has been properly discovered or configured", or something else?

entity-framework/core/modeling/backing-field.md
roji: Pushed a change, let me know what you think.

PR #2007, opened 2019-12-21
Remaining modeling doc cleanup, remove relational modeling section

### LINKED ISSUE #1669: Merge relational database modeling into general modeling

roji:
This is the last now for now, promise :)

Closes #1669


roji:
@AndriySvyryd thanks, updated the dates and set you as the author.

PR #2022, opened 2020-01-03
Note about API surface being null-oblivious

roji:
See https://github.com/aspnet/EntityFrameworkCore/issues/19443


entity-framework/core/miscellaneous/nullable-reference-types.md
```diff
@@ -57,6 +57,7 @@ A similar issue occurs when including multiple levels of relationships across op
...
+* Reverse engineering does not currently support [C# 8 nullable reference types](/dotnet/csharp/tutorials/nullable-reference-types): EF Core always generates C# code that assumes the feature is off. For example, nullable text columns will be scaffolded as a property with type `string` , not `string?`, with either the Fluent API or Data Annotations used to configure whether a property is required or not. You can edit the scaffolded code and replace these with C# nullability annotations. Scaffolding support for nullable reference types is tracked by issue [#15520](https://github.com/aspnet/EntityFrameworkCore/issues/15520).
```
cartermp:
```suggestion
* Reverse engineering does not currently support [C# 8 nullable reference types (NRTs)](/dotnet/csharp/tutorials/nullable-reference-types): EF Core always generates C# code that assumes the feature is off. For example, nullable text columns will be scaffolded as a property with type `string` , not `string?`, with either the Fluent API or Data Annotations used to configure whether a property is required or not. You can edit the scaffolded code and replace these with C# nullability annotations. Scaffolding support for nullable reference types is tracked by issue [#15520](https://github.com/aspnet/EntityFrameworkCore/issues/15520).
```

cartermp: Added the abbreviation here because it's used further below

PR #2046, opened 2020-01-14
Sentence completion

OOberoi:
Added fillers to complete the sentence. Kindly review!

### DIFF
--- a/entity-framework/core/providers/cosmos/unstructured-data.md
+++ b/entity-framework/core/providers/cosmos/unstructured-data.md
@@ -49,7 +49,7 @@ To decouple completely from EF Core get the [CosmosClient](/dotnet/api/Microsoft
 
 ## Missing property values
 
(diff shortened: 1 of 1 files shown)


ajcvickers:
@OOberoi The sentence seems correct to me before your changes and not after. Can you provide a few more details as to the purpose of this change?

OOberoi:
Hey Arthur, I read the description for "Missing property values" couple more times. Even though, you could pass it as is, however, I would word it as the following with minor changes. That being said, I wholeheartedly and respectfully would leave it at your discretion.

"In the previous example, we removed the "TrackingNumber" property from the order object. Because of how indexing works in Cosmos DB, queries that reference the missing property somewhere else, than in the projection, it could return unexpected results. For example:

ajcvickers:
@OOberoi We reviewed as a team and agreed to keep the current text. Thanks anyway!

OOberoi:
No worries at all. Thanks Arthur!

PR #2071, opened 2020-01-29
Updates to release planning

ajcvickers:
Specifically added text about how to read the backlog and the process for deciding what to patch.

### DIFF
--- a/entity-framework/core/what-is-new/release-planning.md
+++ b/entity-framework/core/what-is-new/release-planning.md
@@ -1,21 +1,115 @@
 ---
 title: EF Core release planning
 author: ajcvickers
-ms.date: 01/14/2020
+ms.date: 01/28/2020
 uid: core/what-is-new/release_planning.md
 ---
 
 # Release planning process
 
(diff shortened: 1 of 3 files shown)


entity-framework/core/what-is-new/release-planning.md
```diff
@@ -1,45 +1,122 @@
...
+### Major releases
+
+Major releases change the EF "major" version number.
+For example, EF Core **3**.0.0 is a major release that makes a big step forward over EF Core 2.2.x.
+
+Major releases:
+* Are intended to improve the quality and features of the previous release
+* Typically contain bug fixes and new features
+  * Some of the new features may be fundamental changes to the way EF Core works
+* Typically include intentional breaking changes
```
bricelam: I'd like to say we strive to avoid changes that will break applications and to minimize changes that will break providers and extensions, but I'm not sure that's historically accurate.

ajcvickers: We could say, "We may have been too aggressive with breaking changes in the past. Going forward, we will strive..."

PR #2081, opened 2020-01-31
Move Principal key definition before Foreign key

CGijbels:
When defining the different terms used, the Foreign key is defined before the Principal key although the former mentions the latter, switching positions might be clearer for the reader


ajcvickers:
Thanks!

PR #2100, opened 2020-02-08
Use example.com instead of sample.com

amrbadawy:
example.com is for use in illustrative examples in documents. but sample.com is a real domain


ajcvickers:
Thanks!

PR #2101, opened 2020-02-08
Remove outdated information

loic-sharma:
ASP.NET Core 3.0 does not automatically reference `Microsoft.EntityFrameworkCore.Design`. I did not mention that this is not necessary on ASP.NET Core 2.2 and earlier as this may confuse customers, and they'll need to add this reference once they upgrade anyways.


ajcvickers:
Thanks!

PR #2102, opened 2020-02-08
Update existing-database.md

SimpleSamples:
Sometimes it is not clear whether to say "make use of" or simply "use" but here "make use of" really sounded off. I changed that and a few minor other things.


ajcvickers:
Thanks!

PR #2074, opened 2020-01-30
Update Program.cs for correctly showing field mapping

yadavvineet:
In the new 3.1, field mapping is changed which is not reflected in docs. updating the doc to highlight it accordingly.

Fix #1961

PR #2114, opened 2020-02-14
Fix QueryFilters sample program

Costo:
In #2074, the name of the property "TenantId" was changed to "_tenantId", but the change was not propagated throughout the code, so the sample was not running anymore.


ajcvickers:
Thanks!

PR #2125, opened 2020-02-19
Update plan.md to clarify many-to-many

vslee:
There are multiple GH issues tracking the various components of many-to-many. Previously, only one of them was mentioned here. I have added the rest. Please feel free to adjust/edit as you see fit. (I couldn't find an issue for the configuration sugar)
I have also added additional text to clarify which issues are planned and which are only stretch goals.


ajcvickers:
@vslee I like the idea here! Not merging immediately because I want to think about the exact language and formatting, since this is an important doc.

vslee:
Yes of course. These changes make the style of that section different than the others, so they likely need to be revised to conform better.

ajcvickers:
This now merged. Thanks!

PR #2132, opened 2020-02-21
Update repo org. references

scottaddie:
Replace occurrences of *aspnet/EntityFramework.Docs* with *dotnet/EntityFramework.Docs*.


smitpatel:
Can you also look into the warning? Some thing needs to be updated in OPS it seems.

scottaddie:
@smitpatel I'm looking into that warning right now.

scottaddie:
@smitpatel We shouldn't see that warning in future builds. I've added Arthur as an admin in the repo.

PR #2136, opened 2020-02-23
Simple Windows Forms application

ajcvickers:
This is a starting point for documentation/guidance for using EF Core in Windows Forms applications.

I created this sample by working through the steps in the [EF6 Windows Forms walk-through](https://docs.microsoft.com/en-us/ef/ef6/fundamentals/databinding/winforms) with the following changes:

* NuGet package install:
  * Microsoft.EntityFrameworkCore.Sqlite - Sqlite is easy to use when testing; all this should also work with other types of database
  * Microsoft.EntityFrameworkCore.Proxies - Because the EF6 code uses lazy-loading proxies; we should consider changing this
* _Implementing IListSource for Collections_ is not needed
  * Instead use ObservableCollectionListSource for navigation properties, which is included in EF Core.
  * See also: https://github.com/dotnet/efcore/issues/20034
* Define a model
  * Entity types are the same as for the Code First model from EF6, except using ObservableCollectionListSource
  * Context has OnConfiguring added to use a local SQLite database
* Lazy Loading
  * Added UseLazyLoadingProxies to the context OnConfiguring
* Bind Object to Controls
  * This is the same for EF Core as it is for EF6 (I did this in a .NET Framework app so I could use the designer :-))
* Add the Code that Handles Data Interaction
  * The same as for EF6, except the section on deleting orphans is not needed because EF Core does this by default :-)


ErikEJ:
Nice to see how much better EF Core is with WinForms than EF6!! 

Will there be a related article as well?

ajcvickers:
@ErikEJ Yeah. Also, the experience is close enough to EF6 that porting apps shouldn't be too hard.

We are starting to work with @OliaG and the Windows Forms team to figure out the correct guidance. This is only a starting point.

samples/core/SampleApps/SimpleWindowsFormsApp/Category.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace SimpleWindowsFormsApp
+{
+    public class Category
+    {
+        private readonly ObservableCollectionListSource<Product> _products =
+                new ObservableCollectionListSource<Product>();
+
+        public int CategoryId { get; set; }
+        public string Name { get; set; }
+        public virtual ObservableCollectionListSource<Product> Products { get { return _products; } }
```
bricelam: Could use an auto property:
```cs
Products { get; } = new ObservableCollectionListSource<Product>()
```

ajcvickers: Yeah. I was going to do this in another PR. I wanted to change as little as possible from the EF6 version for the initial checkin.

PR #2171, opened 2020-03-10
fix: EF core depends on .NET Standard 2.0.

prog-rajkamal:
as per nuget.org's package description https://www.nuget.org/packages/Microsoft.EntityFrameworkCore, EF Core depends on NET Standard 2.0, instead of .NET Standard 2.1 as originally mentioned


ajcvickers:
@prog-rajkamal Thanks for catching this.

Issue #2025, opened 2020-01-03
Document best practices for EF Core in Blazor applications


VR-Architect:
Can MS provide guidance and code to Avoiding DbContext threading issues when using Blazor server-side? 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a47bf610-692d-9d40-d85f-aec1e2a87a07
* Version Independent ID: 3c94634a-4b17-bc13-7436-0c61aea03ab6
* Content: [Configuring a DbContext - EF Core](https://docs.microsoft.com/en-us/ef/core/miscellaneous/configuring-dbcontext#avoiding-dbcontext-threading-issues)
* Content Source: [entity-framework/core/miscellaneous/configuring-dbcontext.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/miscellaneous/configuring-dbcontext.md)
* Product: **entity-framework**
* Technology: **entity-framework-core**
* GitHub Login: @rowanmiller
* Microsoft Alias: **avickers**

MisinformedDNA:
I'm having problems here too. The documented guidance is not enough. The error happens anyway.

vincentw56:
Not sure if this will help, but adding the code below fixed this issue for me.

'services.AddDbContext<MirwDbContext>(); //MirwDbContext is the name of by DB Context'

rkone:
The struggle for me at least, was sharing a dbcontext with more than 1 blazor component per page. A component and its subcomponent (and I assume other subcomponents) will run OnInitializedAsync() in parallel, leading to a concurrency error if they share the same dbcontext.  Best practice then, is to design your components so they can use a separate dbcontext. 
This is definitely an issue.  With this still fresh in my mind, the only solutions I have come up with so far was to join everything into 1 supercomponent, or to use different dbcontexts but then orchestrate a save among them all.  
I will be looking further into how to minimize the "editing" dbcontext, and using components that work with self-contained dbcontexts (and don't need to save).  Definitely a big change in process, coming from Silverlight where it was fine to load everything into one great big dbcontext...

ajcvickers:
@JeremyLikness I think we can close this now, right?

JeremyLikness:
@ajcvickers In my opinion, yes.

PR #2201, opened 2020-03-20
Explained Blazor Server specifics for configuring-dbcontext.md

akovac35:
For Blazor Server hosting model, one logical request is used for maintaining the Blazor user circuit, and thus only one scoped DbContext instance is available per user circuit if the default injection scope is used.

### DIFF
--- a/entity-framework/core/miscellaneous/configuring-dbcontext.md
+++ b/entity-framework/core/miscellaneous/configuring-dbcontext.md
@@ -180,11 +180,11 @@ Always await EF Core asynchronous methods immediately.
 
 The [`AddDbContext`](https://docs.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.entityframeworkservicecollectionextensions.adddbcontext) extension method registers `DbContext` types with a [scoped lifetime](https://docs.microsoft.com/aspnet/core/fundamentals/dependency-injection#service-lifetimes) by default.
 
-This is safe from concurrent access issues in ASP.NET Core applications because there is only one thread executing each client request at a given time, and because each request gets a separate dependency injection scope (and therefore a separate `DbContext` instance).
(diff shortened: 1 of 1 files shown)


roji:
Thanks for this - ideally guidance would be provided for how to deal with this, including samples (e.g. register the context as transient?). Also, we'd need to document the situation for client-side Blazor, not just server-side.

Finally, this guidance probably belongs more on the ASP.NET side, with links from EF Core.

smitpatel:
Should live as tutorial alongside this https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/intro?view=aspnetcore-3.1&tabs=visual-studio

akovac35:
> Thanks for this - ideally guidance would be provided for how to deal with this, including samples (e.g. register the context as transient?). Also, we'd need to document the situation for client-side Blazor, not just server-side.
> 
> Finally, this guidance probably belongs more on the ASP.NET side, with links from EF Core.

Guidance is already provided just a few lines down from my update:

---------------------------------------------------
However any code that explicitly executes multiple threads in parallel should ensure that DbContext instances aren't ever accessed concurrently.

*Using dependency injection, this can be achieved by either registering the context as scoped and creating scopes (using IServiceScopeFactory) for each thread, or by registering the DbContext as transient (using the overload of AddDbContext which takes a ServiceLifetime parameter).*

---------------------------------------------------
It makes sense to have Blazor specific detail mentioned on this particular page as is. Perhaps my one-line suggestion could be rewritten to be more generic (other changes should remain for consistency):

*A specific use case where additional consideration is required is scoped dependency injection in Blazor applications: [ASP.NET Core Blazor dependency injection](https://docs.microsoft.com/en-us/aspnet/core/blazor/dependency-injection?view=aspnetcore-3.1#service-lifetime)*

ajcvickers:
@akovac35 Thanks for submitting this. We're going to merge it because it improves what we currently have. However, #2025 covers having much better guidance here.

PR #2276, opened 2020-04-12
New EF Hub and TOC update

wadepickett:
- Add new EF hub page design - index.yml
- Update TOC to point to new index.yml
- Remove old hub index.md
-


wadepickett:
Unable to delete index.md and add index.yml at the same time without errors.  Closing and will try just adding the index.yml, publish that that in a separate pr, remove index.md

PR #2295, opened 2020-04-15
Update index.yml

JeremyLikness:
Modify description to list more databases, and list providers for PostgreSQL and MySQL on "Supported databases" card

### DIFF
--- a/entity-framework/index.yml
+++ b/entity-framework/index.yml
@@ -1,7 +1,7 @@
 ### YamlMime:Hub
 
 title: Entity Framework documentation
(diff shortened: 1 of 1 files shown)


entity-framework/index.yml
```diff
@@ -159,11 +159,15 @@ additionalContent:
...
             text: "SQL Server"
           - url: core/providers/sqlite/index.md
-            text: "SQL Lite"
+            text: "SQLite"
+          - url: http://www.npgsql.org/efcore/index.html
+            text: "PostgreSQL"
+          - url: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql
+            text: "MySQL and MariaDB"
           - url: core/providers/cosmos/index.md
             text: "Cosmos"
```
smitpatel, line 168: Should we name this Cosmos DB SQL API

JeremyLikness: Yes, I believe so ... I made the change

PR #2299, opened 2020-04-16
Missing update-database

oskarj:
There should be an update-database after applying a blank merge before "continue developing"


ajcvickers:
@oskarj What errors do you see if you don't do this step?

oskarj:
@ajcvickers the context is out of sync, if you have a Database initializer that checks if the model has changed when the application is starting you will get an exception with the message "the model backing context has changed" so you need to run update-database in order to get that migration into __MigrationHistory table.

ajcvickers:
Thanks @oskarj!

PR #2324, opened 2020-04-25
(AzureCXP) Correct description

BhargaviAnnadevara-MSFT:
The doc probably means ".. EF6 to EF Core" instead of ".. EF6 to .NET Core"

### DIFF
--- a/entity-framework/efcore-and-ef6/index.md
+++ b/entity-framework/efcore-and-ef6/index.md
@@ -148,7 +148,7 @@ Use EF Core on .NET Core for all new applications unless the app needs something
 
 EF Core is not a drop-in replacement for EF6. Moving from EF6 to EF Core will likely require changes to your application.
 
-When moving an EF6 app to .NET Core:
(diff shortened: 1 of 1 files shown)


roji:
Thanks!

smitpatel:
This is incorrect change. Need to revert!

roji:
Sorry, read through too quickly - reverted. The last bit of the page is somewhat confusing between EF Core and .NET Core.

ajcvickers:
@roji Can elaborate on what's confusing? I read it again and it seems clear to me! :-)

roji:
Just that the whole document deals with EF Core vs. EF6, and even the title of that last section doesn't actually say it's discussing guidance for potentially moving to .NET Core (just "Guidance for existing EF6 applications"). It may be worth putting that content in a top-level section explicitly entitled ".NET Framework and .NET Core" or similar.

It's not big deal - but both I and @BhargaviAnnadevara-MSFT misunderstood on a quick reading.

PR #2327, opened 2020-04-27
Fixes to samples to make them build, plus Github Action to prevent breaks

roji:
* Fix solution project reference
* Target netcoreapp3.1 instead of netcoreapp3.0
* Depend on EF Core 5.0.0-preview3 for new annotations
* Fix typo in type name

### DIFF
--- a/.github/workflows/build-samples.yml
+++ b/.github/workflows/build-samples.yml
@@ -1,0 +1,18 @@
+name: Build Samples
+
+on: push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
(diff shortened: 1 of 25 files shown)


ajcvickers:
@roji I was planning a similar PR to this. Question: should we use the previews everywhere? Or only where the sample has 5.0 code?

smitpatel:
I think using preview for preview feature and latest stable for all others would be good compromise. It would also make sure we update our samples from 3.1 to 5.0 when we release 5.0, can catch some breaking changes in existing functionality if any. Also allows people downloading sample to not have preview setup unless they are trying out preview features.

roji:
Changed to only depend on 5.0 preview3 in the DataAnnotations project which requires it.

PR #2328, opened 2020-04-27
Documentation page on collations and case-sensitivity

### LINKED ISSUE #2273: Document collation support and behavior around case-sensitivity
Document specifying collations on the database (https://github.com/dotnet/efcore/issues/6577) and on columns (https://github.com/dotnet/efcore/issues/19275), as well as the new EF.Functions.Collate (https://github.com/dotnet/efcore/issues/8813).

I think it makes sense to have a dedicated topic page for this, e.g. under fundamentals (which we should think about renaming/splitting up...). We could still of course link from model building.

roji:
Closes #2273


entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
```diff
@@ -0,0 +1,69 @@
+---
+title: Collations and case sensitivity - EF Core
+author: roji
```
smitpatel, line 3: description

entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
```diff
@@ -0,0 +1,69 @@
...
+ms.date: 04/27/2020
+ms.assetid: bde4e0ee-fba3-4813-a849-27049323d301
+uid: core/miscellaneous/collations-and-case-sensitivity.md
+---
+# Collations and Case Sensitivity
+
+> [!NOTE]
+> The APIs shown on this page are being introduced in EF Core 5.0, which is still in preview.
```
smitpatel, line 11: Just say it is introduced in 5.0

entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Database collation
+
+In most database systems, a default collation is defined at the database level; unless overridden, that collation implicitly applies to all text operations occurring within that database. The database collation is typically set at database creation time (via the `CREATE DATABASE` DDL statement), and if not specified, defaults to a some server-level value determined at setup time. For example, the default server-level collation in SQL Server is `SQL_Latin1_General_CP1_CI_AS`, which is a case-insensitive, accent-sensitive collation. Although database systems usually do permit altering the collation of an existing database, doing so can lead to complications; it is recommended to pick a collation before database creation.
+
+The following code in your model's `OnModelCreating` method configures a SQL Server database to use a case-sensitive collation:
```
ajcvickers, line 25: Make it clear that this only works if you're using EF (i.e. Migrations) to create the database.

entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
```diff
@@ -0,0 +1,69 @@
...
+
+[!code-csharp[Main](../../../samples/core/Miscellaneous/Collations/Program.cs?range=40)]
+
+## Column collation
+
+Collations can also be defined on text columns, overriding the database default. This can be useful if certain columns need to be case-insensitive, while the rest of the database needs to be case-sensitive.
+
+The following configures the column for the `Name` property to be case-insensitive in a database that is otherwise configured to be case-sensitive:
+
+[!code-csharp[Main](../../../samples/core/Miscellaneous/Collations/Program.cs?name=OnModelCreating&highlight=6)]
```
ajcvickers, line 35: Likewise.

entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
```diff
@@ -0,0 +1,69 @@
...
+
+> [!WARNING]
+> Always inspect the query plans of your queries, and make sure the proper indexes are being used in performance-critical queries executing over large amounts of data. Overriding case-sensitivity in a query via `EF.Functions.Collate` (or by calling `string.ToLower`) can have a very significant impact on your application's performance.
+
+## Translation of built-in .NET string operations
+
+In .NET, string equality is case-sensitive by default: `s1 == s2` performs an ordinal comparison that requires the strings to be identical. Because the default collation of databases varies, and because it is desirable for simple equality to use indexes, EF Core makes no attempt to translate simple equality to a database case-sensitive operation: C# equality is translated directly to SQL equality, which may or may not be case-sensitive, depending on the specific database in use and its collation configuration.
```
ajcvickers, line 62: Are you sure this is ordinal by default? (string.Equals is culture-sensitive by default, right?)

samples/core/Miscellaneous/Collations/Collations.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RootNamespace>EFCollations</RootNamespace>
+    <AssemblyName>EFCollations</AssemblyName>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="5.0.0-preview.3.20181.2" />
```
ajcvickers, line 10: You're too eager with your docs! ;-) These things aren't in preview 3...

entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
```diff
@@ -0,0 +1,69 @@
...
+
+In addition, .NET provides overloads of [`string.Equals`](https://docs.microsoft.com/dotnet/api/system.string.equals#System_String_Equals_System_String_System_StringComparison_) accepting a [`StringComparison`](https://docs.microsoft.com/dotnet/api/system.stringcomparison) enum, which allows specifying case-sensitivity and culture for the comparison. By design, EF Core refrains from translating these overloads to SQL, and attempting to use them will result in an exception. For one thing, EF Core does know not which case-sensitive or case-insensitive collation should be used. More importantly, applying a collation would in most cases prevent index usage, significantly impacting performance for a very basic and commonly-used .NET construct. To force a query to use case-sensitive or case-insensitive comparison, specify a collation explicitly via `EF.Functions.Collate` as [detailed above](#explicit-collations-and-indexes).
+
+## Database-specific information
+
+* [SQL Server documentation on collations](https://docs.microsoft.com/sql/relational-databases/collations/collation-and-unicode-support)
+* *MORE NEEDED*
```
ajcvickers, line 69: MORE NEEDED

entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
roji: According to the docs, both [string.Equals(string)](https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=netcore-3.1#System_String_Equals_System_String_), and [the string equality operator](https://docs.microsoft.com/en-us/dotnet/api/system.string.op_equality?view=netcore-3.1) are ordinal, i.e. non-culture-sensitive. In fact, re the the equality operator:

> The operator, in turn, calls the static Equals(String, String) method, which performs an ordinal (case-sensitive and culture-insensitive) comparison.

PS This is why I hate seeing `x.Equals(y, StringComparison.Ordinals)` in the codebase, since it can be simplified to `x == y` :)

samples/core/Miscellaneous/Collations/Collations.csproj
roji: Yep, that's why this PR fails the build :) Just wanted to get this done with while everything is still fresh in mind.

We can hold off merging until preview4 is out...

entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
roji: MORE PROVIDED

entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Database-specific information
+
+* [SQL Server documentation on collations](https://docs.microsoft.com/sql/relational-databases/collations/collation-and-unicode-support).
+* [Microsoft.Data.Sqlite documentation on collations](https://docs.microsoft.com/dotnet/standard/data/sqlite/collation).
+* [PostgreSQL documentation on collations](https://www.postgresql.org/docs/current/collation.html).
```
roji, line 71: This will be replaced with an EFCore.PG provider doc page - the provider also supports index collations, and a database-level collation that propagates in EF Core, like identity (since the PostgreSQL support in CREATE DATABASE is unfortunately pretty bad).

entity-framework/core/miscellaneous/collations-and-case-sensitivity.md
```diff
@@ -0,0 +1,72 @@
...
+
+```sql
+SELECT [c].[Id], [c].[Name]
+FROM [Customers] AS [c]
+WHERE [c].[Name] COLLATE SQL_Latin1_General_CP1_CS_AS = N'John'
+```
+
+### Explicit collations and indexes
+
+Indexes are one of the most important factors in database performance - a query that runs efficiently with an index can grind to a halt without that index. Indexes implicitly inherit the collation of their column; this means that all queries on the column are automatically eligible to use indexes defined on that column - provided that the query doesn't specify a different collation. Specifying an explicit collation in a query will generally prevent that query from using an index defined that column, since the collations would no longer match; it is therefore recommended to exercise caution when using this feature. It is always preferable to define the collation at the column (or database) level, allowing all queries to implicitly use that collation and benefit from any index.
```
lajones, line 54: Typo: "an index defined that column" -> "an index defined on that column"

entity-framework/core/modeling/entity-properties.md
```diff
@@ -124,3 +124,16 @@ A property that would be optional by convention can be configured to be required
 [!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/Required.cs?name=Required&highlight=3-5)]
 
 ***
+
+## Column collations
+
+> [!NOTE]
+> This feature is being introduced in EF Core 5.0, which is still in preview.
```
lajones, line 131: You could put this, but then you'll need to create an issue to go back and remove the "is still in preview" part once it no longer is. May be better to just skip the ", which is still in preview." part altogether.

roji: I've replaced this with "This feature is introduced in EF Core 5.0." which is what we use elsewhere. I find it grammatically a bit weird, but...

lajones: Actually it still says "which is still in preview" - which will become out of date at some point? If you're happy to update this later then feel free to ignore this.

roji:
@ajcvickers the EF.Functions.Collate support isn't in preview4 - I guess we should wait until preview5? Or do you think we should push a stripped-down version with it now?

ajcvickers:
@roji Either way is fine with me.

roji:
Then I say let's wait... Better to have the whole feature out. If some asks, we can always point them to this PR...

PR #2334, opened 2020-04-28
Args param is now used for IDesignTimeDbContextFactory

lajones:
Specify that the args param is now used for IDesignTimeDbContextFactory implementations.

### DIFF
--- a/entity-framework/core/miscellaneous/cli/dbcontext-creation.md
+++ b/entity-framework/core/miscellaneous/cli/dbcontext-creation.md
@@ -35,8 +35,9 @@ You can also tell the tools how to create your DbContext by implementing the `ID
 [!code-csharp[Main](../../../../samples/core/Miscellaneous/CommandLine/BloggingContextFactory.cs)]
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


entity-framework/core/miscellaneous/cli/dbcontext-creation.md
```diff
@@ -35,8 +35,8 @@ You can also tell the tools how to create your DbContext by implementing the `ID
 [!code-csharp[Main](../../../../samples/core/Miscellaneous/CommandLine/BloggingContextFactory.cs)]
 
 > [!NOTE]
-> The `args` parameter is currently unused. There is [an issue][8] tracking the ability to specify design-time arguments
-> from the tools.
+> Prior to 5.0.0 the `args` parameter was unused (see [this issue][8]). This was fixed in the 5.0.0 release and now any
```
ajcvickers, line 38: Is "This was fixed in the 5.0.0 release" the wording we are using for these things? I would prefer something like, "This is fixed in EF Core 5.0." since it doesn't imply a release has happened and includes explicit what release we mean--i.e. EF Core 5.0, not .NET 5.

PR #2349, opened 2020-05-01
Fix typos

mderriey:
- Trade-off is always used as a noun, so I think it's a single word.
- Its is used in the possessive form, so it shouldn't have an apostrophe.
- Add full stops to one of the unordered list items for consistency with the rest of the article.


ajcvickers:
Thanks @mderriey!

Issue #1900, opened 2019-11-05
Using links in docs


smitpatel:
Add this to guidelines somewhere

When referencing
- Topic on same page => Use id => #previous-versions
- Topic in EF docs => use unique doc id with xref => xref:core/querying/raw-sql (you can also append above here)
- Topic in non-EF docs => use path starting with / ignoring docs.microsoft.com part => /dotnet/csharp/programming-guide/concepts/linq/basic-linq-query-operations (this makes sure that language is preserved navigating through docs)
- Code from samples => use relative path => ../../../samples/core/Querying/Basics/Sample.cs#Filtering
- Also link to sample, we need to use absolute link since it is different site. Use master branch for all samples link. (e.g. https://docs.microsoft.com/en-us/ef/core/modeling/keyless-entity-types The link used in the tip.) 

Unless referencing something very external, we should not be using absolute links. (else who moved my cheese?)

smitpatel:
@ajcvickers - Can you update the links being changed in #1891 to above? There are build warnings in PR now.

AndriySvyryd:
We already have guidelines: https://github.com/dotnet/docs/blob/master/styleguide/template.md#links

PR #2351, opened 2020-05-01
Making links more relative.

lajones:
See #1900. Removing unneeded parts of links.


lajones:
@smitpatel - any comments before I check in?

PR #2379, opened 2020-05-11
Adding preview 4 features

ajcvickers:
We are light on features for preview 4.

I have not included any of the collation support since the last part of that is in preview 5. (Yes, we should have pulled it in to 4.)

Also, I would like to include examples of better translation exception messages, but I'm having trouble finding them.

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-5.0/whatsnew.md
+++ b/entity-framework/core/what-is-new/ef-core-5.0/whatsnew.md
@@ -2,7 +2,7 @@
 title: What's New in EF Core 5.0
 description: Overview of new features in EF Core 5.0
 author: ajcvickers
-ms.date: 03/30/2020
+ms.date: 05/11/2020
 uid: core/what-is-new/ef-core-5.0/whatsnew.md
 ---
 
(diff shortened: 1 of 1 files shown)


entity-framework/core/what-is-new/ef-core-5.0/whatsnew.md
```diff
@@ -16,6 +16,55 @@ The plan describes the overall themes for EF Core 5.0, including everything we a
...
+    .HasPrecision(16, 4);
+```
+
+Precision and scale can still be set via the full database type, such as "decimal(16,4)". 
+
+Documentation is tracked by issue [#527](https://github.com/dotnet/EntityFramework.Docs/issues/527).
+
+### Specify SQL Server index fill factor
+
+The fill factor can no be specified when creating an index on SQL Server.
```
JeremyLikness, line 39: "can now"

entity-framework/core/what-is-new/ef-core-5.0/whatsnew.md
```diff
@@ -16,6 +16,55 @@ The plan describes the overall themes for EF Core 5.0, including everything we a
...
+EF.CompileQuery(
+    (NorthwindContext context, string s1, string s2, string s3, string s4, string s5, string s6, string s7, string s8,
+        string s9, string s10, string s11, string s12, string s13, string s14, string s15)
+    => context.Set<Customer>()
+        .Count(c => c.CustomerID == s1 || c.CustomerID == s2 || c.CustomerID == s3 || c.CustomerID == s4 || c.CustomerID == s5
+            || c.CustomerID == s6 || c.CustomerID == s7 || c.CustomerID == s8 || c.CustomerID == s9 || c.CustomerID == s10
+            || c.CustomerID == s11 || c.CustomerID == s12 || c.CustomerID == s13 || c.CustomerID == s14 || c.CustomerID == s15));
+```
+
+Note that `EF.CompileQuery` is rarely needed since EF Core compiles queries automatically on first use.
```
JeremyLikness, line 66: Does it make sense to put a little blurb here explaining when it is needed, i.e. what is the rare case?

ajcvickers: Maybe. Or maybe this isn't even worth including in the new features? (I probably won't have included it if we had more other things...)

ajcvickers: Maybe the testing guidance sample should be called explicitly as something new?

roji: TBH I wouldn't include this.

PR #2385, opened 2020-05-15
Fix typos, add missing articles and commas

cucoreanu:


roji:
Thanks for these corrections @cucoreanu!

PR #2394, opened 2020-05-19
query filters docs - using navigations in query filters + sample

maumar:


entity-framework/core/querying/filters.md
```diff
@@ -39,6 +39,16 @@ The predicate expressions passed to the _HasQueryFilter_ calls will now automati
...
 
+## Use of navigations
+
+Navigations can be used when defining global query filters. They are applied recursively - when navigations used in query filters are translated, query filters defined on referenced entities are also applied, potentially adding more navigations.
+
+> [!NOTE]
+> Currently EF Core does not detect cycles in global query filter definitions, so you should be careful when defining them. If specified incorrectly, this could lead to infinite loops during query translation.
+
+> [!TIP]
+> It is advised to use optional navigations to reference entities that use global query filters. Required navigation expects related entity to exist. If required related entity is filtered out by the query filter, the parent entity could end up in unexpected state. This may lead to incorrect query results e.g. parent entity could also be removed from the result set.
```
maumar, line 50: This is not always a problem, when required navigation is defined on derived entity (TPH), we compensane during nav rewrite and produce LEFT JOINs rather than INNER JOINs. That's why i'm using may/could here. 
Also this sentence sounds bit complicated to me, any ideas how to make it simpler?

entity-framework/core/querying/filters.md
```diff
@@ -39,6 +39,16 @@ The predicate expressions passed to the _HasQueryFilter_ calls will now automati
 > [!NOTE]
 > It is currently not possible to define multiple query filters on the same entity - only the last one will be applied. However, you can define a single filter with multiple conditions using the logical _AND_ operator ([`&&` in C#](https://docs.microsoft.com/dotnet/csharp/language-reference/operators/boolean-logical-operators#conditional-logical-and-operator-)).
 
+## Use of navigations
+
+Navigations can be used when defining global query filters. They are applied recursively - when navigations used in query filters are translated, query filters defined on referenced entities are also applied, potentially adding more navigations.
```
maumar, line 44: Should I ditch the explanation what it means that we apply filters recursively? Sentence seems complicated to me, but it ties to the warning about cycles. Thoughts?

entity-framework/core/querying/filters.md
maumar: TODO: also, I need to add something about how to correctly setup required relationship in this case (both sides should define query filters to keep the graph consistent). 
Should I also add a sample for this? (prolly will have to upgrade from tip to actual doc section)

maumar:
updated @JeremyLikness can you create a forward link that points to: https://docs.microsoft.com/en-us/ef/core/querying/filters

entity-framework/core/querying/filters.md
```diff
@@ -39,6 +39,75 @@ The predicate expressions passed to the _HasQueryFilter_ calls will now automati
...
+
+> [!NOTE]
+> Currently EF Core does not detect cycles in global query filter definitions, so you should be careful when defining them. If specified incorrectly, this could lead to infinite loops during query translation.
+
+## Accessing entity with query filter using reqiured navigation
```
maumar, line 49: TMI?

JeremyLikness:
@maumar https://go.microsoft.com/fwlink/?linkid=2131316

maumar:
ping

entity-framework/core/querying/filters.md
ajcvickers: Seems okay to me.

entity-framework/core/querying/filters.md
ajcvickers: Looks good.

PR #2399, opened 2020-05-20
Reference 5.0.0-preview4 in samples

roji:


smitpatel:
It would be easier if we just update all this on RTM (unless we know that we changed the API which would impact the sample.)

ajcvickers:
@smitpatel We have docs waiting to merge that need preview 4.

smitpatel:
@ajcvickers - Each doc sample which requires at least certain preview version can and should update to that particular preview version. e.g. if many to many gets implemented in preview6 then PR which adds documentation about it and sample for it can reference preview6 version in the sample.

ajcvickers:
I would rather if the sample needs a preview that it be the latest preview where possible.

smitpatel:
To me that is unnecessary overhead for seemingly no gain.

roji:
Note I only updated samples which were already on preview2 - it's more to make sure people who get previews get the last version. I agree there's not a huge value there but it's also very easy to do (plus I volunteer to do it)...

PR #2406, opened 2020-05-27
Fix for issue 527 - Precision and Scale.

### LINKED ISSUE #527: Document HasPrecision

lajones:
Fixes #527. Documenting the `HasPrecision()` API.

### DIFF
--- a/entity-framework/core/modeling/entity-properties.md
+++ b/entity-framework/core/modeling/entity-properties.md
@@ -81,7 +81,7 @@ In the following example, configuring a maximum length of 500 will cause a colum
 
 ### Precision and Scale
 
-Fluent API to configure the precision and scale is available in EFCore 5.0. It tells the database provider how much storage is needed for a given column. It only applies to data types where  the provider allows the precision and scale to vary - usually just `decimal` and `DateTime`.
+Starting with EFCore 5.0, you can use fluent API to configure the precision and scale. It tells the database provider how much storage is needed for a given column. It only applies to data types where the provider allows the precision and scale to vary - usually just `decimal` and `DateTime`.
 
 For `decimal` properties, precision defines the maximum number of digits needed to express any value the column will contain, and scale defines the maximum number of decimal places needed. For `DateTime` properties, precision defines the maximum number of digits needed to express fractions of seconds, and scale is not used.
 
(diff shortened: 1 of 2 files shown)


entity-framework/core/modeling/entity-properties.md
```diff
@@ -79,6 +79,26 @@ In the following example, configuring a maximum length of 500 will cause a colum
 
 ***
 
+### Precision and Scale
+
+Fluent API to configure the precision and scale is available in EFCore 5.0. It tells the database provider how much storage is needed for a given column. It only applies to data types where  the provider allows the precision and scale to vary - usually just `decimal` and `DateTime`.
```
smitpatel, line 84: Use sentence which indicate that API is available in any version starting from 5.0

smitpatel:
```suggestion
Fluent API to configure the precision and scale is available in EFCore 5.0. It tells the database provider how much storage is needed for a given column. It only applies to data types where the provider allows the precision and scale to vary - usually just `decimal` and `DateTime`.
```

lajones: @smitpatel "is available in EFCore 5.0" is the form of words copied from other places in the docs, specifically [backing-field.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/backing-field.md) and [keyless-entity-types.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/keyless-entity-types.md).

If you don't like it please specify exactly what form of words should be used.

smitpatel: "is available since EFCore 5.0"

smitpatel: Should also update other pages.

lajones: Hmmm. "is available since ..." reads strangely to me. How about "became available in EFCore 5.0" which is what is in keyless-entity-types.md (and I'll update backing-field.md).

smitpatel: "in" is my issue. That causes ambiguity for what for version EF Core 6.0?
"Starting EF Core 5.0, you can use fluent API to configure the precision and scale."

lajones: Hmmm. I guess we just disagree here. I think the implication that it continues to be available unless we specifically state otherwise is clear.

Also you can't say "Starting EFCore 5.0" - but you could say "Starting with EFCore 5.0".

To compromise I'll update it to that, but I don't think that works for all docs. So I'll just update here. If you dislike the other docs then please raise a PR to come up with a generalized form of words that everyone likes and update everywhere.

PR #2410, opened 2020-05-30
Updates to providers page

roji:
* Update Oracle MySQL provider to 3.1 and bring up in the list
* Update Oracle provider to 3.1 (specify beta) and bring up in the list
* Update Devart providers from 3.0 to 3.1


roji:
@JeremyLikness you're referring to the doc links for our providers, right (e.g. SQL Server)? If so, I think we recently agreed to put/move some provider-specific info there, so these pages should hopefully become more useful. We should probably beef those pages up regardless - we should probably add basic connection string/getting started examples for Sqlite, and a small section on Sqlite limitations compared to other databases...

JeremyLikness:
@roji it was my mistake. In the preview view the first four rows show "docs" without an active link, but I see they are using a directive that will get updated when the page is built. So, to amend my review comments: approved, no changes necessary. Nothing to see here, please move along.

PR #2421, opened 2020-06-03
Add new features for preview 5

ajcvickers:

PR #2424, opened 2020-06-03
publish to live 6/4

smitpatel:
Will merge this after #2421 is merged to master so live site will have preview5 features.


roji:
@smitpatel should we merge #2328 for collations? I can temporarily remove the build check from there to make it green, and restore it afterwards, what do you think?

ajcvickers:
@smitpatel preview 5 features merged.

PR #2438, opened 2020-06-09
Refer to the correct property when discussing IModelCacheKeyFactory

serpent5:
The code in the sample shows `UseIntProperty` but the explanation refers to `IgnoreIntProperty`, so I've updated this accordingly.

I also removed an extraneous space and swapped a trailing `:` for a `.`. I wouldn't have submitted a PR for just those, but I took the opportunity whilst I was in the topic.


AndriySvyryd:
Thanks for your contribution, @serpent5

PR #2440, opened 2020-06-10
Fix to #2436 - Filtered include should also notify users about how it interact with tracked entities

### LINKED ISSUE #2436: Filtered include should also notify users about how it interact with tracked entities
We don't have any information for how filtered include and tracked entities work. e.g.
- Running include on same navigation with different filters in same context will actually union the related data in collection.
- Same for when there are tracked entities in the context and running filtered include query will stitch up tracked entity in collection too even if they don't match the filter.

Bottom line: Using filtered include in a tracking query is walking on a thin line.

---
#### Document Details

* ID: bbd36f35-e257-d0ed-8f28-f539ddcb71ec
* Version Independent ID: 093903d5-f7c7-5a78-c0e5-0e2662dcedef
* Content: [Loading Related Data - EF Core](https://docs.microsoft.com/en-us/ef/core/querying/related-data#filtered-include)
* Content Source: [entity-framework/core/querying/related-data.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/querying/related-data.md)
* GitHub Login: @rowanmiller

maumar:
adding "caution" block and example

Fixes #2436

### DIFF
--- a/entity-framework/core/querying/related-data.md
+++ b/entity-framework/core/querying/related-data.md
@@ -72,6 +72,18 @@ Alternatively, identical operations can be applied for each navigation that is i
 
 [!code-csharp[Main](../../../samples/core/Querying/RelatedData/Sample.cs#MultipleLeafIncludesFiltered2)]
 
+> [!CAUTION]
+> In case of tracking queries, results of Filtered Include may be unexpected due to [navigation fixup](tracking.md). All relevant entities that have been querried for previously and have been stored in the Change Tracker will be present in the results of Filtered Include query, even if they don't meet the requirements of the filter. Consider using `NoTracking` queries or re-create the DbContext when using Filtered Include in those situations.
+
+Example:
+
+```csharp
+var orders = context.Orders.Where(o => o.Id > 1000).ToList();
+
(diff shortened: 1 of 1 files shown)


entity-framework/core/querying/related-data.md
```diff
@@ -72,6 +72,18 @@ Alternatively, identical operations can be applied for each navigation that is i
 
 [!code-csharp[Main](../../../samples/core/Querying/RelatedData/Sample.cs#MultipleLeafIncludesFiltered2)]
 
+> [!CAUTION]
+> In case of tracking queries, results of Filtered Include may be unexpected due to [navigation fixup](tracking.md). All relevant entities that have been querried for previously and have been stored in the Change Tracker will be present in the results of Filtered Include query, even if they don't meet the requirements of the filter. Consider using `NoTracking` queries or re-create the DbContext when using Filtered Include in those situations.
+
+Example:
+
+```csharp
+var orders = context.Orders.Where(o => o.Id > 1000).ToList();
```
smitpatel, line 81: avoid inline code blocks

PR #2442, opened 2020-06-10
Update relationships.md

krptodr:
Update the variable name being passed to context.Entry(), to help the reader discern the difference between the ```Course``` type and the variable ```course```

### DIFF
--- a/entity-framework/ef6/fundamentals/relationships.md
+++ b/entity-framework/ef6/fundamentals/relationships.md
@@ -149,14 +149,14 @@ In Entity Framework you commonly use navigation properties to load entities that
 
 ``` csharp
     // Get the course where currently DepartmentID = 2.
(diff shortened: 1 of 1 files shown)


smitpatel:
I don't think there is any confusion between type and variable name.

krptodr:
I wonder if there was confusion, because the variable name was course2, however the usage of that variable name wasn't carried through to the last line of the code.

roji:
@krptodr the EF docs do assume that the reader is familiar enough with C# to understand the difference between types and variables plus the type is `Course` and the variable is `course`, as is standard. It seems better to follow the way code is actually written - I definitely wouldn't call a variable `course2` in my code.

ajcvickers:
@krptodr Thanks for this suggestion, but we're going to leave the doc as is based on the discussion with the team above.

PR #2448, opened 2020-06-10
Fix up wording when discussing MigrationBuilder.ActiveProvider

serpent5:
I noticed that "differently" should be "different", but I made a couple of other small change to the sentance. Please let me know if these kinds of changes are appreciated, and I will continue to submit PRs as I read through more of the topics.


ajcvickers:
Thanks!

PR #2453, opened 2020-06-11
Update text in links to match topic title

serpent5:
This makes the title used in the link text match the text used in the the ToC, for consistency.


ajcvickers:
Thanks!

PR #2466, opened 2020-06-16
Making sure we reference .NET Core CLI.

### LINKED ISSUE #2335: Consider spelling out PMC and CLI in the documentation
See [this comment](https://github.com/dotnet/EntityFramework.Docs/pull/2333#discussion_r416872086). @AndriySvyryd is suggesting spelling out PMC as "Package Manager Console" and CLI as ".NET Core CLI" throughout the docs so as to help with localization. However these abbreviations are widely used - and in some cases specifically called out as abbreviations - see e.g. https://docs.microsoft.com/en-us/ef/core/managing-schemas/scaffolding. So raising for discussion with the whole team.

lajones:
Fixes #2335 (remaining part). Make sure we reference ".NET Core CLI" where we haven't already defined "CLI" as an abbreviation for that.

### DIFF
--- a/entity-framework/core/miscellaneous/cli/powershell.md
+++ b/entity-framework/core/miscellaneous/cli/powershell.md
@@ -9,7 +9,7 @@ uid: core/miscellaneous/cli/powershell
 
 The Package Manager Console (PMC) tools for Entity Framework Core perform design-time development tasks. For example, they create [migrations](/aspnet/core/data/ef-mvc/migrations?view=aspnetcore-2.0), apply migrations, and generate code for a model based on an existing database. The commands run inside of Visual Studio using the [Package Manager Console](/nuget/tools/package-manager-console). These tools work with both .NET Framework and .NET Core projects.
 
-If you aren't using Visual Studio, we recommend the [EF Core Command-line Tools](dotnet.md) instead. The CLI tools are cross-platform and run inside a command prompt.
+If you aren't using Visual Studio, we recommend the [EF Core Command-line Tools](dotnet.md) instead. The .NET Core CLI tools are cross-platform and run inside a command prompt and work with .NET Core projects only.
 
 ## Installing the tools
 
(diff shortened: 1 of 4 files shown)


entity-framework/core/miscellaneous/cli/powershell.md
```diff
@@ -9,7 +9,7 @@ uid: core/miscellaneous/cli/powershell
 
 The Package Manager Console (PMC) tools for Entity Framework Core perform design-time development tasks. For example, they create [migrations](/aspnet/core/data/ef-mvc/migrations?view=aspnetcore-2.0), apply migrations, and generate code for a model based on an existing database. The commands run inside of Visual Studio using the [Package Manager Console](/nuget/tools/package-manager-console). These tools work with both .NET Framework and .NET Core projects.
 
-If you aren't using Visual Studio, we recommend the [EF Core Command-line Tools](dotnet.md) instead. The CLI tools are cross-platform and run inside a command prompt.
+If you aren't using Visual Studio, we recommend the [EF Core Command-line Tools](dotnet.md) instead. The .NET Core CLI tools are cross-platform and run inside a command prompt and work with .NET Core projects only.
```
lajones, line 12: @bricelam I'd be grateful if you would particularly check the last phrase I added. It's meant to be "as opposed to what we said line 10" where we work with both .NET Framework and .NET Core projects but I'm happy to change it if I've misunderstood what you were trying to stress here.

bricelam: Hmm, I think it just adds confusion. Also, `dotnet ef` works on .NET Framework projects (SDK-style only).

We added the one on line 10 because in the beginning, a lot of users didn't know they also worked with .NET Core projects. There was a bit of misinformation somewhere that you *had* to use `dotnet ef` on .NET Core projects.

lajones: Fair enough. I'll take out that phrase and check in. At that point our docs will be consistent and say ".NET Core CLI" everywhere. If there's a bigger change to change ".NET Core CLI" to ".NET CLI" then that will affect a lot more docs - so let's do that separately.

PR #2484, opened 2020-06-27
doc: add limitations

shahabganji:
Describe the limitation of Shadow Properties in the very first paragraph

### DIFF
--- a/entity-framework/core/modeling/shadow-properties.md
+++ b/entity-framework/core/modeling/shadow-properties.md
@@ -7,7 +7,7 @@ uid: core/modeling/shadow-properties
 ---
 # Shadow Properties
 
-Shadow properties are properties that are not defined in your .NET entity class but are defined for that entity type in the EF Core model. The value and state of these properties is maintained purely in the Change Tracker. Shadow properties are useful when there is data in the database that should not be exposed on the mapped entity types.
+Shadow properties are properties that are not defined in your .NET entity class but are defined for that entity type in the EF Core model. The value and state of these properties is maintained purely in the Change Tracker, this means if the entities loaded outside of the change tracker or with `AsNoTracking`, these properties will not have the corresponding values. Shadow properties are useful when there is data in the database that should not be exposed on the mapped entity types.
 
 ## Foreign key shadow properties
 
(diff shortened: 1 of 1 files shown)


ajcvickers:
@shahabganji This change seems somewhat confusing to me since it implies that the CLR type has properties but they are not being set. Can you explain a bit more what limitation you are trying to convey?

shahabganji:
@ajcvickers , I wanted to convey the difference between the following two queries: 

1. **With** `AsNoTracking()`

```cs 
 var authors = context.Authors.Select(x => x).AsNoTracking();

foreach (var author in authors)
{
    Console.WriteLine(
        $"The entity for AuthorId: {author.Id} " +
        $"was created at: {context.Entry(author).Property("CreatedOn").CurrentValue}");
}
```

2. **Without** `AsNoTracking()`
```cs
var authors2 = context.Authors.Select(x => x);
foreach (var author2 in authors2)
{
    Console.WriteLine(
        $"The entity for AuthorId: {author2.Id} " +
        $"was created at: {context.Entry(author2).Property("CreatedOn").CurrentValue}");
}
```

And the output will respectively be :


```bash
The entity for AuthorId: a2ab54b2-c72c-4644-8bf4-4fa48d3ee89d was created at: 01/01/0001 00:00:00 +00:00
```

```bash
The entity for AuthorId: a2ab54b2-c72c-4644-8bf4-4fa48d3ee89d was created at: 06/27/2020 16:31:47 +00:00
```
 And for clarity, the `CreatedOn` is a Shadow Property and the `Author` class is like this: 

```cs
public class Author
{
     public Guid Id { get; private set; }
     public string FirstName { get; private set; }
     public string LastName { get; private set; }
     public DateTime DateOfBirth { get; private set; }
}
```

shahabganji:
I could also add a section with an example if you think that will provide a better understanding of the problem

ajcvickers:
@shahabganji May add a sentence to `Accessing shadow properties`. Something like, "Shadow properties cannot be accessed after a no-tracking query since the entities returned are not tracked by the change tracker."

shahabganji:
@ajcvickers, The changes applied.

ajcvickers:
Thanks!

PR #2506, opened 2020-07-10
Update SQLite Limitations for 5.0

bricelam:
The future is now...


roji:
It's almost as if it were a real relational database...

PR #2539, opened 2020-07-28
better way to ref docs.ms.com

Rick-Anderson:
This is a better way to reference https://docs.microsoft.com/ documents. Even better is to use @guardrex  awesome tool https://xrefgen.azurewebsites.net/?handler=GetSearchResults and click the copy button.

cc @tdykstra @scottaddie  to see if this PR is worth doing. I think so because lots of folks look at current doc on how to do updates, new docs, etc.


tdykstra:
Yes, it's worth doing, but for xref links the new VS Code Docs Authoring Pack xref link generation function is much better than the xrefgen site.

PR #2554, opened 2020-08-03
Update index.yml

JeremyLikness:
Fix link to WinForms and refresh link to WPF.


JeremyLikness:
Fixes #2555

JeremyLikness:
@smitpatel and @ajcvickers the link on the main page for WinForms is broken (goes to WPF) so if we can expedite this to `live` that would be great.

PR #2573, opened 2020-08-18
Update index.yml and toc.yml to reference new Blazor guidance

JeremyLikness:
Add link to new Blazor Server guidance.

PR #2586, opened 2020-08-24
Update index.yml

JeremyLikness:
Applying suggestions from #2573.


entity-framework/toc.yml
```diff
@@ -66,9 +66,9 @@
         href: core/get-started/install/index.md
       - name: ASP.NET Core tutorial >>
         href: /aspnet/core/data/ef-rp/intro
-      - name: Blazor Server with EF Core reference app
+      - name: Blazor Server with EF Core guidance
```
smitpatel, line 69:
```suggestion
      - name: Blazor Server with EF Core guidance >>
```

PR #2599, opened 2020-08-29
Ordering Caveat in FromSql

nbuuck:
I recently experienced a peculiarity in `FromSql()` in a netcore-2.1 project as described in the commit. I think this may be worth documenting, but I understand if this should be opened as an issue in the EF Core project instead.

### DIFF
--- a/entity-framework/core/querying/raw-sql.md
+++ b/entity-framework/core/querying/raw-sql.md
@@ -50,6 +50,10 @@ You can also construct a DbParameter and supply it as a parameter value. Since a
 
 [!code-csharp[Main](../../../samples/core/Querying/RawSQL/Sample.cs#FromSqlRawStoredProcedureNamedSqlParameter)]
 
+> [!NOTE]
+> **Parameter Ordering**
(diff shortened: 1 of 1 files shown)


entity-framework/core/querying/raw-sql.md
```diff
@@ -50,6 +50,10 @@ You can also construct a DbParameter and supply it as a parameter value. Since a
 
 [!code-csharp[Main](../../../samples/core/Querying/RawSQL/Sample.cs#FromSqlRawStoredProcedureNamedSqlParameter)]
 
+> [!NOTE]
+> **Parameter Ordering**
+> When passing multiple `SqlParameter`s, the ordering in the SQL string must match the order of the parameters in the stored procedure's definition. Failing to do this can result in type conversion exceptions or unexpected behavior when the procedure is executed because Entity Framework Core will apply the order of the `SqlParameter[]` array literally and will not reorder them to match the order defined in the stored procedure's definition. This behavior is not typical of stored procedures when using named parameters and is seemingly specific to `FromSql()`.
```
JeremyLikness, line 55:
```suggestion
> Entity Framework Core passes parameters based on the order of the `SqlParameter[]` array. When passing multiple `SqlParameter`s, the ordering in the SQL string must match the order of the parameters in the stored procedure's definition. Failure to do this may result in type conversion exceptions and/or unexpected behavior when the procedure is executed.
```

JeremyLikness: Suggesting a more direct, concise note.

JeremyLikness (changes requested):

JeremyLikness (approved):

Issue #2479, opened 2020-06-25
Query: Relational: Explain single vs split query


smitpatel:
Explain the differences and characteristics of each scenario.
This would be landing page for fwlink from the warning.

roji:
@smitpatel the docs are live, I think we need to modify the fwlink?

smitpatel:
Updated fwlink.

PR #2606, opened 2020-09-03
Update single/split query docs

roji:
* Document the new global split query configuration and AsSingleQuery
* Refactoring of the section for better readability and some updates

Continues #2479

### DIFF
--- a/entity-framework/core/querying/related-data.md
+++ b/entity-framework/core/querying/related-data.md
@@ -15,7 +15,7 @@ Entity Framework Core allows you to use the navigation properties in your model 
 * **Explicit loading** means that the related data is explicitly loaded from the database at a later time.
 * **Lazy loading** means that the related data is transparently loaded from the database when the navigation property is accessed.
 
-> [!TIP]  
+> [!TIP]
 > You can view this article's [sample](https://github.com/dotnet/EntityFramework.Docs/tree/master/samples/core/Querying) on GitHub.
 
 ## Eager loading
@@ -24,7 +24,7 @@ You can use the `Include` method to specify related data to be included in query
 
 [!code-csharp[Main](../../../samples/core/Querying/RelatedData/Sample.cs#SingleInclude)]
 
-> [!TIP]  
+> [!TIP]
 > Entity Framework Core will automatically fix-up navigation properties to any other entities that were previously loaded into the context instance. So even if you don't explicitly include the data for a navigation property, the property may still be populated if some or all of the related entities were previously loaded.
 
 You can include related data from multiple relationships in a single query.
@@ -51,8 +51,7 @@ You may want to include multiple related entities for one of the entities that i
 
 ### Single and split queries
 
-> [!NOTE]
-> This feature is introduced in EF Core 5.0.
+#### Single queries
 
 In relational databases, all related entities are by default loaded by introducing JOINs:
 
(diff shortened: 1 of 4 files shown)


entity-framework/core/querying/related-data.md
```diff
@@ -82,21 +86,29 @@ INNER JOIN [Post] AS [p] ON [b].[BlogId] = [p].[BlogId]
 ORDER BY [b].[BlogId]
 ```
 
-While this avoids the performance issues associated with JOINs and cartesian explosion, it also has some drawbacks:
+> [!NOTE]
+> One-to-one related entities are always loaded via JOINs, as this has no performance impact.
```
smitpatel, line 90:
```suggestion
> [!NOTE]
> One-to-one related entities are always loaded via JOINs in the same query, as this has no performance impact.
```

entity-framework/core/querying/related-data.md
```diff
@@ -82,21 +86,29 @@ INNER JOIN [Post] AS [p] ON [b].[BlogId] = [p].[BlogId]
...
+
+#### Enabling split queries globally
```
smitpatel, line 92: Also document the warning user get if no default is set and query uses multiple includes.

entity-framework/core/querying/related-data.md
```diff
@@ -97,7 +97,9 @@ You can also configure split queries as the default for your application's conte
 
 When split queries are configured as the default, it is still possible to configure specific queries to execute as single queries:
 
-[!code-csharp[Main](../../../samples/core/Querying/RelatedData/Sample.cs?name=AsSplitQuery&highlight=5)] 
+[!code-csharp[Main](../../../samples/core/Querying/RelatedData/Sample.cs?name=AsSplitQuery&highlight=5)]
+
+By default, EF Core emits a warning if it detects single queries loading collection includes that may cause a performance issue. If you're confident that there are no related performance issue, you can explicitly configure the splitting behavior to SplitQuery to prevent these warnings from being generated.
```
smitpatel, line 102: Warning is not working that way.
We issue a warning if behavior is not specified globally or in the query using operator.
No warning if user explicitly asks for AsSingleQuery.
And to suppress a warning, we need to show the warning configuration in OnConfiguring.

roji: Oops, wrote SplitQuery instead of SingleQuery. Is it OK after this fix? If not do you want to make a suggestion?

smitpatel: This still does not clarify what is the warning, why it is issued in first place and how to suppress the warning.

roji: I've reworded the note, hopefully you'll find it better. If not, please make another suggestion.

> And to suppress a warning, we need to show the warning configuration in OnConfiguring.

I don't think it's necessary to show multiple ways to make the warning go away (set SingleQuery globally/on the query, warning config in OnConfiguring), from the user's perspective it's the same. In addition, configuring warnings in OnConfiguring is a global matter that should be documented elsewhere, rather than repeating this information everywhere where a warning is discussed.

entity-framework/core/querying/related-data.md
```diff
@@ -63,7 +62,12 @@ LEFT JOIN [Post] AS [p] ON [b].[BlogId] = [p].[BlogId]
 ORDER BY [b].[BlogId], [p].[PostId]
 ```
 
-If a typical blog has multiple related posts, rows for these posts will duplicate the blog's information, leading to the so-called "cartesian explosion" problem. As more one-to-many relationships are loaded, the amount of duplicated data may grow and adversely affect the performance of your application.
+If a typical blog has multiple related posts, rows for these posts will duplicate the blog's information, leading to the so-called "cartesian explosion" problem. As more one-to-many relationships are loaded, the amount of duplicated data may grow and adversely affect the performance of your application. By default, EF Core emits a warning if it detects queries loading collection includes that may cause performance issues.
```
smitpatel, line 65:
```suggestion
If a typical blog has multiple related posts, rows for these posts will duplicate the blog's information, leading to the so-called "cartesian explosion" problem. As more one-to-many relationships are loaded, the amount of duplicated data may grow and adversely affect the performance of your application.
```

smitpatel (approved):
Too much to and fro, I will update it later.

entity-framework/core/querying/related-data.md
smitpatel: You can add a global docs somewhere about how to suppress warnings. The unrepeated part which is most crucial for user is what EventId to use to suppress it and that is going to be varying between different scenarios.
Also setting SingleQuery mode is not the (right or only) way to suppress the warning.

I will update this page myself later. You can merge this.

roji: > You can add a global docs somewhere about how to suppress warnings

Opened #2615 to track.

> The unrepeated part which is most crucial for user is what EventId to use to suppress it and that is going to be varying between different scenarios

Shouldn't the EventIds be discoverable from the warning itself, rather than having to make the docs more verbose/heavy with them? Typically a user will want to suppress a warning after they see it in the log, that should contain the EventId etc.

> Also setting SingleQuery mode is not the (right or only) way to suppress the warning.

Looking forward to seeing what you propose.

PR #2623, opened 2020-09-09
Avoid unnecessary views in URL to other product docs

smitpatel:
View is like product version.
By not specifying the view, it will go to the latest version automatically.
In some cases we need to target particular view there we need to add &preserve-view=true

New warnings everyday!

### DIFF
--- a/entity-framework/core/miscellaneous/testing/testing-sample.md
+++ b/entity-framework/core/miscellaneous/testing/testing-sample.md
@@ -11,25 +11,26 @@ no-loc: [Item, Tag, Items, Tags, items, tags]
 
 > [!TIP]
 > The code in this document can be found on GitHub as a [runnable sample](https://github.com/dotnet/EntityFramework.Docs/tree/master/samples/core/Miscellaneous/Testing/ItemsWebApi/).
-> Note that some of these tests **are expected to fail**. The reasons for this are explained below. 
+> Note that some of these tests **are expected to fail**. The reasons for this are explained below.
 
 This doc walks through a sample for testing code that uses EF Core.
 
 ## The application
 
 The [sample](https://github.com/dotnet/EntityFramework.Docs/tree/master/samples/core/Miscellaneous/Testing/ItemsWebApi/) contains two projects:
-- ItemsWebApi: A very simple [Web API backed by ASP.NET Core](/aspnet/core/tutorials/first-web-api?view=aspnetcore-3.1&tabs=visual-studio) with a single controller
+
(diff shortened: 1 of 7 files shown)


entity-framework/core/miscellaneous/testing/testing-sample.md
```diff
@@ -11,25 +11,26 @@ no-loc: [Item, Tag, Items, Tags, items, tags]
...
 ### The model and business rules
 
 The model backing this API has two entity types: Items and Tags.
 
-* Items have a case-sensitive name and a collection of Tags.
-* Each Tag has a label and a count representing the number of times it has been applied to the Item.
-* Each Item should only have one Tag with a given label.
-  * If an item is tagged with the same label more than once, then the count on the existing tag with that label is incremented instead of a new tag being created. 
-* Deleting an Item should delete all associated Tags.
+- Items have a case-sensitive name and a collection of Tags.
```
roji, line 29: Do we have an official standard for list items??

entity-framework/core/miscellaneous/testing/testing-sample.md
```diff
@@ -95,29 +97,29 @@ Most validation and error handling have been removed to reduce clutter.
...
+- The EF in-memory database provider
 
 This is achieved by putting all the tests in a base class, then inheriting from this to test with each provider.
 
 > [!TIP]
 > You will need to change the SQL Server connection string if you're not using LocalDB.
-
-> [!TIP]
-> See [Testing with SQLite](xref:core/miscellaneous/testing/sqlite) for guidance on using SQLite for in-memory testing. 
+> See [Testing with SQLite](xref:core/miscellaneous/testing/sqlite) for guidance on using SQLite for in-memory testing.
```
roji, line 110: Not sure if it's a good idea to combine the tips which are unrelated... If this is just about removing a warning, maybe better to suppress the warning here instead somehow?

roji (approved):

entity-framework/core/miscellaneous/testing/testing-sample.md
smitpatel: I think we should decide one. vscode ask it to be same in the file.

entity-framework/core/miscellaneous/testing/testing-sample.md
smitpatel: I don't think they are unrelated. Or else the sqlite testing should go to a different place.

entity-framework/core/miscellaneous/testing/testing-sample.md
smitpatel: Filed #2624

PR #2654, opened 2020-09-17
Update index.md

### LINKED ISSUE #1123: Add advantages and disadvantages of each loading method
It would be helpful to know the drawbacks of each loading method. For example, lazy loading requires less code (i.e. no `Include` calls needed) but results in sending more queries when compared to loading things with just 1 query with `Include` and `ThenInclude`. It would help to decide which method to use.

---
#### Document Details

* ID: bbd36f35-e257-d0ed-8f28-f539ddcb71ec
* Version Independent ID: 093903d5-f7c7-5a78-c0e5-0e2662dcedef
* Content: [Loading Related Data - EF Core](https://docs.microsoft.com/en-us/ef/core/querying/related-data)
* Content Source: [entity-framework/core/querying/related-data.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/querying/related-data.md)
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

Rick-Anderson:
Fixes #1123

Link points to https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/read-related-data#eager-explicit-and-lazy-loading

### DIFF
--- a/entity-framework/core/querying/related-data/index.md
+++ b/entity-framework/core/querying/related-data/index.md
@@ -14,5 +14,7 @@ Entity Framework Core allows you to use the navigation properties in your model 
 * **[Explicit loading](xref:core/querying/related-data/explicit)** means that the related data is explicitly loaded from the database at a later time.
 * **[Lazy loading](xref:core/querying/related-data/lazy)** means that the related data is transparently loaded from the database when the navigation property is accessed.
 
+For more information on each of these loading approaches, see [Eager, explicit, and lazy loading](/aspnet/core/data/ef-rp/read-related-data).
+
 > [!TIP]
 > You can view the [samples](https://github.com/dotnet/EntityFramework.Docs/tree/master/samples/core/Querying) under this section on GitHub.


Rick-Anderson:
We should expand on the doc. 
Eager loading can be more efficient if most of the related data is accessed.
Lazy loading is more efficient if few of the navigation properties are accessed.

roji:
Was not aware of the [ASP.NET-side page on related entity loading](https://docs.microsoft.com/en-ca/aspnet/core/data/ef-rp/read-related-data?view=aspnetcore-3.1)... I get that it's good to have a tutorial rather than just reference docs, but in this case the ASP.NET docs seem to really duplicate the EF Core docs and provide reference-style descriptions... At the very least I'd consider linking back to the EF docs?

However, I do agree that our [overview page for related entity loading](https://docs.microsoft.com/en-us/ef/core/querying/related-data/) is a bit light, and rather than having a single sentence on each technique, a small paragraph would be more helpful...

> Eager loading can be more efficient if most of the related data is accessed.
> Lazy loading is more efficient if few of the navigation properties are accessed.

At least from personal experience, it's been very rare to see a justified use of lazy loading to improve perf... EF Core 5.0 has [filtered includes](https://docs.microsoft.com/en-us/ef/core/querying/related-data/eager#filtered-include), which allow you to (eagerly) get a subset of related entities which you want - this addresses most of needs (but certainly not all).

Pragmatically-speaking, I think we've seen that lazy loading is bad for perf, as users very easily fall into N+1 patterns (loading parent, enumerating through children), so I'm not sure we'd want to recommend it...

Rick-Anderson:
@roji 
> At the very least I'd consider linking back to the EF docs?

The ASP.NET Core docs do link back to the EF docs.. Here's the markup sample
` [Eager loading](/ef/core/querying/related-data#eager-loading). ` Eager loading is when a query for one type of entity also loads related entities.

The original issue is customers want to know when to use each. Should we merge this PR but keep the issue open until you guys have time to write a paragraph for each type?

JeremyLikness (commented):
I think the benefits or trade-offs should be summarized here. 
i.e. just brain dump:
Eager loading - more overhead up front but easier coding approach
Explicit loading - useful for deep and/or cyclic graphs, requires explicit code
Lazy loading - can be best of both worlds but has side effects

ajcvickers:
Note from triage: We will link from the ASP.NET docs to more information in the EF Core docs, but not vice versa unless we are linking specifically to ASP.NET information. For cases like this that are about EF Core functionality it makes sense for the most detailed documentation to exist in the EF Core docs.

PR #2685, opened 2020-09-25
Added Hyperlink to SQLite

ashishagarwal06:
HI - Added hyperlink to SQLite in below line

EF Core works with SQL Server/SQL Azure, SQLite, Azure Cosmos DB, MySQL, PostgreSQL, and many more databases through a database provider plugin model.

Thanks,
Ashish


smitpatel:
Same resolution as #2688

PR #2688, opened 2020-09-25
Added Hyperlink to PostgreSQL

ashishagarwal06:
Hi  - Added hyperlink to PostgreSQL for redirection

EF Core works with SQL Server/SQL Azure, SQLite, Azure Cosmos DB, MySQL, [PostgreSQL](https://www.postgresql.org/), and many more databases through a [database provider plugin model](xref:core/providers/index).

Thanks,
Ashish


smitpatel:
After discussing with team, we decided that we do not want to add hyper link to particular database docs in any page. Rather we will only add those links for particular provider page for relevant database. See https://github.com/dotnet/EntityFramework.Docs/issues/2690
External provider documentation can link to their own database documentation as needed.

PR #2696, opened 2020-09-26
Added Hyperlink to EF Core

ashishagarwal06:
HI Team - I have added to EF Core for quick redirection.

Thanks,
Ashish


smitpatel:
We do not need to add link back to EF Core or EF 6 since user is already in EF docs.

PR #2698, opened 2020-09-26
Add savepoints docs and other improvements to "saving"

### LINKED ISSUE #2429: Document the new savepoint API
Following #20176

In [the transactions page](https://docs.microsoft.com/en-us/ef/core/saving/transactions).

roji:
* Add savepoint docs
* Make all "saving" samples async, apply some C# 8 goodness
* Make all samples which depend on 5.0.0 depend on rc1

Closes #2429


entity-framework/core/saving/transactions.md
```diff
@@ -14,19 +14,27 @@ Transactions allow several database operations to be processed in an atomic mann
...
+[!code-csharp[Main](../../../samples/core/Saving/Transactions/ControllingTransaction.cs?name=Transaction&highlight=2,16-18)]
 
-[!code-csharp[Main](../../../samples/core/Saving/Transactions/ControllingTransaction/Sample.cs?name=Transaction&highlight=3,17,18,19)]
+While all relational database providers support transactions, other providers types may throw or no-op when transaction APIs are called.
 
-## Cross-context transaction (relational databases only)
+## Savepoints
+
+When `SaveChanges` is invoked and a transaction is already in progress on the context, EF will automatically create a *savepoint* before saving any data. Savepoints are points within a database transaction which may be later be rolled back to, if an error occurs or for any other reason. If `SaveChanges` encounters any error, it will automatically roll the transaction back to the savepoint, leaving the transaction in the same state as if it had never started. This allows you to possibly correct issues and retry saving, in particular when [optimistic concurrency](xref:core/saving/concurrency) issues occur.
```
Rick-Anderson, line 31:
```suggestion
When `SaveChanges` is invoked and a transaction is already in progress on the context, EF automatically creates a *savepoint* before saving any data. Savepoints are points within a database transaction which may be later be rolled back. For example, savepoints can roll back if an error occurs or for other reasons. If `SaveChanges` encounters an error:
* The transaction is automatically rolled back to the savepoint.
* The transaction stays in the same state as if it had never started.
Savepoints:
*  Allow the possibly to correct issues and retry saving.
* Are useful when [optimistic concurrency](xref:core/saving/concurrency) issues occur.
```

Rick-Anderson: Avoid future tense and long sentences.

roji: Definitely accept the future tense comment, need to get used to that.

But I looked at the text again in preview, and the last two sentences don't seem overly long to me... IMHO it's clearer (and shorter) as-is than broken down into bullet points...

smitpatel:
> Make all "saving" samples async, apply some C# 8 goodness

Don't see a particular reason to do this. Also samples like Getting started or others which needs to save data (e.g. query sample with dummy data) may use different. It could be inconsistent and confusing. Do we have customers misunderstanding that all EF API are also available in async?
@dotnet/efteam - thoughts?

roji:
The idea is that current .NET best practices are to use async, and to treat sync as more or less deprecated. ASP.NET has even disabled sync I/O (on the web side) unless a very specific (and discouraged) opt-in is used. Our samples should show modern coding practices, rather than encouraging people to copy-paste sync I/O practices.

I agree that we'd ideally need to do this across the board. If that's the blocking issue for this PR, I volunteer to do this.

smitpatel:
> The idea is that current .NET best practices are to use async, and to treat sync as more or less deprecated.

If that is the "direction" we are taking then we should do it consistently for all samples. We have async operation for every single sync operation in EF Core.

roji:
As I wrote above, I agree and even volunteer to do that. But I'm not sure why that needs to block this PR.

entity-framework/core/saving/transactions.md
```diff
@@ -1,32 +1,40 @@
...
-Not all database providers support transactions. Some providers may throw or no-op when transaction APIs are called.
+[!code-csharp[Main](../../../samples/core/Saving/Transactions/ControllingTransaction.cs?name=Transaction&highlight=2,16-18)]
 
-[!code-csharp[Main](../../../samples/core/Saving/Transactions/ControllingTransaction/Sample.cs?name=Transaction&highlight=3,17,18,19)]
+While all relational database providers support transactions, other providers types may throw or no-op when transaction APIs are called.
 
-## Cross-context transaction (relational databases only)
+## Savepoints
+
+When `SaveChanges` is invoked and a transaction is already in progress on the context, EF automatically creates a *savepoint* before saving any data. Savepoints are points within a database transaction which may later be rolled back to, if an error occurs or for any other reason. If `SaveChanges` encounters any error, it automatically rolls the transaction back to the savepoint, leaving the transaction in the same state as if it had never started. This allows you to possibly correct issues and retry saving, in particular when [optimistic concurrency](xref:core/saving/concurrency) issues occur.
```
Rick-Anderson, line 31: > This allows you to possibly correct 

What's *this* ?

See [Writing for translation](https://review.docs.microsoft.com/en-us/help/contribute/localization-mt-guidance?branch=master)

Short sentences should be typically about 15-20 words.

AndriySvyryd (approved):

entity-framework/core/saving/transactions.md
roji: `this` here is an indefinite anaphoric pronoun, referring in general to what has been mentioned in the previous sentence.

Re 15-20 words, I don't see any sentence in this paragraph which significantly exceeds that. IMO writing guidelines are definitely not meant as hard quantitative rules.

smitpatel:
This PR had blocked label for a reason
image

roji:
Sorry, I didn't notice it - what was the reason, is it the async samples? If considered critical, we can revert.

AndriySvyryd:
We'll discuss it in Triage

phonglevishn:
Upgrade EF core version 3.1 to 5 with new feature SavePoints (auto created when use transaction manual)

I have a memory-optimized table in SQL Server 2016.

When I call the "SaveChanges" command the system throws an exception "Creating a savepoint is not supported with memory optimized tables." How can I turn off savepoints when use memory optimized tables in transaction sql ?

NOTE: If I use TransactionScope then success.

Please give me the solution to handle this situation

roji:
@phonglevishn thanks for flagging this, I've opened https://github.com/dotnet/efcore/issues/23526 to track and posted a workaround there to disable savepoints altogether.

PR #2702, opened 2020-09-26
Updating Xamarin sample with proper linker settings

codemillmatt:
Xamarin.iOS requires the linker behavior to be set to `Don't Link`. Updating the sample to have the linker set to that value for all build configurations and updating the readme to note that as well.

### DIFF
--- a/entity-framework/core/get-started/xamarin.md
+++ b/entity-framework/core/get-started/xamarin.md
@@ -68,6 +68,9 @@ The following sections will walk you through the code in the sample project that
 
 It is assumed that you are familiar with the Xamarin.Forms topics of [displaying data](/xamarin/xamarin-forms/app-fundamentals/data-binding/) and [navigating between pages](/xamarin/xamarin-forms/app-fundamentals/navigation/).
 
+> [!IMPORTANT]
+> The linker behavior on Xamarin.iOS must be set to **Don't Link** in order to run applications with Entity Framework Core. [This article explains more about the linker](/xamarin/ios/deploy-test/linker) including how to set the behavior on Xamarin.iOS.
+
 ## Entity Framework Core NuGet packages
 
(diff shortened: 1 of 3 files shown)


codemillmatt:
@JeremyLikness for review.

JeremyLikness:
@codemillmatt few questions for this PR:

1. Is `--interpreter` still needed or not? If yes, can we update the doc to explain why?

2. When I follow the link for the linker options, I see this:

> Disabling linking will make sure that no assemblies are modified. For performance reasons this is the default setting when your IDE targets for the iOS simulator. For devices builds this should **only be used as a workaround** whenever the linker contains a bug that prevents your application to run. If your application only works with -nolink, please submit a bug report.

(Emphasis added my own). Is it possible to turn off linking just for the Entity Framework package? Is this solution a workaround (we should note that in the doc) or a permanent one?

Jeremy

codemillmatt:
Hey @JeremyLikness ...

1. `--interpreter` is no longer needed.

2. And I wanted to talk with you about linking/not linking Entity Framework. Right now when I tell the linker to skip the EF packages I still get the exception. I would suspect then that EF is using something further down the chain the linker needs. Or maybe not ... I'm going to pull the EF source into the solution & see what I can find out. Ideally we can then add `[Preserve]` attributes to the correct spots and make a permanent fix.

codemillmatt:
fwiw - it looks like it's the `QueryableMethods` constructor where the issue is originating from - lots of reflection going on in there.

codemillmatt:
A little more digging - since `QueryableMethods` is doing a bunch of reflection - something that eventually gets used based off that class (in this particular case it will be `Count`) gets linked out.

But if I do a `--linkskip System.Core` all is well without having to turn off linking altogether.

Now with that said, it would be nice if we could narrow this down a bit more, so I'll still do some digging.

JeremyLikness:
@codemillmatt thanks for your efforts with this! Is it in a state ready for review or are you still working on it?

codemillmatt:
@JeremyLikness - yes - please do give it a review. If it's good let's merge this in.

PR #2705, opened 2020-09-28
Add database functions docs

smitpatel:


entity-framework/toc.yml
```diff
@@ -175,6 +175,10 @@
       items:
       - name: Overview
         href: core/querying/index.md
+      - name: Client vs. server evaluation
```
roji, line 178: I think this undoes a change we recently made... FWIW I think loading related data is more immediately important than both client/server evaluation (definitely since 3.0) and tracking/no-tracking.

entity-framework/toc.yml
```diff
@@ -187,14 +191,18 @@
...
             href: core/querying/related-data/serialization.md
-      - name: Client vs. server evaluation
-        href: core/querying/client-eval.md
-      - name: Tracking vs. no-tracking
-        href: core/querying/tracking.md
       - name: Complex query operators
         href: core/querying/complex-query-operators.md
       - name: Raw SQL queries
         href: core/querying/raw-sql.md
+      - name: Database functions
```
roji, line 198: Maybe a sub-node for functions?

roji (commented):
Not a full review because still a draft, but a couple comments.

entity-framework/toc.yml
smitpatel: I will think about it. I do not agree with that priority order.

AndriySvyryd: Either way this should be done in a separate PR to keep discussion on-topic

entity-framework/toc.yml
AndriySvyryd: Yes, I think these 3 topics can be collapsed unless you plan to write a lot in each of them

smitpatel: > Yes, I think these 3 topics can be collapsed unless you plan to write a lot in each of them

A lot is coming soon.

entity-framework/core/querying/built-in-functions.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Built-in Functions - EF Core
```
bricelam, line 2: FYI, I'm compiling a list of SQLite translations in the following format.

.NET | SQL | Added in
--- | --- | ---
dateTime.AddMilliseconds(value) | `datetime(@dateTime, (@value / 1000.0) \|\| ' seconds')` | EF Core 2.2

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Database Functions - EF Core
+description: Information on database functions supported in EF Core query translation
+author: smitpatel
+ms.date: 11/10/2020
+uid: core/querying/database-functions
+---
+# Database Functions
+
+Database functions are database equivalent of [C# methods](/dotnet/csharp/programming-guide/classes-and-structs/methods). A database function can be invoked with zero or more parameters and it computes the result based on the parameter values. Most databases, which use SQL for query have support for database functions. So SQL generated by EF Core query translation also allows invoking database functions. C# methods don't have to translate to database functions strictly in EF Core.
```
bricelam, line 10:
```suggestion
Database functions are the database equivalent of [C# methods](/dotnet/csharp/programming-guide/classes-and-structs/methods). A database function can be invoked with zero or more parameters and it computes the result based on the parameter values. Most databases, which use SQL for querying have support for database functions. So SQL generated by EF Core query translation also allows invoking database functions. C# methods don't have to translate strictly to database functions in EF Core.
```

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
...
+
+- A C# method may not have an equivalent database function.
+  - <xref:System.String.IsNullOrEmpty%2A?displayProperty=nameWithType> method translates to a null check and a comparison with an empty string in the database rather than a function.
+  - <xref:System.String.Equals(System.String,System.StringComparison)?displayProperty=nameWithType> method doesn't have database equivalent since string comparison can't be represented or mimicked easily in a database.
+- A database function may not have an equivalent C# method. `??` operator in C#, which doesn't have any method, translates to `COALESCE` function in database.
```
bricelam, line 15:
```suggestion
- A database function may not have an equivalent C# method. The `??` operator in C#, which doesn't have any method, translates to the `COALESCE` function in the database.
```

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Types of database functions
+
+EF Core SQL generation supports a subset of functions that can be used in database. This limitation comes from the ability to represent query in LINQ for the given database function. Further, each database has varying support of database functions, so EF Core provides a common subset. A database provider is free to extend EF Core SQL generation to support more patterns. Following are type of the database functions EF Core supports and uniquely identifies. These terms also help in understanding what translations come built in with EF Core provider.
```
bricelam, line 19:
```suggestion
EF Core SQL generation supports a subset of functions that can be used in databases. This limitation comes from the ability to represent a query in LINQ for the given database function. Further, each database has varying support of database functions, so EF Core provides a common subset. A database provider is free to extend EF Core SQL generation to support more patterns. Following are the types of database functions EF Core supports and uniquely identifies. These terms also help in understanding what translations come built in with EF Core providers.
```

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Built-in vs user-defined functions
+
+Built-in functions come with database predefined, but user-defined functions are explicitly defined by the user in the database. When EF Core translates query to use database functions, it uses built-in functions to make sure that the function is always available on database. The distinction of built-in function is necessary in some databases to generate SQL correctly. For example SqlServer requires that every user-defined function is invoked with schema qualified name. But built-in functions in SqlServer don't have any schema. PostgreSQL defines built-in function in `public` schema and it can be invoked with schema qualified name.
```
bricelam, line 23:
```suggestion
Built-in functions come with database predefined, but user-defined functions are explicitly defined by the user in the database. When EF Core translates queries to use database functions, it uses built-in functions to make sure that the function is always available on the database. The distinction of built-in functions is necessary in some databases to generate SQL correctly. For example SqlServer requires that every user-defined function is invoked with a schema-qualified name. But built-in functions in SqlServer don't have a schema. PostgreSQL defines built-in function in the `public` schema but they can be invoked with schema-qualified names.
```

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Aggregate vs scalar vs table-valued functions
+
+- Scalar functions take scalar values as parameter(s) and return a scalar value as the result. Scalar functions can be used anywhere in SQL where a scalar value can be passed.
```
bricelam, line 27: Maybe add an example of a *scalar value* for clarity.

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
...
+- Aggregate functions take a stream of scalar values as parameter and return a scalar value as the result. Aggregate functions are applied on whole query result set or on group of values generated by applying `GROUP BY` operator.
```
bricelam, line 28:
```suggestion
- Aggregate functions take a stream of scalar values as parameters and return a scalar value as the result. Aggregate functions are applied on the whole query result set or on a group of values generated by applying `GROUP BY` operator.
```

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
...
+- Table-valued functions take scalar values as parameter(s) and return a stream of rows as the result. Table-valued functions are used as a table source in `FROM` clause.
+
+### Niladic functions
+
+Niladic functions are special database functions that don't have any parameters and must be invoked without parenthesis. They're similar to property/field access on an instance in C#. Niladic functions differ from parameter-less functions as the latter require empty parenthesis. There's no special name for database functions that requires parenthesis always. Another subset of database functions based on parameter count is variadic functions. Variadic functions can take varying number of parameters when invoked.
+
+## Database function mappings in EF Core
+
+EF Core supports three different kinds of mapping between C# functions and database functions.
```
bricelam, line 37:
```suggestion
EF Core supports three different ways of mapping between C# functions and database functions.
```

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Built-in function mapping
+
+By default EF Core providers provide mappings for various built-in functions over primitive types. For example, <xref:System.String.ToLower?displayProperty=nameWithType> translates to `LOWER` in SqlServer. This functionality allows users to write query in LINQ seamlessly. We usually provide translation in the database, which gives same result as what C# function provides on client side. Sometimes, to achieve that, the actual translation could be something more complicated then a database function. In some scenarios, we also provide the most appropriate translation rather than matching C# semantics. The same feature is also used to provide common translations for some of the C# member accesses. For example, <xref:System.String.Length?displayProperty=nameWithType> translates to `LEN` in SqlServer. Apart from providers, plugin writers can also add additional translations. This extensibility is useful when plugins add support for more types as primitive types and want to translate methods over them.
```
bricelam, line 41:
```suggestion
By default EF Core providers provide mappings for various built-in functions over primitive types. For example, <xref:System.String.ToLower?displayProperty=nameWithType> translates to `LOWER` in SqlServer. This functionality allows users to write queries in LINQ seamlessly. We usually provide a translation in the database which gives the same result as what the C# function provides on the client side. Sometimes, to achieve that, the actual translation could be something more complicated than a database function. In some scenarios, we also provide the most appropriate translation rather than matching C# semantics. The same feature is also used to provide common translations for some of the C# member accesses. For example, <xref:System.String.Length?displayProperty=nameWithType> translates to `LEN` in SqlServer. Apart from providers, plugin writers can also add additional translations. This extensibility is useful when plugins add support for more types as primitive types and want to translate methods over them.
```

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
...
+
+### EF.Functions mapping
+
+Since not all database functions have C# equivalent functions, EF Core providers have special C# methods to invoke certain database functions. These methods are defined as extension methods over `EF.Functions` to be used in LINQ queries. These methods are provider-specific as they're closely tied with particular database functions. So a method that works for one provider will likely not work for any other provider. Further, since the intention of these methods to invoke database function in translated query, trying to evaluate them on client-side results in an exception.
```
bricelam, line 45:
```suggestion
Since not all database functions have equivalent C# functions, EF Core providers have special C# methods to invoke certain database functions. These methods are defined as extension methods over `EF.Functions` to be used in LINQ queries. These methods are provider-specific as they're closely tied with particular database functions. So a method that works for one provider will likely not work for any other provider. Further, since the intention of these methods is to invoke a database function in the translated query, trying to evaluate them on the client side results in an exception.
```

entity-framework/core/querying/database-functions.md
```diff
@@ -0,0 +1,55 @@
...
+
+### User-defined function mapping
+
+Apart from mappings provided by EF Core providers, users can also define custom mapping. User-defined mapping extends query translation according to user needs. This functionality is useful when there are user-defined functions in the database which user wants to invoke from their LINQ query.
```
bricelam, line 49:
```suggestion
Apart from mappings provided by EF Core providers, users can also define a custom mapping. A user-defined mapping extends the query translation according to the user needs. This functionality is useful when there are user-defined functions in the database which the user wants to invoke from their LINQ query.
```

bricelam (approved):
/me takes out a red pen and channels his 9th grade English teacher Mr. Benjamin.

entity-framework/core/querying/database-functions.md
smitpatel: Like add ?
 Scalar values are the values of primitive types supported by the database.
What kind of examples should I add? actual values - 1, "Seattle"
or server types like int, text, nvarchar(max)
or client types like int,string ?

bricelam: Those are good. Even something more subtle would be fine:

...take scalar values—like integers or strings—as parameters...

PR #2713, opened 2020-09-29
Add warning about lost custom operations

dchristensen:
Warning about losing custom operations when resetting/squashing migrations.


smitpatel:
Does it need to be added to the docs though? If you are deleting "Migrations" folder then you are going to lose all data inside it, custom or not-custom. Before deleting any source code file, you should evaluate if it contains anything you want to save.

dchristensen:
My thinking was that although you are deleting the "Migrations" folder you are expecting EF to re-generate that code (just squashed into a single file).    However, if you have added any customizations those will not be re-generated and will be lost.  The resulting "squashed" migration will not contain those original customizations and will therefore be different from the expected output.

smitpatel (approved):

smitpatel:
@dchristensen - Thank you.

Issue #504, opened 2017-10-06
New feature topic: built-in function mapping and EF.Functions


divega:
Originally from #447.

Should cover EF.Functions.Like as well.

PR #2733, opened 2020-10-05
Document SQLite function mappings

bricelam:
Part of #504

### DIFF
--- a/entity-framework/core/providers/sqlite/functions.md
+++ b/entity-framework/core/providers/sqlite/functions.md
@@ -1,0 +1,103 @@
+---
+title: Function Mappings - SQLite Database Provider - EF Core
+description: Function Mappings of the SQLite EF Core database provider
+author: bricelam
+ms.date: 10/06/2020
+uid: core/providers/sqlite/functions
+---
+# Function Mappings of the SQLite EF Core Provider
+
+This page shows which .NET members are translated into which SQL functions when using the SQLite provider.
+
+## Binary functions
+
+.NET                  | SQL                             | Added in
+--------------------- | ------------------------------- | --------
+bytes.Contains(value) | instr(@bytes, char(@value)) > 0 | EF Core 5.0
+bytes.Length          | length(@bytes)                  | EF Core 5.0
+
+## Date and time functions
+
+.NET                            | SQL                                                                      | Added in
+------------------------------- | ------------------------------------------------------------------------ | --------
+DateTime.Now                    | datetime('now', 'localtime')
+DateTime.Today                  | datetime('now', 'localtime', 'start of day')
+DateTime.UtcNow                 | datetime('now')
+dateTime.AddDays(value)         | datetime(@dateTime, @value \|\| ' days')
+dateTime.AddHours(value)        | datetime(@dateTime, @d \|\| ' hours')
+dateTime.AddMilliseconds(value) | datetime(@dateTime, (@value / 1000.0) \|\| ' seconds')                   | EF Core 2.2
+dateTime.AddMinutes(value)      | datetime(@dateTime, @value \|\| ' minutes')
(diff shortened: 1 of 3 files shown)


bricelam:
Hmm, it's probalby worth breaking the table up into sections. Something like...

- Binary functions
- Date and time functions
- Numeric functions
- String functions

smitpatel:
Generally should break on the type. That is generally followed in various database documentations.

entity-framework/core/providers/sqlite/built-in-functions.md
```diff
@@ -0,0 +1,76 @@
...
+This table shows which .NET members are translated into which SQL functions when using the SQLite provider.
+
+.NET                                      | SQL                                                                      | Added in
+----------------------------------------- | ------------------------------------------------------------------------ | --------
+bytes.Contains(value)                     | instr(@bytes, char(@value)) > 0                                          | EF Core 5.0
+bytes.Length                              | length(@bytes)                                                           | EF Core 5.0
+DateTime.Now                              | datetime('now', 'localtime')
+DateTime.Today                            | datetime('now', 'localtime', 'start of day')
+DateTime.UtcNow                           | datetime('now')
+dateTime.AddDays(value)                   | datetime(@dateTime, @value \|\| ' days')
```
roji, line 19:
```suggestion
DateTime.AddDays(value)                   | datetime(@dateTime, @value \|\| ' days')
```

entity-framework/core/providers/sqlite/built-in-functions.md
```diff
@@ -0,0 +1,76 @@
+---
+title: Built-in Functions - SQLite Database Provider - EF Core
```
roji, line 2: Maybe "Function translations" or even just "Translations"? Specifically for Sqlite not everything here is "built-in" (i.e. all the `ef_` stuff).

I'd also link to the spatial page from here when that arrives (this way this page covers all translations), and that's not built-in either.

roji (commented):
Breaking up into sections sounds good - some specific separate pages are also a good idea for larger topics (e.g. spatial, full-text search), with links from this main "function translation" page.

I'd vote for thematic sections rather than by type. There are potentially a huge number of types, and it could be quite bewildering to users who know they want to do something with binary types, but have no idea which .NET type is the right one.

entity-framework/core/providers/sqlite/built-in-functions.md
bricelam: I used lower-case (and the suffix *value*) to represent instance members. Syntax highlighting would also help reinforce that, but I’d have to add it manually since these can’t be code blocks.

entity-framework/core/providers/sqlite/built-in-functions.md
bricelam: I had that, but changed to match Smit’s docs. Happy to change back.
Definitely needs a link to spatial.

smitpatel: Built-in functions is right one. Most of them are truly C# functions in LINQ side. Plus, I am gonna add text at the start that definition is bit blurry. The built-in part in this page is that EF Core provides this function translations rather than use defining it. (Comes in the box)

smitpatel:
Breaking on type is not exactly individual types but rather groups - numeric/ datetime/ string etc. Anyway I can update when I "add" other providers.

entity-framework/core/providers/sqlite/functions.md
```diff
@@ -0,0 +1,103 @@
+---
+title: Function Mappings - SQLite Database Provider - EF Core
+description: Function Mappings of the SQLite EF Core database provider
+author: bricelam
+ms.date: 10/05/2020
+uid: core/providers/sqlite/built-in-functions
```
smitpatel, line 6: also update uid

Issue #2720, opened 2020-10-01
Document property bag mapping


ajcvickers:
See https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-5.0/whatsnew#property-bags

Issue #2018, opened 2020-01-02
Docs for indexer properties


AndriySvyryd:

ajcvickers:
See https://github.com/dotnet/efcore/issues/13610
https://github.com/dotnet/efcore/issues/15799

voronov-maxim:
Navigation property can be indexer property?

PR #2750, opened 2020-10-09
Add documentation for property bag entity types.

AndriySvyryd:
Part of #2720

### DIFF
--- a/entity-framework/core/modeling/shadow-properties.md
+++ b/entity-framework/core/modeling/shadow-properties.md
@@ -1,11 +1,11 @@
 ---
-title: Shadow Properties - EF Core
-description: Configuring shadow properties in an Entity Framework Core model
+title: Shadow and Indexer Properties - EF Core
+description: Configuring shadow and indexer properties in an Entity Framework Core model
 author: AndriySvyryd
-ms.date: 01/03/2020
+ms.date: 10/09/2020
 uid: core/modeling/shadow-properties
 ---
-# Shadow Properties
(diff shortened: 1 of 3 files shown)


entity-framework/core/modeling/shadow-properties.md
```diff
@@ -43,3 +43,12 @@ var blogs = context.Blogs
 ```
 
 Shadow properties cannot be accessed after a no-tracking query since the entities returned are not tracked by the change tracker.
+
+## Property bag entity types
+
+> [!NOTE]
+> Support for Property bag entity types was added in EF Core 5.0.
+
+Entity types that contain only indexer properties are known as property bag entity types. These entity types don't have shadow properties. Currently only `Dictionary<string, object>` is supported as a property bag entity type. This means that it must be configured as a shared entity type with a unique name and the corresponding `DbSet` property must be implemented using a `Set` call.
```
smitpatel, line 52: Where do we document indexer properties?

AndriySvyryd: Same page

AndriySvyryd: https://github.com/dotnet/EntityFramework.Docs/issues/2018

smitpatel: So you updated title without adding docs for indexer properties. I got confused in that.

smitpatel (approved):

PR #2774, opened 2020-10-15
Replace bold with headings in breaking changes pages

roji:
Keep the linter happy at all costs

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -5,6 +5,5 @@
     "MD025": { "front_matter_title": "" },
     "MD027": false,
     "MD028": false,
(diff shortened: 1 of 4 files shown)


roji:
Also reduced heading level to make the breaking changes appear in the right-hand contents pane, do we like?

AndriySvyryd:
The right-hand pane is pretty much unusable with this

roji:
I thought it was useful to have the list, but can info they part of it seems bad.

AndriySvyryd:
Perhaps just group them by impact?

roji:
I'll give it a go.

roji:
Done, looks pretty nice IMHO

AndriySvyryd (approved):

PR #2775, opened 2020-10-15
chore: Add paths filtering for markdownlint CI

nschonni:
Only triggers this job if Markdown files are touched, and not when the files under `entity-framework/ef6/` are changed


roji (approved):

roji:
Thanks @nschonni

PR #2776, opened 2020-10-15
chore: Only build samples if files are changed

nschonni:
Also includes the CI config itself to check when that is changed


roji (approved):
Thanks!

PR #2783, opened 2020-10-16
Remove notes about features being added to old versions

roji:


bricelam (changes requested):
Please keep all of these until 2.1 is out of support.

PR #2796, opened 2020-10-21
Fix incorrect tabs in backing field docs

roji:


entity-framework/core/modeling/backing-field.md
```diff
@@ -34,6 +34,8 @@ You can also configure backing fields by using a Data Annotation (available in E
 
 [!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/BackingField.cs?name=BackingField&highlight=5)]
 
+***
```
smitpatel, line 37: Is *** has any meaning?

roji: This is what terminates the tabs section that comes before. If you look at [the page](https://docs.microsoft.com/en-us/ef/core/modeling/backing-field?tabs=fluent-api#field-only-properties), you can see everything that comes after this section (e.g. field-only properties) is mistakenly included in the "Fluent API" tab because this is missing.

smitpatel: Or the blank line did it?

roji: Why the blank line? The *** is [the documented way](https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html#tabbed-content) to terminate tabbed content in docfx, we do it everywhere else (e.g. [here](https://raw.githubusercontent.com/dotnet/EntityFramework.Docs/master/entity-framework/core/modeling/entity-properties.md)).

smitpatel (approved):

PR #2827, opened 2020-11-02
Add scaffolding examples for --no-onconfiguring

jeremycook:


entity-framework/core/miscellaneous/cli/dotnet.md
```diff
@@ -206,6 +206,18 @@ The following example scaffolds only selected tables and creates the context in
 dotnet ef dbcontext scaffold "Server=(localdb)\mssqllocaldb;Database=Blogging;Trusted_Connection=True;" Microsoft.EntityFrameworkCore.SqlServer -o Models -t Blog -t Post --context-dir Context -c BlogContext --context-namespace New.Namespace
 ```
 
+The following example demonstrates using the `name=<name of connection string>` syntax to get the connection string from the configuration sources that are set up for the project. In this case "BloggingConnection" would be the name of a connection string in configuration.
+
+```dotnetcli
+dotnet ef dbcontext scaffold "c" Microsoft.EntityFrameworkCore.SqlServer
```
ErikEJ, line 212: Looks like something is missing here

jeremycook: Thanks @EricEJ . RDP delays are killing me today.

entity-framework/core/miscellaneous/cli/dotnet.md
```diff
@@ -206,6 +206,18 @@ The following example scaffolds only selected tables and creates the context in
 dotnet ef dbcontext scaffold "Server=(localdb)\mssqllocaldb;Database=Blogging;Trusted_Connection=True;" Microsoft.EntityFrameworkCore.SqlServer -o Models -t Blog -t Post --context-dir Context -c BlogContext --context-namespace New.Namespace
 ```
 
+The following example demonstrates using the `name=<name of connection string>` syntax to get the connection string from the configuration sources that are set up for the project. In this case "BloggingConnection" would be the name of a connection string in configuration.
+
+```dotnetcli
+dotnet ef dbcontext scaffold "name=BloggingConnection" Microsoft.EntityFrameworkCore.SqlServer
+```
+
+The following example skips scaffolding an `OnConfiguring` method. This can be useful for ensuring a development connection string or one with a username and password does not get written into the DbContext that is scaffolded. Added in EF Core 5.0.
```
bricelam, line 215: > This can be useful for ensuring a development connection string or one with a username and password does not get written into the DbContext that is scaffolded.

I disagree with this statement. I believe it's only useful when your DbContext is configured outside of the class, for example, inside Startup.ConfigureServices.

The `name=` syntax is more useful for keeping the connection string out of the code since it (unlike `--no-onconfiguring`) generates a DbContext that can be used without modification.

bricelam: @ajcvickers Thoughts?

jeremycook: I think @ajcvickers is on the right trail. Feel free to tell me what you would like in the sentence or if you would prefer it removed and I'll do that. People after all can draw their own conclusions about why a flag is useful to them.

bricelam, line 215:
```suggestion
The following example skips scaffolding an `OnConfiguring` method. This can be useful when you want to configure the DbContext outside of the class. For example, ASP.NET Core apps typically configure it in Startup.ConfigureServices. Added in EF Core 5.0.
```

bricelam (approved):

PR #2853, opened 2020-11-09
Add documentation for complete discriminator mapping

### LINKED ISSUE #2231: Document complete discriminator mapping
See https://github.com/dotnet/efcore/issues/18106

smitpatel:
Resolves #2231

### DIFF
--- a/entity-framework/core/modeling/inheritance.md
+++ b/entity-framework/core/modeling/inheritance.md
@@ -29,7 +29,7 @@ If you don't want to expose a `DbSet` for one or more entities in the hierarchy,
 
 By default, EF maps the inheritance using the *table-per-hierarchy* (TPH) pattern. TPH uses a single table to store the data for all types in the hierarchy, and a discriminator column is used to identify which type each row represents.
 
-The model above is mapped to the following database schema (note the implicitly-created `Discriminator` column, which identifies which type of `Blog` is stored in each row).
+The model above is mapped to the following database schema (note the implicitly created `Discriminator` column, which identifies which type of `Blog` is stored in each row).
 
 ![Screenshot of the results of querying the Blog entity hierarchy using table-per-hierarchy pattern](_static/inheritance-tph-data.png)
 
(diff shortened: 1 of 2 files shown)


entity-framework/core/modeling/inheritance.md
```diff
@@ -45,6 +45,10 @@ Finally, the discriminator can also be mapped to a regular .NET property in your
 
 [!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/NonShadowDiscriminator.cs?name=NonShadowDiscriminator&highlight=4)]
 
+When querying for derived entities, which are part of TPH pattern, EF Core adds a predicate over discriminator column in the query. This filter makes sure that we don't get any additional row for base types or sibling types not in the result. This filter predicate is skipped for base entity type since querying for base entity will get results for all the entities in the hierarchy. When materializing results from a query, if we come across a discriminator value, which isn't mapped to any entity type in the model, we throw an exception since we don't know how to materialize the results. This error only occurs if your database contains rows with discriminator values, which aren't mapped in EF Model. If you have such data, then you can mark the discriminator mapping in EF Core model as incomplete to indicate that we should always add filter predicate for querying any type in the hierarchy. `IsComplete` API on discriminator configuration marks the mapping to be complete/incomplete based on bool value passed.
```
AndriySvyryd, line 48:
```suggestion
When querying for derived entities, which use the TPH pattern, EF Core adds a predicate over discriminator column in the query. This filter makes sure that we don't get any additional rows for base types or sibling types not in the result. This filter predicate is skipped for the base entity type since querying for the base entity will get results for all the entities in the hierarchy. When materializing results from a query, if we come across a discriminator value which isn't mapped to any entity type in the model, we throw an exception since we don't know how to materialize the results. This error only occurs if your database contains rows with discriminator values which aren't mapped in the EF model. If you have such data, then you can mark the discriminator mapping in EF Core model as incomplete to indicate that we should always add filter predicate for querying any type in the hierarchy. `IsComplete(false)` call on the discriminator configuration marks the mapping to be incomplete.
```

AndriySvyryd (approved):

PR #2865, opened 2020-11-11
adding documentation for query null semantics

### LINKED ISSUE #1612: Create topic for relational vs. in-memory null equality/semantics
I couldn't find anything about this in the docs today, and it is one of the most referenced ideas from our uservoice site: http://data.uservoice.com/forums/72025-ado-net-entity-framework-ef-feature-suggestions/suggestions/1015361-incorrect-handling-of-null-variables-in-where-cl?ref=title#suggestion-1015361

---
#### Document Details

* ID: 91606b46-d8a2-8824-a1ae-522f35362ec4
* Version Independent ID: 22a92f99-5e43-96d7-82d1-cd02ccdc5af8
* Content: [Querying Data - EF Core](https://docs.microsoft.com/en-us/ef/core/querying/)
* Content Source: [entity-framework/core/querying/index.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/querying/index.md)
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

maumar:
Fixes #1612

### DIFF
--- a/entity-framework/core/performance/efficient-querying.md
+++ b/entity-framework/core/performance/efficient-querying.md
@@ -130,7 +130,7 @@ info: Microsoft.EntityFrameworkCore.Database.Command[20101]
 
 What's going on here? Why are all these queries being sent for the simple loops above? With lazy loading, a Blog's Posts are only (lazily) loaded when its Posts property is accessed; as a result, each iteration in the inner foreach triggers an additional database query, in its own roundtrip. As a result, after the initial query loading all the blogs, we then have another query *per blog*, loading all its posts; this is sometimes called the *N+1* problem, and it can cause very significant performance issues.
 
-Assuming we're going to need all of the blogs' posts, it makes sense to use eager loading here instead. We can use the [Include](xref:core/querying/related-data/eager#eager-loading) operator to perform the loading, but since we only need the Blogs' URLs (and we should only [load what's needed](xref:core/performance/efficient-updating#project-only-properties-you-need)). So we'll use a projection instead:
+Assuming we're going to need all of the blogs' posts, it makes sense to use eager loading here instead. We can use the [Include](xref:core/querying/related-data/eager#eager-loading) operator to perform the loading, but since we only need the Blogs' URLs (and we should only [load what's needed](xref:core/performance/efficient-querying#project-only-properties-you-need)). So we'll use a projection instead:
 
 [!code-csharp[Main](../../../samples/core/Performance/Program.cs#EagerlyLoadRelatedAndProject)]
 
@@ -202,3 +202,7 @@ For more information, see the page on [async programming](xref:core/miscellaneou
 
 > [!WARNING]
 > Avoid mixing synchronous and asynchronous code in the same application - it's very easy to inadvertently trigger subtle thread-pool starvation issues.
+
+## Additional resources
+
+See the [performance section](xref:core/querying/null-comparisons#writing-performant-queries) of the null comparison documentation page for some best practices when comparing nullable values.
--- a/entity-framework/core/querying/user-defined-function-mapping.md
+++ b/entity-framework/core/querying/user-defined-function-mapping.md
@@ -3,7 +3,7 @@ title: User-defined function mapping - EF Core
 description: Mapping user-defined functions to database functions
 author: maumar
 ms.date: 11/23/2020
-uid: core/user-defined-function-mapping
+uid: core/querying/user-defined-function-mapping
 ---
 # User-defined function mapping
 
@@ -88,6 +88,52 @@ Produces the following SQL:
 SELECT 100 * (ABS(CAST([p].[BlogId] AS float) - 3) / ((CAST([p].[BlogId] AS float) + 3) / 2))
 FROM [Posts] AS [p]
 ```
+
+## Configuring nullability of user-defined function based on its arguments
+
+If the user-defined function can only return `null` when one or more of its arguments are `null`, EFCore provides way to specify that, resulting in more performant SQL. It can be done by adding a `PropagatesNullability()` call to the relevant function parameters model configuration.
+
+To illustrate this, define user function `ConcatStrings`:
+
+```sql
+CREATE FUNCTION [dbo].[ConcatStrings] (@prm1 nvarchar(max), @prm2 nvarchar(max))
+RETURNS nvarchar(max)
+AS
+BEGIN
+    RETURN @prm1 + @prm2;
+END
+```
+
+and two CLR methods that map to it:
+
+[!code-csharp[Main](../../../samples/core/Querying/UserDefinedFunctionMapping/Model.cs#NullabilityPropagationFunctionDefinition)]
+
+The model configuration (inside `OnModelCreating` method) is as follows:
+
+[!code-csharp[Main](../../../samples/core/Querying/UserDefinedFunctionMapping/Model.cs#NullabilityPropagationModelConfiguration)]
+
+The first function is configured in the standard way. The second function is configured to take advantage of the nullability propagation optimization, providing more information on how the function behaves around null parameters.
+
+When issuing the following queries:
+
+[!code-csharp[Main](../../../samples/core/Querying/UserDefinedFunctionMapping/Program.cs#NullabilityPropagationExamples)]
+
+We get this SQL:
+
+```sql
+SELECT [b].[BlogId], [b].[Rating], [b].[Url]
+FROM [Blogs] AS [b]
+WHERE ([dbo].[ConcatStrings]([b].[Url], CONVERT(VARCHAR(11), [b].[Rating])) <> N'Lorem ipsum...') OR [dbo].[ConcatStrings]([b].[Url], CONVERT(VARCHAR(11), [b].[Rating])) IS NULL
+
+SELECT [b].[BlogId], [b].[Rating], [b].[Url]
+FROM [Blogs] AS [b]
+WHERE ([dbo].[ConcatStrings]([b].[Url], CONVERT(VARCHAR(11), [b].[Rating])) <> N'Lorem ipsum...') OR ([b].[Url] IS NULL OR [b].[Rating] IS NULL)
+```
+
+The second query doesn't need to re-evaluate the function itself to test its nullability.
+
+> [!NOTE]
+> This optimization should only be used if the function can only return `null` when it's parameters are `null`.
 
 ## Mapping a queryable function to a table-valued function
 
--- a/samples/core/Querying/NullSemantics/NullSemanticsContext.cs
+++ b/samples/core/Querying/NullSemantics/NullSemanticsContext.cs
@@ -1,0 +1,33 @@
+﻿using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+
+namespace NullSemantics
+{
+    public class NullSemanticsContext : DbContext
+    {
+        public DbSet<NullSemanticsEntity> Entities { get; set; }
+
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            var relationalNulls = false;
+            if (relationalNulls)
+            {
+                #region UseRelationalNulls
+                new SqlServerDbContextOptionsBuilder(optionsBuilder).UseRelationalNulls(true);
+                #endregion
+            }
+
+            optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=NullSemanticsSample;Trusted_Connection=True;MultipleActiveResultSets=true");
+        }
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+        {
+            modelBuilder.Entity<NullSemanticsEntity>().HasData(
+                new NullSemanticsEntity { Id = 1, Int = 1, NullableInt = 1, String1 = "A", String2 = "A" },
+                new NullSemanticsEntity { Id = 2, Int = 2, NullableInt = 2, String1 = "A", String2 = "B" },
+                new NullSemanticsEntity { Id = 3, Int = 2, NullableInt = null, String1 = null, String2 = "A" },
+                new NullSemanticsEntity { Id = 4, Int = 2, NullableInt = null, String1 = "B", String2 = null },
+                new NullSemanticsEntity { Id = 5, Int = 1, NullableInt = 3, String1 = null, String2 = null });
+        }
+    }
+}
--- a/samples/core/Querying/NullSemantics/Program.cs
+++ b/samples/core/Querying/NullSemantics/Program.cs
@@ -1,0 +1,71 @@
+﻿using Microsoft.EntityFrameworkCore;
+using System.Linq;
+
+namespace NullSemantics
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            using var context = new NullSemanticsContext();
+            context.Database.EnsureDeleted();
+            context.Database.EnsureCreated();
+
+            //#region FunctionSqlRaw
+            //context.Database.ExecuteSqlRaw(
+            //    @"create function [dbo].[ConcatStrings] (@prm1 nvarchar(max), @prm2 nvarchar(max))
+            //        returns nvarchar(max)
+            //        as
+            //        begin
+            //            return @prm1 + @prm2;
+            //        end");
+            //#endregion
+
+            //BasicExamples();
+            Functions();
+            //ManualOptimization();
+        }
+
+        static void BasicExamples()
+        {
+            using var context = new NullSemanticsContext();
+            #region BasicExamples
+            var query1 = context.Entities.Where(e => e.Id == e.Int);
+            var query2 = context.Entities.Where(e => e.Id == e.NullableInt);
+            var query3 = context.Entities.Where(e => e.Id != e.NullableInt);
+            var query4 = context.Entities.Where(e => e.String1 == e.String2);
+            var query5 = context.Entities.Where(e => e.String1 != e.String2);
+            #endregion
+
+            var result1 = query1.ToList();
+            var result2 = query2.ToList();
+            var result3 = query3.ToList();
+            var result4 = query4.ToList();
+            var result5 = query5.ToList();
+        }
+
+        static void Functions()
+        {
+            using var context = new NullSemanticsContext();
+
+            #region Functions
+            var query = context.Entities.Where(e => e.String1.Substring(0, e.String2.Length) == null);
+            #endregion
+
+            var result = query.ToList();
+        }
+
+        static void ManualOptimization()
+        {
+            using var context = new NullSemanticsContext();
+
+            #region ManualOptimization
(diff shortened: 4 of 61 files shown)


entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+
+The generated SQL is as follows (we don't need to evaluate the `SUBSTRING` function):
+
+```sql
+SELECT [e].[Id], [e].[Int], [e].[NullableInt], [e].[String1], [e].[String2]
+FROM [Entities] AS [e]
+WHERE [e].[String1] IS NULL OR [e].[String2] IS NULL
+```
+
+This optimization can also be used for user defined functions. This can be done by adding `PropagatesNullability()` call to relevant function parameters model configuration.
```
maumar, line 68: TODO: reference udf docs

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+FROM [Entities] AS [e]
+WHERE [e].[String1] IS NOT NULL AND [e].[String2] IS NOT NULL
+```
+
+Just like with built-in `Substring` function, the second query doesn't need to evaluate the function itself to test it's nullability.
+
+> [!NOTE]
+> This optimization should only be used if the function can only return `null` becuase it's parameters are `null`.
+
+## Optimizations and considerations
```
maumar, line 100: Any ideas for better name? Wanted to use "best practices" but acrolinx is complaining

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
+---
+title: Query null semantics in EF Core
```
smitpatel, line 2: Null comparisons in queries

(EF Core is auto-appended)

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+description: Information on how Entity Framework Core handles null comparisons in queries 
+author: maumar
+ms.date: 11/11/2020
+uid: core/querying/null-semantics
```
smitpatel, line 6: null-comparisons also rename file.

We refer to null semantics but let's use a more understandable word?

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+
+Third query introduces the null check. When `NullableInt` is `null` the comparison `Id <> NullableInt` yields `null`, which would be filtered out by `WHERE` operation. However from the boolean logic perspective this case should be returned as part of the result. Hence EF Core adds the necessary check to ensure that.
+
+```sql
+SELECT [e].[Id], [e].[Int], [e].[NullableInt], [e].[String1], [e].[String2]
+FROM [Entities] AS [e]
+WHERE ([e].[Id] <> [e].[NullableInt]) OR [e].[NullableInt] IS NULL
+```
+
+Queries four and five show the pattern when both columns are nullable. It's worth noting that `<>` operation produces more complicated (and slower) query than equality operation.
```
smitpatel, line 41:
```suggestion
Queries four and five show the pattern when both columns are nullable. It's worth noting that `<>` operation produces more complicated (and potentially slower) query than `==` operation.
```

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+## Null semantics in functions
+
+Many functions in SQL can only return `null` result if some of their arguments are `null`. EF Core takes advantage of this to produce more efficient queries.
+The query below illustrates the optimization:
+
+[!code-csharp[Main](../../../samples/core/Querying/NullSemantics/Program.cs#Functions)]
+
+The generated SQL is as follows (we don't need to evaluate the `SUBSTRING` function):
```
smitpatel, line 60:
```suggestion
The generated SQL is as follows (we don't need to evaluate the `SUBSTRING` function since it will be only null when either of the arguments to it is null.):
```

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+
+```sql
+SELECT [e].[Id], [e].[Int], [e].[NullableInt], [e].[String1], [e].[String2]
+FROM [Entities] AS [e]
+WHERE [e].[String1] IS NULL OR [e].[String2] IS NULL
+```
+
+The optimization can also be used for user defined functions. It can be done by adding `PropagatesNullability()` call to the relevant function parameters model configuration.
```
smitpatel, line 68: Should move this to user-defined function mapping page and add reference to here.

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+
+SELECT [e].[Id], [e].[Int], [e].[NullableInt], [e].[String1], [e].[String2]
+FROM [Entities] AS [e]
+WHERE [e].[String1] IS NOT NULL AND [e].[String2] IS NOT NULL
+```
+
+Just like with built-in `Substring` function, the second query doesn't need to evaluate the function itself to test its nullability.
+
+> [!NOTE]
+> This optimization should only be used if the function can only return `null` becuase it's parameters are `null`.
```
smitpatel, line 98:
```suggestion
> This optimization should only be used if the function can only return `null` when it's parameters are `null`.
```

entity-framework/core/querying/null-semantics.md
smitpatel: Write performant queries

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Optimizations and considerations
+
+- Comparing non-nullable columns is simpler and faster than comparing nullable columns. Consider marking columns as non-nullable wherever it is possible.
+
+- `Equal` comparison is simpler and faster than `NotEqual`, because query doesn't need to distinguish between `null` and `false` result. Use equality comparison whenever it's possible.
```
smitpatel, line 104: Should use ==/!= operators.

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+
+> [!NOTE]
+> Wrapping `Equal` comparison around `Not` is effectively `NotEqual`. `Equal` comparison is only faster/simpler if it is not negated.
```
smitpatel, line 107: Same operators. Can use example too.

smitpatel: Don't add note in middle of a bulleted list. Inline it or make a new point.

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+The optimization can also be used for user defined functions. It can be done by adding `PropagatesNullability()` call to the relevant function parameters model configuration.
+To illustrate this, define two user functions inside the `DbContext`:
+
+[!code-csharp[Main](../../../samples/core/Querying/NullSemantics/NullSemanticsContext.cs#UdfBody)]
+
+The model configuration (inside `OnModelCreating` method) is as follows:
+
+[!code-csharp[Main](../../../samples/core/Querying/NullSemantics/NullSemanticsContext.cs#UdfModelConfiguration)]
+
+First function is configured in a standard way. The second function is configured to take advantage of the nullability propagation optimization.
```
roji, line 77:
```suggestion
The first function is configured in the standard way. The second function is configured to take advantage of the nullability propagation optimization, providing more information on how the function behaves around null parameters.
```

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+SELECT [e].[Id], [e].[Int], [e].[NullableInt], [e].[String1], [e].[String2]
+FROM [Entities] AS [e]
+WHERE [dbo].[ConcatStrings]([e].[String1], [e].[String2]) IS NOT NULL
+
+SELECT [e].[Id], [e].[Int], [e].[NullableInt], [e].[String1], [e].[String2]
+FROM [Entities] AS [e]
+WHERE [e].[String1] IS NOT NULL AND [e].[String2] IS NOT NULL
+```
+
+Just like with built-in `Substring` function, the second query doesn't need to evaluate the function itself to test its nullability.
```
roji, line 95:
```suggestion
Just like with the built-in `Substring` function, the second query doesn't need to evaluate the function itself to test its nullability.
```

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+
+> [!NOTE]
+> This optimization should only be used if the function can only return `null` becuase it's parameters are `null`.
+
+## Optimizations and considerations
+
+- Comparing non-nullable columns is simpler and faster than comparing nullable columns. Consider marking columns as non-nullable wherever it is possible.
```
roji, line 102:
```suggestion
- Comparing non-nullable columns is simpler and faster than comparing nullable columns. Consider marking columns as non-nullable whenever possible.
```

entity-framework/core/querying/null-semantics.md
roji: Another alternative is "Checking for equality (`==`) is simpler and faster than checking for non-equality (`!=`)".

Also as above:

```suggestion
- `Equal` comparison is simpler and faster than `NotEqual`, because query doesn't need to distinguish between `null` and `false` result. Use equality comparison whenever possible.
```

entity-framework/core/querying/null-semantics.md
roji, line 107:
```suggestion
> Wrapping an `Equal` comparison around `Not` is effectively the same as `NotEqual`. An `Equal` comparison is only faster/simpler if it is not negated.
```

samples/core/Querying/NullSemantics/Program.cs
```diff
@@ -0,0 +1,80 @@
...
+            var query1 = context.Entities.Where(e => context.ConcatStrings(e.String1, e.String2) != null);
+            var query2 = context.Entities.Where(e => context.ConcatStringsOptimized(e.String1, e.String2) != null);
+            #endregion
+
+            var result = query.ToList();
+            var result1 = query1.ToList();
+            var result2 = query2.ToList();
+        }
+
+        static void ManulaOptimization()
```
roji, line 67:
```suggestion
        static void ManualOptimization()
```

samples/core/Querying/NullSemantics/Program.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            using var context = new NullSemanticsContext();
+
+            #region ManualOptimization
+            var query1 = context.Entities.Where(e => e.String1 != e.String1 || e.String1.Length == e.String1.Length);
```
roji, line 72: I get that this technically shows what we want to show, but comparing a string to itself like this is something nobody will ever write and may be confusing to readers... Maybe consider a query that could make more sense in the real world (i.e. with two different properties).

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+
+- `Equal` comparison is simpler and faster than `NotEqual`, because query doesn't need to distinguish between `null` and `false` result. Use equality comparison whenever it's possible.
+
+> [!NOTE]
+> Wrapping `Equal` comparison around `Not` is effectively `NotEqual`. `Equal` comparison is only faster/simpler if it is not negated.
+
+- In some cases, it is possible to simplify a complex comparison by filtering out `null` values from a column explicitly - for example when no `null` values are present or these values are not relevant in the result. Consider the following example:
```
roji, line 109: Users are probably going to ask why we're not doing that for them :)

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+SELECT [e].[Id], [e].[Int], [e].[NullableInt], [e].[String1], [e].[String2]
+FROM [Entities] AS [e]
+WHERE ((([e].[String1] <> [e].[String1]) OR [e].[String1] IS NULL) AND [e].[String1] IS NOT NULL) OR ((CAST(LEN([e].[String1]) AS int) = CAST(LEN([e].[String1]) AS int)) OR [e].[String1] IS NULL)
+
+SELECT [e].[Id], [e].[Int], [e].[NullableInt], [e].[String1], [e].[String2]
+FROM [Entities] AS [e]
+WHERE [e].[String1] IS NOT NULL AND (CAST(LEN([e].[String1]) AS int) = CAST(LEN([e].[String1]) AS int))
+```
+
+In the second query, `null` results are filtered out from `String1` column explicitly. EF Core can safely treat the `String1` column as non-nullable during comparison, which results is simpler query.
```
roji, line 125:
```suggestion
In the second query, `null` results are filtered out from `String1` column explicitly. EF Core can safely treat the `String1` column as non-nullable during comparison, resulting in a simpler query.
```

entity-framework/core/querying/null-semantics.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Using relational null semantics
+
+It's possible to disable the null comparison compensation and use relational null semantics directly. This can be done by calling `UseRelationalNulls(true)` method on the options builder inside `OnConfiguring` method:
```
roji, line 129: We may want to add a warning such as:

> [!WARNING]
> When using relational null semantics, your LINQ queries no longer have the same meaning as they do in C#, and may yield different results than expected. Exercise caution when using this mode.

entity-framework/core/querying/null-semantics.md
roji: We should also link to this from [the perf docs on efficient querying](https://docs.microsoft.com/en-us/ef/core/performance/efficient-querying), possibly add an "Additional resources" section there to link to here.

roji (commented):
This is a really nice addition to our docs.

Once general comment about the samples, is that in some cases the queries don't seem like something users would actually write, and might therefore be confusing:

For example:

```c#
var query1 = context.Entities.Where(e => context.ConcatStrings(e.String1, e.String2) != null);
```

It's hard to imagine someone writing this (as opposed to comparing the result of ConcatStrings to some constant).

entity-framework/core/querying/null-semantics.md
maumar: we have a good alibi - we don't know if there are nulls in the data or when they are present, that they are irrelevant.

samples/core/Samples.sln
```diff
@@ -113,20 +113,22 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebApplication1.Migrations"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebApplication1", "Schemas\ThreeProjectMigrations\WebApplication1\WebApplication1.csproj", "{A15F08F8-966D-4A38-A9FF-F0B3FB3335BE}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ConnectionInterception", "Miscellaneous\ConnectionInterception\ConnectionInterception.csproj", "{1CE06110-E69A-44BD-ACBE-D7F96B02288B}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ConnectionInterception", "Miscellaneous\ConnectionInterception\ConnectionInterception.csproj", "{1CE06110-E69A-44BD-ACBE-D7F96B02288B}"
```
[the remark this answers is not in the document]
maumar: nope, will revert

samples/core/Querying/NullSemantics/Program.cs
maumar: I remember starting with different strings but wasn't happy with the sql it generated. IIRC the optimization wasn't very clear, so I decided to go with this one. Will use different strings and we can judge which one feels better.

entity-framework/core/querying/null-semantics.md
maumar: I just responded to an issue about this and thought of the same thing. Some extra context/warning will be useful.

samples/core/Querying/NullSemantics/NullSemanticsContext.cs
```diff
@@ -0,0 +1,56 @@
+﻿using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+
+namespace NullSemantics
+{
+    public class NullSemanticsContext : DbContext
+    {
+		public DbSet<NullSemanticsEntity> Entities { get; set; }
```
roji, line 8: @maumar looks like there's tabs :fearful::fearful::fearful::fearful:

maumar: eradicated with extreme prejudice

smitpatel: Apocalypse now!

maumar:
updated

entity-framework/core/querying/user-defined-function-mapping.md
```diff
@@ -89,6 +89,52 @@ SELECT 100 * (ABS(CAST([p].[BlogId] AS float) - 3) / ((CAST([p].[BlogId] AS floa
 FROM [Posts] AS [p]
 ```
 
+## Configuring nullability of user-defined function based on it's arguments
```
roji, line 92:
```suggestion
## Configuring nullability of user-defined function based on its arguments
```

entity-framework/core/querying/user-defined-function-mapping.md
```diff
@@ -89,6 +89,52 @@ SELECT 100 * (ABS(CAST([p].[BlogId] AS float) - 3) / ((CAST([p].[BlogId] AS floa
...
+
+If the user-defined function can only return `null` when one or more of it's arguments are `null` EFCore provides way to specify that, which results in more performant SQL. It can be done by adding a `PropagatesNullability()` call to the relevant function parameters model configuration.
```
roji, line 94:
```suggestion
If the user-defined function can only return `null` when one or more of its arguments are `null`, EFCore provides way to specify that, resulting in more performant SQL. It can be done by adding a `PropagatesNullability()` call to the relevant function parameters model configuration.
```

entity-framework/core/performance/efficient-querying.md
```diff
@@ -202,3 +202,7 @@ For more information, see the page on [async programming](xref:core/miscellaneou
 
 > [!WARNING]
 > Avoid mixing synchronous and asynchronous code in the same application - it's very easy to inadvertently trigger subtle thread-pool starvation issues.
+
+## Additional resources
+
+See [performance section](xref:core/querying/null-comparisons#writing-performant-queries) of the null comparison documentation page for some best practices when comparing nullable values.
```
roji, line 208:
```suggestion
See the [performance section](xref:core/querying/null-comparisons#writing-performant-queries) of the null comparison documentation page for some best practices when comparing nullable values.
```

roji (approved):
Looks good from my perspective!

roji:
BTW just noticed there there's both null-comparisons.md and null-semantics.md (duplicate)

PR #2887, opened 2020-11-19
Format identifiers as code

MSDN-WhiteKnight:
When identifiers are not formatted as code it causes issues with localization (https://github.com/MicrosoftDocs/feedback/issues/2083)


ajcvickers:
Thanks!

PR #2950, opened 2020-12-16
Adding notice about DB creation and migration

sguryev:
`SqliteException` is thrown without calling it. Not sure about `Migrate()` though. I have a single migration and just `EnsureCreated()` works. Missing `EnsureCreated()` call causes `SqliteException : SQLite Error 1: 'no such table: XXX'` exception.

I have added exception message to the notice on purpose because devs usually paste them in google in order to find an answer

### DIFF
--- a/entity-framework/core/testing/sqlite.md
+++ b/entity-framework/core/testing/sqlite.md
@@ -39,6 +39,7 @@ Notice:
 * The `CreateInMemoryDatabase` method creates a SQLite in-memory database and opens the connection to it.
 * The created `DbConnection` is extracted from the `ContextOptions` and saved.
 * The connection is disposed when the test is disposed so that resources are not leaked.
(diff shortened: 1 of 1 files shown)


ajcvickers:
@sguryev I don't think this guidance is not correct. Can you explain in more detail why this is needed when working through the sample? Did you change/miss some step somewhere?

sguryev:
@ajcvickers I have followed all the steps and then went to google in order to find what's wrong. I would save фищге 20 mins if this info would be in place.

ajcvickers:
@sguryev The sample uses a seeding pattern that calls EnsureCreated. I verified that there is no error when using the sample code provided. I don't think it makes sense to add this text here because it implies that you need to do something additional to make the sample work, which is not the case.

Issue #2964, opened 2020-12-26
Calculate Rating as Double, optimise=true and other tweaks


DickBaker:
arising from roji's 2934 and
https://docs.microsoft.com/en-us/ef/core/performance/performance-diagnosis?tabs=simple-logging%2Cload-entities

The calculations sum / count as-is yield int when should yield double
Need optimise=true in the .csproj to satisfy Benchmark.net
needs a Console.Read() at end so results can be viewed/kept
updated EF packages to latest
various analyzer-suggested enhancements and needs .editorconfig file for consistency

I am raising a PR for this

BTW my results are
// * Summary *

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.1256 (1909/November2018Update/19H2)
Intel Core i7-4500U CPU 1.80GHz (Haswell), 1 CPU, 4 logical and 2 physical cores
.NET Core SDK=5.0.101
  [Host]     : .NET Core 5.0.1 (CoreCLR 5.0.120.57516, CoreFX 5.0.120.57516), X64 RyuJIT  [AttachedDebugger]
  DefaultJob : .NET Core 5.0.1 (CoreCLR 5.0.120.57516, CoreFX 5.0.120.57516), X64 RyuJIT

|                 Method |      Mean |     Error |    StdDev | Ratio | RatioSD | Gen 0 | Gen 1 | Gen 2 |  Allocated |
|----------------------- |----------:|----------:|----------:|------:|--------:|------:|------:|------:|-----------:|
|           LoadEntities | 41.559 ms | 11.102 ms | 32.032 ms |  5.96 |    6.30 |     - |     - |     - | 1312.48 KB |
| LoadEntitiesNoTracking | 31.226 ms |  6.760 ms | 18.845 ms |  4.26 |    3.17 |     - |     - |     - |  542.91 KB |
|     ProjectOnlyRanking | 12.785 ms |  1.413 ms |  3.891 ms |  1.57 |    0.73 |     - |     - |     - |  255.06 KB |
|    CalculateInDatabase |  9.215 ms |  1.529 ms |  4.313 ms |  1.00 |    0.00 |     - |     - |     - |   36.38 KB |

roji:
Closing as #2990 has been merged.

PR #2965, opened 2020-12-26
Dick bits

DickBaker:
calculate Rating as Double, optimise=true and other tweaks
Issue #2964


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/EntityFramework.Docs?pullRequest=2965) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/EntityFramework.Docs?pullRequest=2965) before we can accept your contribution.<br/><br/>:x: DickBaker [sign now](https://cla.dotnetfoundation.org/dotnet/EntityFramework.Docs?pullRequest=2965)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/EntityFramework.Docs?pullRequest=2965) it.</sub>

DickBaker (commented):
not sure why/where/how the .slnf came from [sorry!], but is not material to the PR

.editorconfig
```diff
@@ -0,0 +1,200 @@
+# editorconfig.org
```
roji, line 1: Having a .editorconfig for the samples isn't necessary a bad idea, but it's a good idea to split this off to another PR. Specifically, there are two different .editorconfig files in this PR, which also aren't the same as the one we currently use in EF Core itself.

roji (commented):
@DickBaker while some of your changes are interesting, the intent behind this specific benchmark is to show users how to set up a quick benchmark with minimal complexity, rather than to actually produce the most accurate results and write the most powerful/flexible/comprehensive benchmark. Most of your changes seem like they would make things harder for user new to benchmarking and EF.

It's best to understand exactly what it is you're trying to achieve here, keeping in mind the minimalistic goal for this benchmark.

Some specific comments:

> The calculations sum / count as-is yield int when should yield double

This is technically correct, though it wouldn't have any bearing on what this benchmark is actually doing. We can change this though.

>  Need optimise=true in the .csproj to satisfy Benchmark.net

This is a good idea, as it would allow users to get up and running more quickly.

>  needs a Console.Read() at end so results can be viewed/kept

This depends on how you're running the benchmarks - I'm assuming you're on VS, where the terminal window gets closed immediately; when running from the commandline, or from an IDE such as Jetbrains Rider, the additional ReadKey is actually disruptive. IMO this doesn't really belong in the benchmark code.

.editorconfig
DickBaker: I noticed that my CodeCleanup had wobbled some of the formatting, so I added mine as an explicit .editorconfig rather than have ambiguity. Got carried away and added 2 [whoops, like NYC so good they named it twice]. Happy that you should ditch my templates in favour of an "official one". I liked the simplicity of Roji's original that seemed to command instant wisdom of just pulling what you actually need (just cols/rows critters required not the whole prairie], but spotting the integer-division just had to follow-thru with a full-on OCD response! Honestly, not trying to over-complicate Roji's simple v1 main-dish, but to offer some vn+1 dessert as next course [maybe this should appear later once the newbie has joined the party and not frightened by too much noise. Yeah, maybe.

Anyway I have been exploring various Find/First style as well as Cold/Hot caching and even few basic ADO flavours for comparison & horses/courses. Missing so far is a commentary of when/what suitable, but I hope the BM results will speak for themselves. Your call, and I hope to revise PR 2965 with much more meat shortly and hope doesn't give you too much indigestion as pure vegans. [ok I give up speech-writing and will stay with my day-job!].

Thanks for your launching this perf journey as I find on my consultancy journey that there are many "unknown unknowns" that need a strong torch.

roji:
@DickBaker thanks for understanding and explaining - we're very interested in any perf results you find during your investigations.

Specifically for this PR and the samples... If you'd like to amend this PR to only contain the more minor improvements I mentioned above (e.g. add the division, the optimize-true), we can definitely merge that. Let me know how you'd like to proceed.

roji:
Am closing this as the relevant changes have been committed via #2990

PR #2966, opened 2020-12-26
fix: MD033/no-inline-html

nschonni:
Prevents missing escaping of generics.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -6,7 +6,13 @@
     },
     "MD025": { "front_matter_title": "" },
     "MD028": false,
-    "MD033": false,
+    "MD033": {
+        "allowed_elements": [
+            "a",
+            "nobr",
+            "sup"
+        ]
+    },
     "MD036": false,
     "MD046": {
         "style": "fenced"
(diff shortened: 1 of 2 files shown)


ajcvickers:
@nschonni  The reason this is disabled is because it prevents putting C# in HTML comments. You seem to have removed these comments. We can't take this unless there is a way to have both.

nschonni:
@ajcvickers I'll file something upstream, as I don't think it should be flagging stuff in the HTML comments, but I'm curious why removing/keeping commented out code for the migrated samples is important

ajcvickers:
@nschonni Because when editing documents I want to be able to see the code in the sample without having to find it in the cs file. If previews of the doc showed the code when editing it in VS Code, then I wouldn't need this.

nschonni:
That's probably a decent feature request for https://github.com/microsoft/vscode-docs-authoring/

I think the copy/paste of the code samples into comments means something is likely to be/get out of sync, and is why the current snippet system is used

ajcvickers:
Agreed. The code in the samples is the source of truth. If someone updates the samples without changing the copy in the docs it would not be the end-of-the-world. It's just useful for me when editing to know basically what the code does.

ajcvickers:
Closing this for now. We can revisit if things change.

PR #2994, opened 2021-01-09
Fix violation of MD009 - Trailing spaces

dennisseders:
This rule is triggered on any lines that end with unexpected white space. The trailing spaces should be removed from the end of the line.

### DIFF
--- a/entity-framework/ef6/what-is-new/index.md
+++ b/entity-framework/ef6/what-is-new/index.md
@@ -29,13 +29,13 @@ The EF 6.3.0 runtime was released to NuGet in September 2019. The main goal of t
 
 ### EF designer support
 
-There's currently no support for using the EF designer directly on .NET Core or .NET Standard projects or on an SDK-style .NET Framework project. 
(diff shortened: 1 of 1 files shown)


dennisseders:
See [markdownlint](https://github.com/DavidAnson/markdownlint/blob/v0.22.0/doc/Rules.md#md009) documentation for further reference.

dennisseders:
I have closed this pull request as it turned out that dozens occurrences of this warning exist within this repository. Fixing this for only one file therefore does not make sense. @roji I would like to propose that I create a pull request which contains fixes for all occurrences of this warning.

roji:
@dennisseders the EF6 part of the docs is no longer being actively developed - we are not trying to fix issues such as trailing whitespace, but generally only fix typos or false information. Note that we run markdownlint as part of the build pipeline for this docs repo (see .github/workflows/markdownlist.yml) and specifically exclude the EF6 docs (which indeed contain various linter issues); the EF Core part of the docs is completely clean, which is the important part.

PR #2780, opened 2020-10-15
chore: Markdownlint cleanups

nschonni:
Inline disabled in the case of old tab style headings

PR #2998, opened 2021-01-10
Improvements to value generation docs

### LINKED ISSUE #2947: Clarification for GUID generation behavior in EF Core 5.0 for SQL Server
## Ask a question

If a developer's goal is to have “sequential” GUID primary keys generated for faster performance, either by EF Core or the SQL Server’s NEWSEQUENTIALID, what is the expected behavior in EF Core 5.0 and what is the best practice?

[This documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.valuegeneration.sequentialguidvaluegenerator?view=efcore-5.0) says:

> (SequentialGuidValueGenerator Class) Generates sequential Guid values optimized for use in Microsoft SQL server clustered keys or indexes, yielding better performance than random values. This is the default generator for SQL Server Guid columns which are set to be generated on add."

Does this mean _if we don’t configure anything_ on a GUID property at all, EF Core 5 will automatically use its optimized sequential GUID generator to generate a non-temporary sequential GUID that will be saved to the DB? Or _do we have to specify ValueGeneratedOnAdd_?

But then there is this [other documentation](https://docs.microsoft.com/en-us/ef/core/modeling/generated-properties?tabs=data-annotations) which says something different:

>How the value is generated for added entities will depend on the database provider being used. Database providers may automatically set up value generation for some property types, but others may require you to manually set up how the value is generated.

>For example, when using SQL Server, values will be automatically generated for GUID properties (**using the SQL Server sequential GUID algorithm**)…

This seems to contradict the first link, which said the GUIDs would be generated on client side.

Some clarifications on this behavior and best practice would be appreciated, since we’d often opt to use GUID instead of int for primary keys to avoid giving away business information to end users.

It would also be nice to know the best approach to take when we do want non-sequential, random GUIDs for properties that are security-sensitive.

Thank you!

### Include provider and version information

EF Core version: 5.0
Database provider: Microsoft.EntityFrameworkCore.SqlServer
Target framework:  .NET 5.0
Operating system: Windows 10 Pro
IDE: Visual Studio 2019 16.8.3

### LINKED ISSUE #2980: Merge repeated sections
Related: https://github.com/dotnet/EntityFramework.Docs.pl-pl/issues/11

---
#### Document Details

* ID: 97d622b2-1e4b-4d76-b3df-fefaa5ba99cb
* Version Independent ID: ec833af2-275e-981f-689f-8c65edfc00cd
* Content: [Generated Values - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/generated-properties?tabs=fluent-api)
* Content Source: [entity-framework/core/modeling/generated-properties.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/generated-properties.md)
* GitHub Login: @AndriySvyryd

### LINKED ISSUE #2999: Document SQL Server IDENTITY options (seed/increment)
https://github.com/dotnet/efcore/issues/11924

roji:
This ended up rabbit-holing into bigger change than just #2980. @AndriySvyryd let me know what you think.

Fixes #2980
Fixes #2947
Closes #2999


.markdownlint.json
```diff
@@ -1,9 +1,7 @@
 {
     "default": true,
     "MD013": false,
-    "MD024": {
-        "allow_different_nesting": true
-    },
+    "MD024": false,
```
roji, line 4: /cc @nschonni, this reverts your change from #2780. The problem is when trying to have multiple tab-sets under the same heading with the same titles (e.g. FluentAPI/Data Annotations).

entity-framework/core/modeling/generated-properties.md
```diff
@@ -62,54 +55,84 @@ You can configure any property to have its value generated for inserted entities
...
+    WHERE BlogId = @Id
+END
+```
+
+For information on creating triggers, [see the documentation on using raw SQL in migrations](xref:core/managing-schemas/migrations/managing#adding-raw-sql).
 
-[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/ComputedColumn.cs?name=StoredComputedColumn)]
+## Overriding value generation
+
+Although a property is configured for value generation, in many cases you may still explicitly specify a value for it. Whether this will actually work depends on the specific value generation mechanism that has been configured; while you may specify an explicit value instead of using a column's default value, the same cannot be done with computed columns.
```
roji, line 120: This may be a bit light, compared to the "Setting Explicit Values for Generated Properties" page which this PR removes (though note that some of its content went into the SQL Server docs).

entity-framework/core/modeling/generated-properties.md
```diff
@@ -62,54 +55,84 @@ You can configure any property to have its value generated for inserted entities
...
+
+To override value generation with an explicit value, simply set the property to any value that is not the CLR default value for that property's type (`null` for `string`, `0` for `int`, `Guid.Empty` for `Guid`, etc.).
 
 > [!NOTE]
-> Support for creating stored computed columns was added in EF Core 5.0.
+> Trying to insert explicit values into SQL Server IDENTITY fails by default; [see these docs for a workaround](xref:core/providers/sql-server/value-generation#inserting-explicit-values-into-identity-columns).
+
+To provide an explicit value for properties that have been configured as value generated on add or update, you must also configure the property as follows:
+
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/ValueGeneratedOnAddOrUpdateWithPropertySaveBehavior.cs?name=ValueGeneratedOnAddOrUpdateWithPropertySaveBehavior&highlight=5)]
```
roji, line 129: Note that the original section in "Setting Explicit Values for Generated Properties" said to use PropertySaveBehavior.Ignore which didn't work for me - PropertySaveBehavior.Save does. Also, no exception was thrown when specifying Ignore (which is what the text said).

One last thing, when the property `ValueGeneratedOnAddOrUpdate()`, then overriding with explicit values works when *updating* (with PropertySaveBehavior.Save), but doesn't seem to work when *adding*...

.markdownlint.json
nschonni: Dotnet/docs uses individual `<!-- markdownlint disable MD024-->` statements when the old tab format causes issues instead of turning off rules entirely

roji: Is there an alternative new tab format the we can use instead, that doesn't have this problem?

If not, forcing these linter suppression in the doc codes seems to me more annoying/ugly than the value that MD024 brings, though that's subjective...

nschonni: I think there was supposed to be a `:::` version at some point awhile ago, but I think most of the docs for that type of thing is in the https://review.docs.microsoft.com/ site that isn't available to the public

entity-framework/core/modeling/keys.md
```diff
@@ -32,6 +34,10 @@ You can also configure multiple properties to be the key of an entity - this is
 
 [!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/KeyComposite.cs?name=KeyComposite&highlight=4)]
 
+## Value generation
+
+For non-composite numeric and Guid primary keys, EF Core sets up value generation for you by convention. For example, a numeric primary key in SQL Server is automatically set up to be an IDENTITY column. For more information, see [the documentation on value generation](xref:core/modeling/generated-properties).
```
AndriySvyryd, line 39:
```suggestion
For non-composite numeric and GUID primary keys, EF Core sets up value generation for you by convention. For example, a numeric primary key in SQL Server is automatically set up to be an IDENTITY column. For more information, see [the documentation on value generation](xref:core/modeling/generated-properties).
```

entity-framework/core/modeling/generated-properties.md
```diff
@@ -62,54 +55,84 @@ You can configure any property to have its value generated for inserted entities
...
+[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/ValueGeneratedOnAddOrUpdate.cs?name=ValueGeneratedOnAddOrUpdate&highlight=5)]
 
-You can also specify a SQL fragment that is used to calculate the default value:
+***
 
-[!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/DefaultValueSql.cs?name=DefaultValueSql&highlight=5)]
+> [!WARNING]
+> Unlike with default values or computed columns, we are not specifying *how* the values are to be generated; that depends on the database provider being used. Database providers may automatically set up value generation for some property types, but others may require you to manually set up how the value is generated.
+>
+> For example, on SQL Server, when a Guid property is configured as value generated on add, the provider automatically performs value generation client-side, using an algorithm to generate optimal sequential GUID values. However, specifying `ValueGeneratedOnAdd()` on a DateTime property will have no effect ([see the section below for DateTime value generation](#datetime-value-generation)).
```
AndriySvyryd, line 73:
```suggestion
> For example, on SQL Server, when a GUID property is configured as value generated on add, the provider automatically performs value generation client-side, using an algorithm to generate optimal sequential GUID values. However, specifying `ValueGeneratedOnAdd()` on a DateTime property will have no effect ([see the section below for DateTime value generation](#datetime-value-generation)).
```

AndriySvyryd (approved):

entity-framework/core/modeling/generated-properties.md
roji: Opened https://github.com/dotnet/efcore/issues/23912 to track the apparent adding bug.

Issue #685, opened 2018-04-18
Create article with workarounds for SQL Server features that don't have first-class support


divega:
| Feature          | Workaround                                                                                                             |
|:-----------------|:-----------------------------------------------------------------------------------------------------------------------|
| Spatial          | Link to workarounds for missing EF6 features                                                                           |
| JSON             | Value converter to parse/serialize JSON, JSON_VALUE, JSON_QUERY and JSON_MODIFY in FromSql or function mapping         |
| SQL Graph        | MATCH in FromSql, custom SQL in migrations. Can node and edge tables be treated as normal tables?                      |
| Temporal tables  | Custom SQL in EF Migrations and FromSql                                                                                |
| Always encrypted | Value converters, custom SQL in migrations, AsEnumerable() for explicit client eval (e.g. for comparisons, LIKE, etc.) |

Dzivo:
Any news on this ?

glautrou:
Hi @divega ,

> * [ ]  Temporal tables:
>   
>   * [ ]  AS OF, FROM  TO, BETWEEN AND, CONTAINED IN, ALL in FromSql

I wrote the [EfCoreTemporalTable library](https://github.com/glautrou/EfCoreTemporalTable) that automatically generates `FromSql` from LINQ queries with associated documentation, which is the goal of the task you mentionned.

However EF Core 6.0 milestone should natively handle temporal tables, let me know if you need my help about this soon outdated feature in EF docs and what can I do.

ErikEJ:
@glautrou Add it to the Tools page in Docs

PR #3020, opened 2021-01-17
Updates to value converter docs

### LINKED ISSUE #614: Value conversions documentation
I believe this can be added as a new page under https://github.com/aspnet/EntityFramework.Docs/tree/master/entity-framework/core/modeling.

### LINKED ISSUE #784: Use rowversion/timestamp convert to ulong as example for type conversions
See https://github.com/aspnet/EntityFrameworkCore/issues/5936

### LINKED ISSUE #2252: Missing disclaimer about incompatibility of generated values with ValueConversion
For past few hours, I have been trying to use strongly typed ids (e.g. struct CustomerId with int inside instead of a simple int) in a model and so far it looks to be impossible (for Sql Server at least).

HiLo, Identity, just ignore it and use .ValueGeneratedOnAdd.. I always hit brick wall of runtime exception.

Examples:
'Identity value generation cannot be used for the property 'Id' on entity type 'Blog' because the property type is 'Id'. Identity value generation can only be used with signed integer properties.'

System.NotSupportedException: 'Value generation is not supported for property 'Blog.Id' because it has a 'ValueConverter&lt;Id, int&gt;' converter configured. Configure the property to not use value generation using 'ValueGenerated.Never' or 'DatabaseGeneratedOption.None' and specify explicit values instead.'

This should be documented on the page.

---
#### Document Details

* ID: 97d622b2-1e4b-4d76-b3df-fefaa5ba99cb
* Version Independent ID: ec833af2-275e-981f-689f-8c65edfc00cd
* Content: [Generated Values - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/generated-properties?tabs=data-annotations)
* Content Source: [entity-framework/core/modeling/generated-properties.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/generated-properties.md)
* GitHub Login: @AndriySvyryd
* Microsoft Alias: **ansvyryd**

### LINKED ISSUE #2415: value-conversions - limitations
is there an open issue being tracked for this statement:

"Removal of these limitations is being considered for a future release"

If yes, should / could it be linked to in the docs?

---
#### Document Details

* ID: 03dfe9f6-e831-ac3f-0b90-acb9b4c0f828
* Version Independent ID: a26156d1-a4b3-7df0-e63e-b8e9f804abbf
* Content: [Value Conversions - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/value-conversions)
* Content Source: [entity-framework/core/modeling/value-conversions.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/value-conversions.md)
* GitHub Login: @ajcvickers

### LINKED ISSUE #2456: Missing Built-in Value Converters
The supported value converters [here](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.storage.valueconversion?view=efcore-3.1), e.g. `StringToDateTimeConverter()`, are not all listed on this page.

---
#### Document Details

* ID: 03dfe9f6-e831-ac3f-0b90-acb9b4c0f828
* Version Independent ID: a26156d1-a4b3-7df0-e63e-b8e9f804abbf
* Content: [Value Conversions - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/value-conversions)
* Content Source: [entity-framework/core/modeling/value-conversions.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/value-conversions.md)
* GitHub Login: @ajcvickers

### LINKED ISSUE #2824: Should mention the necessity of a value comparer
Many developers will read this (excellent) writeup on value converters, and not realize the necessity of a related value comparer. You really need a section that warns readers that creating a value converter often requires an associated value comparer, with a sentence or two explaining, at a high level, under what conditions a value comparer is necessary, and giving a link to the page on value converters.

---
#### Document Details

* ID: 03dfe9f6-e831-ac3f-0b90-acb9b4c0f828
* Version Independent ID: a26156d1-a4b3-7df0-e63e-b8e9f804abbf
* Content: [Value Conversions - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/value-conversions)
* Content Source: [entity-framework/core/modeling/value-conversions.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/modeling/value-conversions.md)
* GitHub Login: @ajcvickers

### LINKED ISSUE #2979: Document how to use a value comparer to compensate for different behavior of string comparisons in client and store
Typically string comparisons performed on the store can have different behaviors to comparisons in memory, e.g. trailing blanks and casing may be ignored in the store. 

In previous versions of EF this made things like fixup of string keys hard to predict: related data would be pulled from the database because the FK and PK values matched according to the server semantics but once in memory the strings didn't match and therefore fixup didn't happen. This is described in more detail in [CodePlex dotnet/efcore#284](https://entityframework.codeplex.com/workitem/284).

This kind of difference appears more often when working with existing data, so how important this is depends on how much we prioritize those scenarios. If the data was stored by EF then string keys have a better chance to be an exact match.

The idea to address this is that on the client, EF Core could always use a custom string comparison algorithm that would mimic the behavior in the database. That is, one that ignores trailing blanks and casing.

ajcvickers:
Lots of examples!

Fixes #614
Fixes #784
Part of #685
Fixes #2252
Fixes #2415
Fixes #2456
Fixes #2824
Fixes #2979

### DIFF
--- a/entity-framework/core/modeling/value-comparers.md
+++ b/entity-framework/core/modeling/value-comparers.md
@@ -37,7 +37,7 @@ Consider byte arrays, which can be arbitrarily large. These could be compared:
 * By reference, such that a difference is only detected if a new byte array is used
 * By deep comparison, such that mutation of the bytes in the array is detected
 
-By default, EF Core uses the first of these approaches for non-key byte arrays. That is, only references are compared and a change is detected only when an existing byte array is replaced with a new one. This is a pragmatic decision that avoids copying entire arrays and comparing them byte-to-byte when executing <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A>. It means the common scenario of replacing, say, one image with another is handled in a performant way.
+By default, EF Core uses the first of these approaches for non-key byte arrays. That is, only references are compared and a change is detected only when an existing byte array is replaced with a new one. This is a pragmatic decision that avoids copying entire arrays and comparing them byte-to-byte when executing <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A>. It means that the common scenario of replacing, say, one image with another is handled in a performant way.
 
 On the other hand, reference equality would not work when byte arrays are used to represent binary keys, since it's very unlikely that an FK property is set to the _same instance_ as a PK property to which it needs to be compared. Therefore, EF Core uses deep comparisons for byte arrays acting as keys; this is unlikely to have a big performance hit since binary keys are usually short.
 
@@ -70,9 +70,7 @@ EF Core has built-in support for generating compiled, memberwise comparisons of 
 
 ## Mutable classes
 
-It is recommended that you use immutable types (classes or structs) with value converters when possible. This is usually more efficient and has cleaner semantics than using a mutable type.
-
-However, that being said, it is common to use properties of types that the application cannot change. For example, mapping a property containing a list of numbers:
+It is recommended that you use immutable types (classes or structs) with value converters when possible. This is usually more efficient and has cleaner semantics than using a mutable type. However, that being said, it is common to use properties of types that the application cannot change. For example, mapping a property containing a list of numbers:
 
 [!code-csharp[ListProperty](../../../samples/core/Modeling/ValueConversions/MappingListProperty.cs?name=ListProperty)]
 
--- a/entity-framework/core/modeling/value-comparers.md
+++ b/entity-framework/core/modeling/value-comparers.md
@@ -2,7 +2,7 @@
 title: Value Comparers - EF Core
 description: Using value comparers to control how EF Core compares property values
 author: ajcvickers
-ms.date: 03/20/2020
+ms.date: 01/16/2021
 uid: core/modeling/value-comparers
 ---
 
@@ -16,9 +16,9 @@ uid: core/modeling/value-comparers
 
 ## Background
 
-Change tracking means that EF Core automatically determines what changes were performed by the application on a loaded entity instance, so that those changes can be saved back to the database when <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A> is called. EF Core usually performs this by taking a *snapshot* of the instance when it's loaded from the database, and *comparing* that snapshot to the instance handed out to the application.
+[Change tracking](xref:core/change-tracking/index) means that EF Core automatically determines what changes were performed by the application on a loaded entity instance, so that those changes can be saved back to the database when <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A> is called. EF Core usually performs this by taking a *snapshot* of the instance when it's loaded from the database, and *comparing* that snapshot to the instance handed out to the application.
 
-EF Core comes with built-in logic for snapshotting and comparing most standard types used in databases, so users don't usually need to worry about this topic. However, when a property is mapped through a [value converter](xref:core/modeling/value-conversions), EF Core needs to perform comparison on arbitrary user types, which may be complex. By default, EF Core uses the default equality comparison defined by types (e.g. the `Equals` method); for snapshotting, value types are copied to produce the snapshot, while for reference types no copying occurs, and the same instance is used as the snapshot.
+EF Core comes with built-in logic for snapshotting and comparing most standard types used in databases, so users don't usually need to worry about this topic. However, when a property is mapped through a [value converter](xref:core/modeling/value-conversions), EF Core needs to perform comparison on arbitrary user types, which may be complex. By default, EF Core uses the default equality comparison defined by types (e.g. the `Equals` method); for snapshotting, [value types](/dotnet/csharp/language-reference/builtin-types/value-types) are copied to produce the snapshot, while for [reference types](/dotnet/csharp/language-reference/keywords/reference-types) no copying occurs, and the same instance is used as the snapshot.
 
 In cases where the built-in comparison behavior isn't appropriate, users may provide a *value comparer*, which contains logic for snapshotting, comparing and calculating a hash code. For example, the following sets up value conversion for `List<int>` property to be value converted to a JSON string in the database, and defines an appropriate value comparer as well:
 
@@ -37,7 +37,7 @@ Consider byte arrays, which can be arbitrarily large. These could be compared:
 * By reference, such that a difference is only detected if a new byte array is used
 * By deep comparison, such that mutation of the bytes in the array is detected
 
-By default, EF Core uses the first of these approaches for non-key byte arrays. That is, only references are compared and a change is detected only when an existing byte array is replaced with a new one. This is a pragmatic decision that avoids copying entire arrays and comparing them byte-to-byte when executing <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A>, and the common scenario of replacing, say, one image with another is handled in a performant way.
+By default, EF Core uses the first of these approaches for non-key byte arrays. That is, only references are compared and a change is detected only when an existing byte array is replaced with a new one. This is a pragmatic decision that avoids copying entire arrays and comparing them byte-to-byte when executing <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A>. It means the common scenario of replacing, say, one image with another is handled in a performant way.
 
 On the other hand, reference equality would not work when byte arrays are used to represent binary keys, since it's very unlikely that an FK property is set to the _same instance_ as a PK property to which it needs to be compared. Therefore, EF Core uses deep comparisons for byte arrays acting as keys; this is unlikely to have a big performance hit since binary keys are usually short.
 
@@ -66,18 +66,13 @@ The mapping for simple structs is also simple and requires no special comparers 
 
 [!code-csharp[ConfigureImmutableStructProperty](../../../samples/core/Modeling/ValueConversions/MappingImmutableStructProperty.cs?name=ConfigureImmutableStructProperty)]
 
-EF Core has built-in support for generating compiled, memberwise comparisons of struct properties.
-This means structs don't need to have equality overridden for EF Core, but you may still choose to do this for [other reasons](/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type).
-Also, special snapshotting is not needed since structs are immutable and are always copied memberwise anyway.
-(This is also true for mutable structs, but [mutable structs should in general be avoided](/dotnet/csharp/write-safe-efficient-code).)
+EF Core has built-in support for generating compiled, memberwise comparisons of struct properties. This means structs don't need to have equality overridden for EF Core, but you may still choose to do this for [other reasons](/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type). Also, special snapshotting is not needed since structs are immutable and are always copied memberwise anyway. (This is also true for mutable structs, but [mutable structs should in general be avoided](/dotnet/csharp/write-safe-efficient-code).)
 
 ## Mutable classes
 
-It is recommended that you use immutable types (classes or structs) with value converters when possible.
-This is usually more efficient and has cleaner semantics than using a mutable type.
+It is recommended that you use immutable types (classes or structs) with value converters when possible. This is usually more efficient and has cleaner semantics than using a mutable type.
 
-However, that being said, it is common to use properties of types that the application cannot change.
-For example, mapping a property containing a list of numbers:
+However, that being said, it is common to use properties of types that the application cannot change. For example, mapping a property containing a list of numbers:
 
 [!code-csharp[ListProperty](../../../samples/core/Modeling/ValueConversions/MappingListProperty.cs?name=ListProperty)]
 
@@ -106,24 +101,16 @@ The <xref:Microsoft.EntityFrameworkCore.ChangeTracking.ValueComparer%601> constr
 
 In this case the comparison is done by checking if the sequences of numbers are the same.
 
-Likewise, the hash code is built from this same sequence.
-(Note that this is a hash code over mutable values and hence can [cause problems](https://ericlippert.com/2011/02/28/guidelines-and-rules-for-gethashcode/).
-Be immutable instead if you can.)
+Likewise, the hash code is built from this same sequence. (Note that this is a hash code over mutable values and hence can [cause problems](https://ericlippert.com/2011/02/28/guidelines-and-rules-for-gethashcode/). Be immutable instead if you can.)
 
-The snapshot is created by cloning the list with `ToList`.
-Again, this is only needed if the lists are going to be mutated.
-Be immutable instead if you can.
+The snapshot is created by cloning the list with `ToList`. Again, this is only needed if the lists are going to be mutated. Be immutable instead if you can.
 
 > [!NOTE]
-> Value converters and comparers are constructed using expressions rather than simple delegates.
-> This is because EF Core inserts these expressions into a much more complex expression tree that is then compiled into an entity shaper delegate.
-> Conceptually, this is similar to compiler inlining.
-> For example, a simple conversion may just be a compiled in cast, rather than a call to another method to do the conversion.
+> Value converters and comparers are constructed using expressions rather than simple delegates. This is because EF Core inserts these expressions into a much more complex expression tree that is then compiled into an entity shaper delegate. Conceptually, this is similar to compiler inlining. For example, a simple conversion may just be a compiled in cast, rather than a call to another method to do the conversion.
 
 ## Key comparers
 
-The background section covers why key comparisons may require special semantics.
-Make sure to create a comparer that is appropriate for keys when setting it on a primary, principal, or foreign key property.
+The background section covers why key comparisons may require special semantics. Make sure to create a comparer that is appropriate for keys when setting it on a primary, principal, or foreign key property.
 
 Use <xref:Microsoft.EntityFrameworkCore.MutablePropertyExtensions.SetKeyValueComparer%2A> in the rare cases where different semantics is required on the same property.
 
@@ -132,9 +119,7 @@ Use <xref:Microsoft.EntityFrameworkCore.MutablePropertyExtensions.SetKeyValueCom
 
 ## Overriding the default comparer
 
-Sometimes the default comparison used by EF Core may not be appropriate.
-For example, mutation of byte arrays is not, by default, detected in EF Core.
-This can be overridden by setting a different comparer on the property:
+Sometimes the default comparison used by EF Core may not be appropriate. For example, mutation of byte arrays is not, by default, detected in EF Core. This can be overridden by setting a different comparer on the property:
 
 [!code-csharp[OverrideComparer](../../../samples/core/Modeling/ValueConversions/OverridingByteArrayComparisons.cs?name=OverrideComparer)]
 
--- a/entity-framework/core/modeling/value-conversions.md
+++ b/entity-framework/core/modeling/value-conversions.md
@@ -42,7 +42,7 @@ Value conversions are configured in <xref:Microsoft.EntityFrameworkCore.DbContex
 -->
 [!code-csharp[BeastAndRider](../../../samples/core/Modeling/ValueConversions/EnumToStringConversions.cs?name=BeastAndRider)]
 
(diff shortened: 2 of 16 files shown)


entity-framework/core/modeling/value-comparers.md
```diff
@@ -37,7 +37,7 @@ Consider byte arrays, which can be arbitrarily large. These could be compared:
 * By reference, such that a difference is only detected if a new byte array is used
 * By deep comparison, such that mutation of the bytes in the array is detected
 
-By default, EF Core uses the first of these approaches for non-key byte arrays. That is, only references are compared and a change is detected only when an existing byte array is replaced with a new one. This is a pragmatic decision that avoids copying entire arrays and comparing them byte-to-byte when executing <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A>, and the common scenario of replacing, say, one image with another is handled in a performant way.
+By default, EF Core uses the first of these approaches for non-key byte arrays. That is, only references are compared and a change is detected only when an existing byte array is replaced with a new one. This is a pragmatic decision that avoids copying entire arrays and comparing them byte-to-byte when executing <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A>. It means the common scenario of replacing, say, one image with another is handled in a performant way.
```
roji, line 40:
```suggestion
By default, EF Core uses the first of these approaches for non-key byte arrays. That is, only references are compared and a change is detected only when an existing byte array is replaced with a new one. This is a pragmatic decision that avoids copying entire arrays and comparing them byte-to-byte when executing <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A>. It means that the common scenario of replacing, say, one image with another is handled in a performant way.
```

entity-framework/core/modeling/value-comparers.md
```diff
@@ -66,18 +66,13 @@ The mapping for simple structs is also simple and requires no special comparers
...
 
 ## Mutable classes
 
-It is recommended that you use immutable types (classes or structs) with value converters when possible.
-This is usually more efficient and has cleaner semantics than using a mutable type.
+It is recommended that you use immutable types (classes or structs) with value converters when possible. This is usually more efficient and has cleaner semantics than using a mutable type.
 
-However, that being said, it is common to use properties of types that the application cannot change.
-For example, mapping a property containing a list of numbers:
+However, that being said, it is common to use properties of types that the application cannot change. For example, mapping a property containing a list of numbers:
```
roji, line 75: Consider combining with previous sentence into a single short paragraph

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+        {
+            Donkey,
+            Mule,
+            Horse,
+            Unicorn
+        }
+-->
+[!code-csharp[BeastAndRider](../../../samples/core/Modeling/ValueConversions/EnumToStringConversions.cs?name=BeastAndRider)]
+
+Conversions can be configured in <xref:Microsoft.EntityFrameworkCore.DbContext.OnModelCreating%2A> to store the enum values as strings such as "Donkey", "Mule", etc. in the database:
```
roji, line 45:
```suggestion
Conversions can be configured in <xref:Microsoft.EntityFrameworkCore.DbContext.OnModelCreating%2A> to store the enum values as strings such as "Donkey", "Mule", etc. in the database; you simply need to provide one function which converts from the `ModelClrType` to the `ProviderClrType`, and another for the opposite conversion:
```

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+            {
+                modelBuilder
+                    .Entity<Rider>()
+                    .Property(e => e.Mount)
+                    .HasConversion<string>();
+            }
+-->
+[!code-csharp[ConversionByClrType](../../../samples/core/Modeling/ValueConversions/EnumToStringConversions.cs?name=ConversionByClrType)]
+
+The same thing can be achieved by explicitly specifying the database column type. For example, if the entity type is defined like so:
```
roji, line 80: Provide both the Fluent API and the Data Annotations alternatives via tabs, like elsewhere?

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+        {
+            public int Id { get; set; }
+
+            [Column(TypeName = "nvarchar(24)")]
+            public EquineBeast Mount { get; set; }
+        }
+-->
+[!code-csharp[ConversionByDatabaseType](../../../samples/core/Modeling/ValueConversions/EnumToStringConversions.cs?name=ConversionByDatabaseType)]
+
+Then the enum values will be saved as strings in the database without any further configuration in OnModelCreating.
```
roji, line 93:
```suggestion
Then the enum values will be saved as strings in the database without any further configuration in <xref:Microsoft.EntityFrameworkCore.DbContext.OnModelCreating%2A>.
```

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+
+                modelBuilder
+                    .Entity<User>()
+                    .Property(e => e.IsActive)
+                    .HasConversion(converter);
+            }
+-->
+[!code-csharp[ConversionByBuiltInBoolToIntExplicit](../../../samples/core/Modeling/ValueConversions/EnumToStringConversions.cs?name=ConversionByBuiltInBoolToIntExplicit)]
+
+The following table summarizes commonly used pre-defined conversions from model/property types to database provider types.
```
roji, line 146:
```suggestion
The following table summarizes commonly-used pre-defined conversions from model/property types to database provider types.
```

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+
+In the table `any_numeric_type` means one of `int`, `short`, `long`, `byte`, `uint`, `ushort`, `ulong`, `sbyte`, `char`, `decimal`, `float`, or `double`.
```
roji, line 148: Considering merging with previous sentence into a single paragraph.

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+                    .Property(e => e.Price)
+                    .HasConversion(
+                        v => JsonSerializer.Serialize(v, null),
+                        v => JsonSerializer.Deserialize<Money>(v, null));
+-->
+[!code-csharp[ConfigureCompositeValueObject](../../../samples/core/Modeling/ValueConversions/CompositeValueObject.cs?name=ConfigureCompositeValueObject)]
 
-## Built-in converters
+> [!NOTE]
+> We plan to allow mapping an object to multiple columns in EF Core 6.0. This will remove the need to use serialization here. This is tacked by [GitHub issue #13947](https://github.com/dotnet/efcore/issues/13947).
```
roji, line 340:
```suggestion
> We plan to allow mapping an object to multiple columns in EF Core 6.0, removing the need to use serialization here. This is tracked by [GitHub issue #13947](https://github.com/dotnet/efcore/issues/13947).
```

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+[!code-csharp[ConfigureValueObjectCollection](../../../samples/core/Modeling/ValueConversions/ValueObjectCollection.cs?name=ConfigureValueObjectCollection)]
 
-```csharp
-public class Rider
-{
-    public int Id { get; set; }
+> [!NOTE]
+> As before, this conversion requires a <xref:Microsoft.EntityFrameworkCore.ChangeTracking.ValueComparer%601>. See [Value Comparers](xref:core/modeling/value-comparers) for more information.
+
+### Value objects as keys
```
roji, line 474: Do we really want to show this, given that I'm not sure we like it? :smiling_imp:

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+                        new ValueComparer<ICollection<string>>(
+                            (c1, c2) => c1.SequenceEqual(c2),
+                            c => c.Aggregate(0, (a, v) => HashCode.Combine(a, v.GetHashCode())),
+                            c => (ICollection<string>)c.ToList()));
+-->
+[!code-csharp[ConfigurePrimitiveCollection](../../../samples/core/Modeling/ValueConversions/PrimitiveCollection.cs?name=ConfigurePrimitiveCollection)]
+
+`ICollection<string>` represents a mutable reference type. This means that a <xref:Microsoft.EntityFrameworkCore.ChangeTracking.ValueComparer%601> is needed so that EF Core can track and detect changes correctly. See [Value Comparers](xref:core/modeling/value-comparers) for more information.
+
+### Collections of value objects
```
roji, line 393: This specific example seems like it's giving less new info than the others (just showing another example with principles that were already shown above)... Concision is also a virtue (but am also OK with leaving as-is).

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+            }
+-->
+[!code-csharp[ConfigureKeyValueObjects](../../../samples/core/Modeling/ValueConversions/KeyValueObjects.cs?name=ConfigureKeyValueObjects)]
+
+> [!NOTE]
+> Currently key properties with conversions cannot use generated key values. Vote for [GitHub issue #11597](https://github.com/dotnet/efcore/issues/11597) to have this limitation removed.
+
+### Use ulong for timestamp/rowversion
+
+SQL Server supports automatic [optimistic concurrency](xref:core/saving/concurrency) using [8-byte binary rowversion/timestamp columns](/sql/t-sql/data-types/rowversion-transact-sql). These are always read from and written to the database using an 8-byte array. However, byte arrays are a mutable reference type, which makes them someone painful to deal with. Value converters allow the rowversion to instead be mapped to a `ulong` property, which is much more appropriate and easy to use than the byte array. For example, consider a `Blog` entity with a ulong concurrency token:
```
roji, line 548:
```suggestion
SQL Server supports automatic [optimistic concurrency](xref:core/saving/concurrency) using [8-byte binary `rowversion`/`timestamp` columns](/sql/t-sql/data-types/rowversion-transact-sql). These are always read from and written to the database using an 8-byte array. However, byte arrays are a mutable reference type, which makes them somewhat painful to deal with. Value converters allow the rowversion to instead be mapped to a `ulong` property, which is much more appropriate and easy to use than the byte array. For example, consider a `Blog` entity with a ulong concurrency token:
```

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+        public class Blog
+        {
+            public int Id { get; set; }
+            public string Name { get; set; }
+            public ulong Version { get; set; }
+        }
+-->
+[!code-csharp[ULongConcurrencyModel](../../../samples/core/Modeling/ValueConversions/ULongConcurrency.cs?name=ULongConcurrencyModel)]
+
+This can be mapped to a SQL server rowversion column using a value converter:
```
roji, line 560:
```suggestion
This can be mapped to a SQL server `rowversion` column using a value converter:
```

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+                modelBuilder.Entity<Blog>()
+                    .Property(e => e.Version)
+                    .IsRowVersion()
+                    .HasConversion<byte[]>();
+-->
+[!code-csharp[ConfigureULongConcurrency](../../../samples/core/Modeling/ValueConversions/ULongConcurrency.cs?name=ConfigureULongConcurrency)]
+
+### Specify the DateTime.Kind when reading dates
+
+SQL Server discards the <xref:System.DateTime.Kind%2A?displayProperty=nameWithType> flag when storing a <xref:System.DateTime> as a [datetime](/sql/t-sql/data-types/datetime-transact-sql) or [datetime2](/sql/t-sql/data-types/datetime2-transact-sql). This means that DateTime values coming back from the database always have <xref:System.DateTimeKind> of `Unspecified`.
```
roji, line 572:
```suggestion
SQL Server discards the <xref:System.DateTime.Kind%2A?displayProperty=nameWithType> flag when storing a <xref:System.DateTime> as a [`datetime`](/sql/t-sql/data-types/datetime-transact-sql) or [`datetime2`](/sql/t-sql/data-types/datetime2-transact-sql). This means that DateTime values coming back from the database always have <xref:System.DateTimeKind> of `Unspecified`.
```

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+<!--
+                modelBuilder.Entity<Post>()
+                    .Property(e => e.PostedOn)
+                    .HasConversion<long>();
+-->
+[!code-csharp[ConfigurePreserveDateTimeKind1](../../../samples/core/Modeling/ValueConversions/PreserveDateTimeKind.cs?name=ConfigurePreserveDateTimeKind1)]
+
+This allows DateTime values with different `Kind` flags to be mixed in the database.
+
+The problem with this approach is that the database no longer has recognizable `datetime` or `datetime2` columns. So instead it is common to always store UTC time (or, less commonly, always local time) and then either ignore the `Kind` flag or set it to the appropriate value using a value converter. For example, the converter below ensures that the `DateTime` value read from the database will have the <xref:System.DateTimeKind> `UTC` kind:
```
roji, line 585: Consider merging with previous sentence.

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+
+<!--
+                modelBuilder.Entity<Post>()
```
roji, line 588: I understand this example is only about correcting the Kind for values coming back from the database... But should we also show conversion to UTC when *sending* to the database (e.g. in case a DateTime with Kind=Local is assigned to the entity property)? This would mean calling [DateTime.ToUniversalTime](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.touniversaltime?view=net-5.0#System_DateTime_ToUniversalTime) in the model->provider conversion side.

Am thinking about unsuspecting users copy-pasting this and thinking they have a full, UTC-everywhere solution.

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+                    .Property(e => e.LastUpdated)
+                    .HasConversion(
+                        v => v,
+                        v => new DateTime(v.Ticks, DateTimeKind.Utc));
+-->
+[!code-csharp[ConfigurePreserveDateTimeKind2](../../../samples/core/Modeling/ValueConversions/PreserveDateTimeKind.cs?name=ConfigurePreserveDateTimeKind2)]
+
+### Use case-insensitive string keys
+
+Some databases, including SQL Server, perform case-insensitive string comparisons by default. .NET on the other hand performs case-sensitive string comparisons by default. This means that a foreign key value like "DotNet" will match the primary key value "dotnet" on SQL Server, but will not match it in EF Core. A value comparer for keys can be used for force EF Core into case-insensitive string comparisons like in the database. For example, consider a blog/posts model with string keys:
```
roji, line 598:
```suggestion
Some databases, including SQL Server, perform case-insensitive string comparisons by default. .NET, on the other hand, performs case-sensitive string comparisons by default. This means that a foreign key value like "DotNet" will match the primary key value "dotnet" on SQL Server, but will not match it in EF Core. A value comparer for keys can be used for force EF Core into case-insensitive string comparisons like in the database. For example, consider a blog/posts model with string keys:
```

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+            public string Title { get; set; }
+            public string Content { get; set; }
+
+            public string BlogId { get; set; }
+            public Blog Blog { get; set; }
+        }
+-->
+[!code-csharp[CaseInsensitiveStringsModel](../../../samples/core/Modeling/ValueConversions/CaseInsensitiveStrings.cs?name=CaseInsensitiveStringsModel)]
+
+This will not work as expected if some of the `Post.BlogId` values have different casing. A value comparer can be used to correct this:
```
roji, line 621: It may be good to add a sentence detailing exactly how this will fail if the comparer isn't defined, to make users better understand why this is important.

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+            }
+-->
+[!code-csharp[ConfigureCaseInsensitiveStrings](../../../samples/core/Modeling/ValueConversions/CaseInsensitiveStrings.cs?name=ConfigureCaseInsensitiveStrings)]
+
+> [!NOTE]
+> .NET string comparisons and database string comparisons can differ in more than just case sensitivity. This pattern works for simple ASCII keys, but may fail for keys with any kind of culture-specific characters. See [Collations and Case Sensitivity](xref:core/miscellaneous/collations-and-case-sensitivity) for more information.
+
+### Handle fixed-length database strings
+
+The previous example did not need a value converter. However, a converter can be useful for fixed-length database string types like `char(20)` or `nchar(20)`. Fixed-length strings are padded to their full length whenever a value is inserted into the database. This means that a key value of "dotnet" will be read back from the database as "dotnet              ". This will then not compare correctly with key values that are not padded.
```
roji, line 650:
```suggestion
The previous example did not need a value converter. However, a converter can be useful for fixed-length database string types like `char(20)` or `nchar(20)`. Fixed-length strings are padded to their full length whenever a value is inserted into the database. This means that a key value of `dotnet` will be read back from the database as `dotnet              `. This will then not compare correctly with key values that are not padded.
```

entity-framework/core/modeling/value-conversions.md
```diff
@@ -2,145 +2,746 @@
...
+-->
+[!code-csharp[ConfigureEncryptPropertyValues](../../../samples/core/Modeling/ValueConversions/EncryptPropertyValues.cs?name=ConfigureEncryptPropertyValues)]
+
+> [!NOTE]
+> There is currently no way to get a reference to the current DbContext, or other session state, from within a value converter. This limits the kinds of encryption that can be used. Vote for [GitHub issue #11597](https://github.com/dotnet/efcore/issues/12205) to have this limitation removed.
+
+> [!WARNING]
+> Make sure to understand all the implications if you roll your own encryption to protect sensitive data. Consider instead using pre-built encryption mechanisms, such as [Always Encrypted](/sql/relational-databases/security/encryption/always-encrypted-database-engine) on SQL Server.
+
+### Column facets and mapping hints
```
roji, line 698: This seems less like an example, and more like it should be part of the main documentation somewhere above.

roji (approved):
Awesome, love the examples.

entity-framework/core/modeling/value-conversions.md
ajcvickers: Yes, because so many people ask about this, and many of them get it wrong.

entity-framework/core/modeling/value-conversions.md
ajcvickers: I wanted to give an example which was a bit less trivial than the others and used nested types. The new concept here is that value objects can be easily composed. This doesn't require making them owned types, for example.

entity-framework/core/modeling/value-conversions.md
ajcvickers: Interesting idea; I'll add something.

entity-framework/core/modeling/value-conversions.md
ajcvickers: Are fences really appropriate here? These are actually string values, not, not language concepts or typpes.

entity-framework/core/modeling/value-conversions.md
ajcvickers: Agreed. WiIll move.

entity-framework/core/modeling/value-conversions.md
roji: That's true, but (a) the spaces don't really show, and (b) "dotnet" may get machine-translated (to who knows what)...

ajcvickers: Turns out we are not allowed trailing spaces in fences, so leaving as is.

roji: Oh well...

entity-framework/core/modeling/value-conversions.md
roji: Make sense!

Issue #502, opened 2017-10-06
New feature topic: explicitly compled query


divega:
Originally form #447.

roji:
We have [this section](https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-2.0#explicitly-compiled-queries) in the 2.0 release notes which we can use.

ajcvickers:
Also add a note about this limitation: https://github.com/dotnet/efcore/issues/13483

roji:
Rewrite the custom benchmark under samples/core/CompiledQueries/Program.cs as a proper BDN benchmark.

roji:
Since this has been undocumented for a very long time, and we have no concrete plans to redo the API surface in the near future, I'm submitting a documentation proposal in #3502.

PR #3022, opened 2021-01-17
Samples code cleanup and verification that all samples run

### LINKED ISSUE #2935: Validate that all samples run
There are multiple runtime exceptions I'm experiencing.

An example is the Saving example [here](https://github.com/dotnet/EntityFramework.Docs/blob/2667d7425ab1043f5599c182c75b9112389b46c3/samples/core/Saving/Basics/Sample.cs#L62):

`Queries performing 'Last' operation must have a deterministic sort order. Rewrite the query to apply an 'OrderBy' clause on the sequence before calling 'Last'.'`

I noticed that the the github actions [yml](https://github.com/dotnet/EntityFramework.Docs/blob/2667d7425ab1043f5599c182c75b9112389b46c3/.github/workflows/build-samples.yml#L35) doesn't run the samples, so which makes it difficult to determine if the problem is in my environment or not.

Are there plans to add `dotnet run` to CI/CD and if not would you take a PR for that?

ajcvickers:
Fixes #2935


samples/core/CompiledQueries/Program.cs
```diff
@@ -10,6 +10,21 @@ public class Program
     {
         private static void Main()
         {
+            using (var db = new AdventureWorksContext())
```
roji, line 13: We should convert this to a proper BDN benchmark, added a note in #502.

roji (approved):
Should also bring in the EF Core .editorconfig

ajcvickers:
Added editor.config. (Good catch; this results in fewer changes in indentation.)

PR #3025, opened 2021-01-19
fix grammar: add definite/ indefinite articles

Geoff1900:
The definite (the) and indefinite (a) articles are sometimes missing.
It is inconsistent so does not appear to be a conscious writing style.
I propose to add them, where missing, to improve comprehension.


roji:
@Geoff1900 I fixed up one sentence and merged, thanks for your contribution!

PR #3033, opened 2021-01-21
Filter out API members with [EntityFrameworkInternal]

### LINKED ISSUE #3029: Exclude APIs members marked with [EntityFrameworkCoreInternal]
Public docfx allows [filtering out APIs marked with a specific attribute](https://dotnet.github.io/docfx/tutorial/howto_filter_out_unwanted_apis_attributes.html#3-exclude-or-include-apis-by-containing-matched-attributes). We can do this to keep out members annotated with [EntityFrameworkCoreInternal].

roji:
Closes #3029

@smitpatel any way to validate/test this?

### DIFF
--- a/entity-framework/docfx.json
+++ b/entity-framework/docfx.json
@@ -13,7 +13,8 @@
           ]
         }
       ],
-      "dest": "api"
+      "dest": "api",
+      "filter": "filterConfig.yml"
     }
   ],
   "build": {
--- a/entity-framework/filterConfig.yml
+++ b/entity-framework/filterConfig.yml
@@ -1,0 +1,83 @@
+apiRules:
+- exclude:
+    hasAttribute:
+      uid: Microsoft.EntityFrameworkCore.Infrastructure.EntityFrameworkInternalAttribute
+# From here on, this is the default filter config as in https://dotnet.github.io/docfx/tutorial/howto_filter_out_unwanted_apis_attributes.html#3-default-filter-configuration
+- exclude:
+    hasAttribute:
+      uid: System.ComponentModel.EditorBrowsableAttribute
+      ctorArguments:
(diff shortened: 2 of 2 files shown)


roji:
Also:

image

roji:
Doesn't seem to work, though I'm seeing examples of this for public docfx...

smitpatel:
This is incorrect repository.

roji:
Ah, OK - though note that there is a metadata section in the docfx referencing sources, maybe that should be cleaned up.

In any case, do you know where this should go?

smitpatel:
Since there is no src in this repo, that section is actually no-op.
Is this build time filtering or render time? i.e. Does this filtering apply to ECMAXML or the generation of it?

roji:
> Since there is no src in this repo, that section is actually no-op.

Yeah, I can change this PR to simply remove it.

>  Is this build time filtering or render time? i.e. Does this filtering apply to ECMAXML or the generation of it?

image

smitpatel:
Can you link me to the doc of filtering in public docfx build? I will ask the team which is generating our API docs for us.

roji:
Here you go: https://dotnet.github.io/docfx/tutorial/howto_filter_out_unwanted_apis_attributes.html#3-exclude-or-include-apis-by-containing-matched-attributes

roji:
@smitpatel if you're looking at this with the docs people, you may want to also mention that the Microsoft.EntityFrameworkCore.Metadata.Builders.Internal namespace does seem to show up in our API docs, though Internals are supposed to be filtered out...

(thanks for managing this stuff!)

smitpatel:
We manage the filtering. I updated our filters. Website should be updated in some time.

roji:
@smitpatel thanks, closing this. #3029 tracks filtering out the attribute as well.

PR #3042, opened 2021-01-26
[6.0-preview1] Add new translations

bricelam:


entity-framework/core/providers/cosmos/functions.md
```diff
@@ -12,6 +12,7 @@ This page shows which .NET members are translated into which SQL functions when
 .NET                          | SQL                              | Added in
 ----------------------------- | -------------------------------- | --------
 collection.Contains(item)     | @item IN @collection
+EF.Functions.Random()         | RAND()                           | EF Core 6.0
```
bricelam, line 15: @ajcvickers Is just `EF Core 6.0` good or should we mention something about a preview or prerelease?

roji (approved):
EF Core 6.0 seems sufficient to me personally.

entity-framework/core/providers/cosmos/functions.md
ajcvickers: The only case where this seems useful is when we haven't released the preview yet, but the docs are updated already. But this may happen multiple times through the release, so my slight preference is to include the preview for now and then scrub these after release. But I don't feel strongly.

PR #3046, opened 2021-01-28
Fix small typo

garyng:
Small changes like this is fine right?


roji (approved):
Yep - these are always welcome, thanks!

PR #3058, opened 2021-01-31
Updated dotnet ef cli command

timonsk:
According to new changes fo dotnet ef cli new command should use to match new cli requirement, so changed from dotnet ef to dotnet-ef


ajcvickers:
@timonsk Can you provide some more details on why this is needed? Do you have a reference for what these "new changes" are?

timonsk:
@ajcvickers looks like this https://docs.microsoft.com/en-us/ef/core/cli/dotnet, where dotnet-ef is as a separate tool.

ajcvickers:
@timonsk The underlying binary is `dotnet-ef`, but once the command is installed, either globally or as a project tool, then it is accessed through the dotnet runtime with `dotnet ef`. This is not something that has changed.

timonsk:
okay, closing the pr then. thanks!

PR #3119, opened 2021-02-26
Update dotnet.md

Rick-Anderson:
This took a long time to find when using `Down`.

Per MS style guide avoid parenthetical expressions (they're hard to read in English (and impossible (nearly) to MT (Machine Translate))) :)

### DIFF
--- a/entity-framework/core/cli/dotnet.md
+++ b/entity-framework/core/cli/dotnet.md
@@ -274,7 +274,7 @@ Options:
 
 | Option                 | Short             | Description                                                                     |
 |:-----------------------|:------------------|:--------------------------------------------------------------------------------|
-| <nobr>`--force`</nobr> | <nobr>`-f`</nobr> | Revert the latest migration, that is, rolls back the changes that were applied to the database on the last migration. |
+| <nobr>`--force`</nobr> | <nobr>`-f`</nobr> | Revert the latest migration, rolling back the changes that were applied to the database in the last migration. |
 
 The [common options](#common-options) are listed above.
 
--- a/entity-framework/core/cli/dotnet.md
+++ b/entity-framework/core/cli/dotnet.md
@@ -268,13 +268,13 @@ The [common options](#common-options) are listed above.
 
 ## `dotnet ef migrations remove`
 
-Removes the last migration (rolls back the code changes that were done for the migration).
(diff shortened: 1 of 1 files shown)


entity-framework/core/cli/dotnet.md
```diff
@@ -268,13 +268,13 @@ The [common options](#common-options) are listed above.
...
 
-Removes the last migration (rolls back the code changes that were done for the migration).
+Removes the last migration by calling the `Down` method of the latest migration. The command rolls back the code changes that were done for the latest migration.
 
 Options:
 
 | Option                 | Short             | Description                                                                     |
 |:-----------------------|:------------------|:--------------------------------------------------------------------------------|
-| <nobr>`--force`</nobr> | <nobr>`-f`</nobr> | Revert the migration (roll back the changes that were applied to the database). |
+| <nobr>`--force`</nobr> | <nobr>`-f`</nobr> | Revert the latest migration, that is, rolls back the changes that were applied to the database on the last migration. |
```
roji, line 277: How about:

```suggestion
| <nobr>`--force`</nobr> | <nobr>`-f`</nobr> | Revert the latest migration, rolling back the changes that were applied to the database in the last migration. |
```

roji (approved):

ajcvickers:
@bricelam Is this change correct? Removing a migration doesn't cause `Down` to be called, does it?

bricelam:
This part isn't correct:

> Removes...by calling the `Down` method

We only run the operations described in the `Down` method if the user tells us to. (By confirmation or by specifying `--force`) And, it's not related to rolling back the code changes.

roji:
Sorry, I was focusing on the language here and missed this.

Submitted #3156 to fix this, @bricelam is that correct?

FWIW I know we've discussed this before, though I still find it odd for `--force` to do something different from what the command does without it (with it we touch the database, without it we don't). From my Unix mindset, `--force` is supposed to do the exact same thing, just to force the tool to complete silently without aborting on risky situations. Hopefully I'm not still misunderstanding.

bricelam:
`--force` will skip checking the database. This might in turn skip reverting the migration. But we also warn during `--force` that you might have to manually revert the database if the migration was applied. In other words, `--force` avoids some built-in guardrails that might be preventing you from removing the C# code for the migrations.

PR #3122, opened 2021-02-28
Firebird is updated to EFCore 5.0

mariuz:
https://twitter.com/cincura_net/status/1364886579353972740

https://www.nuget.org/packages/FirebirdSql.EntityFrameworkCore.Firebird/8.0.0-alpha3


roji:
@mariuz the nuget you're pointing at is an alpha prerelease... We should probably wait until there's a non-preview release.

mariuz:
@roji Ok then we can wait until then , all looks green in the testing area 

https://github.com/FirebirdSQL/NETProvider/runs/1977984662

ajcvickers:
@cincuranet This change advertises 5.0 of your provider. We don't usually do this until the 5.0 version is released. If you wish to have this updated before then, then it probably should indicate that the 5.0 version is prerelease. What would you prefer?

cincuranet:
I'm finishing few other pieces and will release final version in a week or two. Therefore I would prefer to wait and update when final version is out.

ajcvickers:
Closing this for now per @cincuranet's response.

PR #3132, opened 2021-03-05
Update BackingField.cs

mrlife:
Added missing private field for the example at https://docs.microsoft.com/en-us/ef/core/modeling/backing-field?tabs=data-annotations#basic-configuration

[Additional ref](https://www.learnentityframeworkcore5.com/whats-new-in-ef-core-5/backing-fields)

### DIFF
--- a/samples/core/Modeling/Conventions/BackingField.cs
+++ b/samples/core/Modeling/Conventions/BackingField.cs
@@ -14,7 +14,11 @@ namespace EFModeling.Conventions.BackingField
         
         public int BlogId { get; set; }
 
-        public string Url { get; set; }
+        public string Url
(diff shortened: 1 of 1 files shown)


samples/core/Modeling/Conventions/BackingField.cs
```diff
@@ -10,6 +10,8 @@ internal class MyContext : DbContext
     #region Sample
     public class Blog
     {
+        private string _url;
```
AndriySvyryd, line 13: Also use it in `Url` implementation

ajcvickers (approved):

ajcvickers:
Thanks!

PR #3214, opened 2021-04-19
docs: fix typo in tracking.md

nelsonprsousa:


roji (commented):
We consistently refer to "tracking queries", i.e. queries which track the result they return. It isn't the queries themselves which are tracked, as your corrections implies.

nelsonprsousa:
> We consistently refer to "tracking queries", i.e. queries which track the result they return. It isn't the queries themselves which are tracked, as your corrections implies.

Oh, it felt wrong but I think I got it.

Closing...

PR #3225, opened 2021-04-28
Update whatsnew.md to fix typo for SQL DATALENGTH function name

alexernest:
Fix a typo for the name of the SQL DATALENGTH function, changing from DATELENGTH


roji (approved):
Thanks!

PR #3244, opened 2021-05-11
IDbContextFactory<TContext> Instead

wuyuansushen:
IDbContextFactory is used in EF or early version of EF Core.
Please use IDesignTimeDbContextFactory<TContext> Interface to instead of  it.
[ref](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.infrastructure.idbcontextfactory-1?view=efcore-2.2)


wuyuansushen:
I have discovery it, sorry. @ajcvickers

PR #3256, opened 2021-05-23
Update keyless-entity-types.md: Note about ToView()

HSchwichtenberg:
I think the sentence "`ToView` assumes that the object already exists in the database and it won't be created by migrations." is old / misleading.


ajcvickers (approved):

ajcvickers:
@AndriySvyryd Does this change look good to you?

AndriySvyryd (approved):

ajcvickers:
Thanks!

PR #3311, opened 2021-07-01
Update plan.md

bricelam:


AndriySvyryd (approved):
Also mark these as in progress:
* Cosmos database provider
* Enhancements to System.Data

PR #3313, opened 2021-07-06
Removed extra parenthesis

JohnMHigginsPMC:
There was an extra left parenthesis with no matching right one, so I removed it.


AndriySvyryd (approved):
Thanks!

PR #3321, opened 2021-07-09
Add a note on querying shared columns

joakimriedel:
From https://github.com/dotnet/efcore/issues/25210, suggestion to add a note in the section on shared columns that some providers do not filter results by discriminator when casting a base entity to a child entity.

ping @AndriySvyryd

### DIFF
--- a/entity-framework/core/modeling/inheritance.md
+++ b/entity-framework/core/modeling/inheritance.md
@@ -56,7 +56,7 @@ By default, when two sibling entity types in the hierarchy have a property with 
 [!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/SharedTPHColumns.cs?name=SharedTPHColumns&highlight=9,13)]
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


entity-framework/core/modeling/inheritance.md
```diff
@@ -55,6 +55,9 @@ By default, when two sibling entity types in the hierarchy have a property with
 
 [!code-csharp[Main](../../../samples/core/Modeling/FluentAPI/SharedTPHColumns.cs?name=SharedTPHColumns&highlight=9,13)]
 
+> [!NOTE]
+> Some database providers, such as SQL Server, will not automatically use the discriminator predicate when querying shared columns using a cast. The query `Url = (blob as RssBlog).Url` would also return the `Url` value for the sibling `Blog` rows. To restrict the query to `RssBlog` entities in the hierarchy you need to add a filter manually on the discriminator, such as `Url = blob is RssBlog ? (blob as RssBlog).Url : null`.
```
AndriySvyryd, line 59:
```suggestion
> Relational database providers, such as SQL Server, will not automatically use the discriminator predicate when querying shared columns when using a cast. The query `Url = (blob as RssBlog).Url` would also return the `Url` value for the sibling `Blog` rows. To restrict the query to `RssBlog` entities you need to manually add a filter on the discriminator, such as `Url = blob is RssBlog ? (blob as RssBlog).Url : null`.
```

AndriySvyryd (approved):
Thanks!

PR #3326, opened 2021-07-14
Update index.md

OOberoi:
Just a handful of missing punctuation marks.

### DIFF
--- a/entity-framework/core/change-tracking/index.md
+++ b/entity-framework/core/change-tracking/index.md
@@ -32,7 +32,7 @@ Entity instances are no longer tracked when:
 - The change tracker is cleared (EF Core 5.0 and later)
 - The entities are explicitly detached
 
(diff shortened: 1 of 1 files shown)


roji (commented):
This doesn't seem like idiomatic English to me ([see e.g. this](https://jakubmarian.com/comma-before-that-and-which/)) - in any case it's not the style we've used in our docs so far.

OOberoi:
Sounds good!

Thanks,

Obi Oberoi – (M.Sc.)
Cell: (647) 298 7130

smitpatel:
I ran through the tool and tool suggest that comma shouldn't be used (along with the fact that this should not be used without noun too). Closing this PR as the change doesn't align with doc writing guidelines.

OOberoi:
Thanks for the input!

PR #3331, opened 2021-07-14
Document IScaffoldingModelFactory

bricelam:


smitpatel (approved):
While this work, the link currently points to 1.1 version of this. It will point to correct one, once the API docs for 6.0 are updated.

bricelam:
lol, I was sure it wouldn't work at all. Got lucky.

PR #3339, opened 2021-07-20
Update value-conversions.md

mrlife:
False values are listed first in most of the document but not all places. The first edit also fixes implied mapping of false to "Y" and true to "N".


ajcvickers (approved):

ajcvickers:
Thanks!

PR #3354, opened 2021-07-30
Remove EF Core version 2 mentions

### LINKED ISSUE #2922: Proposal: Remove tools and extensions that only support EF Core 2.
Too remove stale/obsolete extensions and tools, thus making the list more valuable.

ErikEJ:
Remove mentions of version 2, which is no longer supported (by August 21)

fixes #2922

### DIFF
--- a/entity-framework/core/extensions/index.md
+++ b/entity-framework/core/extensions/index.md
@@ -17,13 +17,13 @@ These tools and extensions provide additional functionality for Entity Framework
 
 ### LLBLGen Pro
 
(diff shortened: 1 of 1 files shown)


ajcvickers:
@ErikEJ We discussed this and decided to leave this page as is. We know that people still use unsupported versions and there doesn't seem much harm in leaving this in the docs for people who are doing that. However, we agreed that we should update the other docs at some point in the future to remove 2.x references, but not utile 2.1 has been out of support for some considerable time.

PR #3364, opened 2021-08-01
Remove duplicate note

ashotmuradian:
Remove the duplicate mention of using EF Core tools when the `DbContext` is in a different assembly.
The same note is already at the beginning of the page


ashotmuradian:
The first occurrence of this duplicate note is at the line 13

ajcvickers (approved):

ajcvickers:
Thanks!

PR #3371, opened 2021-08-08
Update ManyToManyShared.cs with fix for PostTag entity

cwoodruff:
There are two Post entity references in the modelBuilder in the OnModelCreating method and the second one should be reference PostTag since it has both the PostId and TagId references.


cwoodruff:
you can refuse this PR as it is wrong.

smitpatel:
This is incorrect change. It should be post only. This example specifically talks about when the join entity is created by EF Core internally and there is no concrete user class for it. Hence to seed data it accesses the join entity through UsingEntity API after HasMany/WithMany.

PR #3409, opened 2021-08-31
Fix NuGet packages link

rbgnk:
The link for "Spatial support for SQL Server" was pointing to the SQLite package


roji (approved):
Thanks!

PR #3414, opened 2021-09-02
Mention about --startup-project param

EltonAlvess:
Is important to mention that should be inside data folder and need to use --startup-project to run migrations. I saw a lot of questions opened in stack overflow about it. Actually i spend a lot of time trying to discover a solution until tried to do like i mentioned before.


ajcvickers:
@EltonAlvess Thanks for pointing this out. We're not going to take a PR in this form, but we have filed #3418 to follow up in this area.

PR #3415, opened 2021-09-02
Inherit from Controller class

vanillajonathan:


ajcvickers:
@vanillajonathan This is not a genuine ASP.NET controller, but rather almost pseudo-code showing the context. This could perhaps be made clearer, but I don't think adding the base class does this.

vanillajonathan:
I thought it would make it more obvious that it is a ASP.NET Controller.

ajcvickers (approved):

ajcvickers:
@vanillajonathan Fair enough. Thanks!

PR #3419, opened 2021-09-05
update IBM provider to 5.0

vslee:
- package is now available on NuGet so updated link to that
- even though it is on NuGet, it appears to still require a license (https://stackoverflow.com/a/63612855/2847567) so kept the `Paid` note
- linked the "Getting Started" page as that is more useful
- lower cased the `T` from the Google Spanner link to be consistent with the rest of the column


roji (approved):
Thanks!

PR #3392, opened 2021-08-21
End of life 2.1

bricelam:
I took a pretty extreme approach of removing all irrelevant information about unsupported versions from the main docs. Mostly because I hope that everyone has already upgraded to a supported version and isn't still actively developing on these older versions.

Please push back on any information that you still think is still relevant to readers.

PR #3433, opened 2021-09-15
Updated linq2db.EntityFrameworkCore section with support EF Core 2

sdanyliv:
Just released version which supports EF Core 2.x

### DIFF
--- a/entity-framework/core/extensions/index.md
+++ b/entity-framework/core/extensions/index.md
@@ -236,7 +236,7 @@ Adds hierarchyid support to the SQL Server EF Core provider. For EF Core: 3, 5.
 
 ### linq2db.EntityFrameworkCore
 
-Alternative translator of LINQ queries to SQL expressions. For EF Core: 3, 5, 6.
+Alternative translator of LINQ queries to SQL expressions. For EF Core: 2, 3, 5, 6.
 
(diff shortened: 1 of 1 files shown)


entity-framework/core/extensions/index.md
```diff
@@ -236,7 +236,7 @@ Adds hierarchyid support to the SQL Server EF Core provider. For EF Core: 3, 5.
 
 ### linq2db.EntityFrameworkCore
 
-Alternative translator of LINQ queries to SQL expressions. For EF Core: 3, 5, 6.
+Alternative translator of LINQ queries to SQL expressions. For EF Core: 2, 3, 5, 6.
```
ErikEJ, line 239: But EF Core 2 is now out of support ??

smitpatel:
Following #3392 we shouldn't merge this PR.

ErikEJ:
@smitpatel Why was this closed then?

https://github.com/dotnet/EntityFramework.Docs/pull/3354

smitpatel:
I think decision of #3392 is we want to remove where it causes confusion or force users to learn more things than they need to by leaving 2.x content but at the same time we want to preserve content for 2.x (where it is already written), where it could be useful and away from user. The extensions which have the version specified already can stay there (may be we will remove them at some point in future), there is certainly no need to add more content for 2.x at this stage.

ErikEJ:
Got it!

PR #3453, opened 2021-10-01
Added note regarding non-existent -IgnoreChanges

StanlyLife:
Added note regarding non-existen parameter for -IgnoreChanges


ajcvickers:
@StanlyLife Looks like you are adding a note to the EF6 documentation, but it is about EF Core. Was this your intention?

StanlyLife:
@ajcvickers No that was not my intention, my mistake!

PR #3456, opened 2021-10-01
Document SQL Server sparse columns

### LINKED ISSUE #3017: Document SQL Server sparse columns
https://github.com/dotnet/efcore/issues/8023

roji:
Closes #3017


entity-framework/core/providers/sql-server/columns.md
```diff
@@ -0,0 +1,22 @@
...
+ms.date: 10/1/2021
+uid: core/providers/sql-server/columns
+---
+# Column features specific to the Entity Framework Core SQL Server provider
+
+This page details column configuration options that are specific to the SQL Server provider.
+
+## Sparse columns
+
+Sparse columns are ordinary columns that have an optimized storage for null values, reducing the space requirements for null values at the cost of more overhead to retrieve non-NULL values.
```
AndriySvyryd, line 14:
```suggestion
Sparse columns are ordinary columns that have an optimized storage for null values, reducing the space requirements for null values at the cost of more overhead to retrieve non-null values.
```

entity-framework/core/providers/sql-server/columns.md
```diff
@@ -0,0 +1,22 @@
...
+
+As an example, consider a type hierarchy mapped via [the table-per-hierarchy (TPH) strategy](xref:core/modeling/inheritance#table-per-hierarchy-and-discriminator-configuration). In TPH, a single database table is used to hold all types in a hierarchy; this means that the table must contain columns for each and every property across the entire hierarchy, and for columns belonging to rare types, most rows will contain a NULL value for that column. In these cases, it may make sense to configure the column as *sparse*, in order to reduce the space requirements. The decision whether to make a column sparse must be made by the user, and depends on expectations for actual data in the table.
```
AndriySvyryd, line 16:
```suggestion
As an example, consider a type hierarchy mapped via [the table-per-hierarchy (TPH) strategy](xref:core/modeling/inheritance#table-per-hierarchy-and-discriminator-configuration). In TPH, a single database table is used to hold all types in a hierarchy; this means that the table must contain columns for each and every property across the entire hierarchy, and for columns belonging to rare types, most rows will contain a null value for that column. In these cases, it may make sense to configure the column as *sparse*, in order to reduce the space requirements. The decision whether to make a column sparse must be made by the user, and depends on expectations for actual data in the table.
```

AndriySvyryd (approved):

PR #3461, opened 2021-10-03
Update translation for SQL Server string.IsNullOrWhitespace

### LINKED ISSUE #3462: Docs out of date for SQL Server translation of IsNullOrWhiteSpace
Documentation here https://docs.microsoft.com/en-us/ef/core/providers/sql-server/functions

says that INOW() translates as 

    @value IS NULL OR LTRIM(RTRIM(@value)) = N''
    
If a string is whitespace, to my mind RTRIM would have reduced it to `''` already, so the decision can be made as to whether it's WS without the need to call LTRIM on it..?

roji:
Closes #3462

We could document that the translation changed across versions, but I honestly don't think it's worth it.


bricelam (approved):

bricelam:
> We could document that the translation changed across versions, but I honestly don't think it's worth it.

Agreed, a few other translations listed here were also different in previous releases. (Heck, some of them are straight up oversimplifications of the actual translation.)

PR #3472, opened 2021-10-11
Emphasize Lazy Loading does not work well with context pooling

raymens:
I added this small remark as I did not make the link between Lazy Loading and state when we first added Context Pooling. Adding ever-increasing memory usage.


ajcvickers:
**EF Team Triage:** Closing this issue as the requested additional details have not been provided and we have been unable to reproduce it.

*BTW this is a canned response and may have info or details that do not directly apply to this particular issue. While we'd like to spend the time to uniquely address every incoming issue, we get a lot traffic on the EF projects and that is not practical. To ensure we maximize the time we have to work on fixing bugs, implementing new features, etc. we use canned responses for common triage decisions.*

PR #3482, opened 2021-10-15
Add Pagination.EntityFrameworkCore.Extensions

SitholeWB:
This is a library for Pagination on EntityFrameworkCore. Works well with Entity Framework Core as an extension, support both asynchronous and synchronous. It also has many useful features commonly used especially on web development, examples and documentation here (https://github.com/SitholeWB/Pagination.EntityFrameworkCore.Extensions)


ajcvickers:
@SitholeWB Can you update the formatting to be consistent with the other extensions listed here.

Issue #1937, opened 2019-11-21
Document conventions API


AndriySvyryd:
For users and provider writers.

- ConventionSource
- ConventionContext
- ConventionDispatcher
- ConventionBatch
- ConventionSet
- ConventionsBuilder
- ModelCustomizer
- The dangers of `IConvention*` API

Move the paragraph about conventions from "Model Bulk Configuration" to the new page and replace it with a link.
Always check whether the entity type is still in the model in OnRemoved conventions

ajcvickers:
Update https://aka.ms/efcore-docs-conventions when done.

ajcvickers:
Update AKA link

PR #3533, opened 2021-11-05
Add documentation for pre-convention model configuration 

### LINKED ISSUE #571: Bulk configuring models
We should have a page showing how do to things like this:

``` csharp
foreach (var property in modelBuilder.Model.GetEntityTypes()
    .SelectMany(t => t.GetProperties())
    .Where(p => p.ClrType == typeof(decimal)))
{
    property.Relational().ColumnType = "decimal(18, 6)";
}
```

### LINKED ISSUE #3278: Document pre-convention model configuration
See https://github.com/dotnet/efcore/issues/12229, https://github.com/dotnet/efcore/issues/3867, https://github.com/dotnet/efcore/issues/10784

AndriySvyryd:
Part of #3278
Fixes #571

### DIFF
--- a/entity-framework/core/modeling/bulk-configuration.md
+++ b/entity-framework/core/modeling/bulk-configuration.md
@@ -1,0 +1,71 @@
+---
+title: Model Bulk Configuration - EF Core
+description: How to apply bulk configuration during model building in Entity Framework Core
+author: AndriySvyryd
+ms.date: 11/05/2021
+uid: core/modeling/bulk-configuration
+---
+# Model bulk configuration
+
+When an aspect needs to be configured in the same way across multiple entity types, the following techniques allow to reduce code duplication and consolidate the logic.
+
+See the [full sample project](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Modeling/BulkConfiguration) containing the code snippets presented below.
+
+## Bulk configuration in OnModelCreating
+
+Every builder object returned from <xref:Microsoft.EntityFrameworkCore.ModelBuilder> exposes a <xref:Microsoft.EntityFrameworkCore.ModelBuilder.Model> or `Metadata` property that provides a low-level access to the objects that comprise the model. In particular, there are methods that allow you to iterate over specific objects in the model and apply common configuration to them.
+
+In the following example the model contains a custom value type `Currency`:
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/Currency.cs?name=Currency)]
+
+Properties of this type are not discovered by default as the current EF provider doesn't know how to map it to a database type. This snippet of `OnModelCreating` adds all properties of the type `Currency` and configures a value converter to a supported type - `decimal`:
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/MetadataAPIContext.cs?name=MetadataAPI)]
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/CurrencyConverter.cs?name=CurrencyConverter)]
+
+### Drawbacks of the Metadata API
+
+- Unlike Fluent API, every modification to the model needs to be done explicitly. For example, if some of the `Currency` properties were configured as navigations by a convention then you need to first remove the navigation referencing the CLR property before adding an entity type property for it. [#9117](https://github.com/dotnet/efcore/issues/9117) will improve this.
+- The conventions run after each change. If you remove a navigation discovered by a convention then the convention will run again and could add it back. To prevent this from happening you would need to either delay the conventions until after the property is added by calling <xref:Microsoft.EntityFrameworkCore.Metadata.Conventions.IConventionContext.DelayConventions> and later disposing the returned object or to mark the CLR property as ignored using <xref:Microsoft.EntityFrameworkCore.Metadata.IMutableModel.AddIgnored%2A>.
+- Entity types might be added after this iteration happens and the configuration won't be applied to them. This can usually be prevented by placing this code at the end of `OnModelCreating`, but if you have two interdependent sets of configurations there might not be an order that will allow them to be applied consistently.
+
+## Pre-convention configuration
+
+EF Core 6.0 allows the mapping configuration to be specified once for a given CLR type; that configuration is then applied to all properties of that type in the model as they are discovered. This is called "pre-convention model configuration", since it configures aspects of the model that are then used by the model building conventions. Such configuration is applied by overriding `<xref:Microsoft.EntityFrameworkCore.DbContext.ConfigureConventions>` on the type derived from `<xref:Microsoft.EntityFrameworkCore.DbContext>`.
+
+This example shows how configure all properties of type `Currency` to have a value converter:
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/PreConventionContext.cs?name=CurrencyConversion)]
+
+And this example shows how to configure some facets on all properties of type `string`:
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/PreConventionContext.cs?name=StringFacets)]
+
+> [!NOTE]
+> The type specified in a call from `ConfigureConventions` can be a base type, an interface or a generic type definition. All matching configurations will be applied in order from the least specific:
+>
+> 1. Interface
+> 2. Base type
+> 3. Generic type definition
+> 4. Non-nullable value type
+> 5. Exact type
+
+### Ignoring types
+
+Pre-convention configuration also allows to ignore a type and prevent it from being discovered by conventions either as an entity type or as a property on an entity type:
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/PreConventionContext.cs?name=IgnoreInterface)]
+
+### Default type mapping
+
+Generally, EF is able to translate queries with constants of a type that is not supported by the provider, as long as you have specified a value converter for a property of this type. However, in queries that don't involve any properties of this type, there is no way for EF to find the correct value converter. In this case, it's possible to call <xref:Microsoft.EntityFrameworkCore.ModelConfigurationBuilder.DefaultTypeMapping> to add or override a provider type mapping:
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/PreConventionContext.cs?name=DefaultTypeMapping)]
+
+### Limitations of pre-convention configuration
+
+- Many aspects cannot be configured with this approach. [#6787](https://github.com/dotnet/efcore/issues/6787) will expand this to more types.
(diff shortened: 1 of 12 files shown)


entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
+---
+title: Model Bulk Configuration - EF Core
+description: How to apply bulk configuration during model building in Entity Framework Core
+author: AndriySvyryd
+ms.date: 11/05/2021
+uid: core/modeling/bulk-configuration
+---
+# Model bulk configuration
+
+When an aspect needs to be configured in the same way across multiple entity types the following techniques allow to reduce code duplication and consolidate the logic.
```
roji, line 10:
```suggestion
When an aspect needs to be configured in the same way across multiple entity types, the following techniques allow to reduce code duplication and consolidate the logic.
```

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+
+See the [full sample project](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Modeling/BulkConfiguration) containing the code snippets presented below.
+
+## Metadata API
+
+Every builder object returned from <xref:Microsoft.EntityFrameworkCore.ModelBuilder> exposes a <xref:Microsoft.EntityFrameworkCore.ModelBuilder.Model> or `Metadata` property that provides a low-level access to the objects that comprise the model. In particular, there are methods that allow you to iterate over specific objects in the model and apply common configuration to them.
+
+In the following example the model contains a custom value type `Money`:
```
roji, line 18: Money might be a bit confusing since there's actually a `money` type in databases...

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+ms.date: 11/05/2021
+uid: core/modeling/bulk-configuration
+---
+# Model bulk configuration
+
+When an aspect needs to be configured in the same way across multiple entity types the following techniques allow to reduce code duplication and consolidate the logic.
+
+See the [full sample project](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Modeling/BulkConfiguration) containing the code snippets presented below.
+
+## Metadata API
```
roji, line 14: I understand that this is about using the (regular) metadata API for bulk config, but the actual focus of this section seems to be "Bulk configuration in OnModelCreating" (the metadata API is just a means?)

```suggestion
## Bulk configuration in OnModelCreating
```

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+
+Every builder object returned from <xref:Microsoft.EntityFrameworkCore.ModelBuilder> exposes a <xref:Microsoft.EntityFrameworkCore.ModelBuilder.Model> or `Metadata` property that provides a low-level access to the objects that comprise the model. In particular, there are methods that allow you to iterate over specific objects in the model and apply common configuration to them.
```
roji, line 16: Is it important to discuss `Metadata` and low-level access? The section seems to simply be about bulk configuration in OnModelCreating via foreach, no?

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/MetadataAPIContext.cs?name=MetadataAPI)]
+
+### Drawbacks of the Metadata API
+
+- Unlike Fluent API, every modification to the model needs to be done explicitly. For example, if some of the `Money` properties were configured as navigations by a convention then you need to first remove the navigation referencing the CLR property before adding an entity type property for it. [#9117](https://github.com/dotnet/efcore/issues/9117) will improve this.
+- The conventions run after each change. If you remove a navigation discovered by a convention then the convention will run again and add it back. To prevent this from happening you would need to either delay the conventions until after the property is added by calling <xref:Microsoft.EntityFrameworkCore.Metadata.Conventions.IConventionContext.DelayConventions> or to mark the CLR property as ignored using <xref:Microsoft.EntityFrameworkCore.Metadata.IMutableModel.AddIgnored%2A>.
```
roji, line 29: I doubt many would understand what this means :sweat_smile:

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+- Entity types might be added after this iteration happens and the configuration won't be applied to them. This can usually be prevented by placing this code at the end of `OnModelCreating`, but if you have two interdependent sets of configurations there might not be an order that will allow them to be applied consistently.
+
+## Pre-convention configuration
+
+EF Core 6.0 allows the mapping configuration to be specified once for a given type. It will then be applied to all properties of that type in the model as they are discovered. This is called "pre-convention model configuration", since it configures aspects of the model that are then used by the model building conventions. Such configuration is applied by overriding `<xref:Microsoft.EntityFrameworkCore.DbContext.ConfigureConventions>` on the type derived from `<xref:Microsoft.EntityFrameworkCore.DbContext>`:
```
roji, line 34:
```suggestion
EF Core 6.0 allows the mapping configuration to be specified once for a given CLR type; that configuration is then applied to all properties of that type in the model as they are discovered. This is called "pre-convention model configuration", since it configures aspects of the model that are then used by the model building conventions. Such configuration is applied by overriding `<xref:Microsoft.EntityFrameworkCore.DbContext.ConfigureConventions>` on the type derived from `<xref:Microsoft.EntityFrameworkCore.DbContext>`:
```

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/PreConventionContext.cs?name=MoneyConversion)]
```
roji, line 36: Do we want to recommend/show putting the value converter as an inner type inside Money? It seems like a mixing business types (Money) with low-level data conversion stuff - consider moving it out.

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+>
+> 1. Interface
+> 2. Base type
+> 3. Generic type definition
+> 4. Non-nullable value type
+> 5. Exact type
+
+### Ignoring types
+
+Pre-convention configuration also allows to ignore a type and prevent it from being discovered by conventions either as an entity type or a entity type property:
```
roji, line 51:
```suggestion
Pre-convention configuration also allows to ignore a type and prevent it from being discovered by conventions either as an entity type or as a property on an entity type:
```

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/PreConventionContext.cs?name=IgnoreInterface)]
+
+### Default type mapping
+
+Generally, EF will be able to translate queries with constants of a type that is not supported by the provider if you have specified a value converter for a property of this type. However, in the queries that don't involve any properties of this type there is no way for EF to find the correct value converter. In this case it's possible to call <xref:Microsoft.EntityFrameworkCore.ModelConfigurationBuilder.DefaultTypeMapping> to add or override a provider type mapping:
```
roji, line 57:
```suggestion
Generally, EF is able to translate queries with constants of a type that is not supported by the provider, as long as you have specified a value converter for a property of this type. However, in queries that don't involve any properties of this type, there is no way for EF to find the correct value converter. In this case, it's possible to call <xref:Microsoft.EntityFrameworkCore.ModelConfigurationBuilder.DefaultTypeMapping> to add or override a provider type mapping:
```

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/PreConventionContext.cs?name=DefaultTypeMapping)]
+
+### Limitations of pre-convention configuration
+
+- Many aspects cannot be configured with this approach. [#6787](https://github.com/dotnet/efcore/issues/6787) will expand this to more types.
+- Currently the configuration is only determined by the CLR type. [#20418](https://github.com/dotnet/efcore/issues/20418) would allow custom predicates.
+- This configuration is performed before a model is created, if there are any conflicts that arise when applying it the exception stack trace will not contain the `ConfigureConventions` method, so it might be harder to find the cause.
```
roji, line 65:
```suggestion
- This configuration is performed before a model is created. If there are any conflicts that arise when applying it, the exception stack trace will not contain the `ConfigureConventions` method, so it might be harder to find the cause.
```

roji (approved):

entity-framework/core/modeling/bulk-configuration.md
AndriySvyryd: Yes, this is about how to get something to iterate on.

entity-framework/core/modeling/bulk-configuration.md
AndriySvyryd: The problem, the solution or both?

entity-framework/core/modeling/bulk-configuration.md
```diff
@@ -0,0 +1,67 @@
...
+- Entity types might be added after this iteration happens and the configuration won't be applied to them. This can usually be prevented by placing this code at the end of `OnModelCreating`, but if you have two interdependent sets of configurations there might not be an order that will allow them to be applied consistently.
+
+## Pre-convention configuration
+
+EF Core 6.0 allows the mapping configuration to be specified once for a given CLR type; that configuration is then applied to all properties of that type in the model as they are discovered. This is called "pre-convention model configuration", since it configures aspects of the model that are then used by the model building conventions. Such configuration is applied by overriding `<xref:Microsoft.EntityFrameworkCore.DbContext.ConfigureConventions>` on the type derived from `<xref:Microsoft.EntityFrameworkCore.DbContext>`:
+
+[!code-csharp[Main](../../../samples/core/Modeling/BulkConfiguration/PreConventionContext.cs?name=CurrencyConversion)]
```
smitpatel, line 38: Should write something about these code samples?

entity-framework/core/modeling/bulk-configuration.md
roji: I think both. To be clear, this isn't about your text or anything - it's just that understanding the reactivity of conventions and what that could mean is a very advanced, internal thing.

Maybe it's better to go with something vague, just to convey the "unreliability" of this mechanism (e.g. "Some changes done via the Fluent API may not result in the desired results. For example, if you remove a navigation discovered by a convention, then that convention will run again and add it back."). Otherwise we could go with this too - I just doubt it would help many people.

AndriySvyryd: We'll expand and move this in https://github.com/dotnet/EntityFramework.Docs/issues/1937

Issue #2167, opened 2020-03-06
> You can pass a full connection string to DbContext instead of just the database or connection string name.


StingyJack:
Please example this so we can see what this looks like. The overload says `string nameOrConnectionString` but I always see it in code with `name=SomeContext`. Is there some special sauce to differentiate between connection string and connection names?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7f431e6c-6c5f-9056-1486-99ba6b02c954
* Version Independent ID: 37943f43-c279-f190-b134-48f54ead1e38
* Content: [Connection strings and models - EF6](https://docs.microsoft.com/en-us/ef/ef6/fundamentals/configuring/connection-strings)
* Content Source: [entity-framework/ef6/fundamentals/configuring/connection-strings.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/ef6/fundamentals/configuring/connection-strings.md)
* Product: **entity-framework**
* Technology: **entity-framework-6**
* GitHub Login: @divega
* Microsoft Alias: **avickers**

ajcvickers:
Thank you for requesting a change to the EF6 documentation. We are prioritizing improvements to the EF Core documentation and at this time we are not making changes to the EF6 documentation.

However, we will consider well-written community pull requests to improve the EF6 documentation. The documentation source is located in the [EntityFramework.Docs](https://github.com/aspnet/EntityFramework.Docs) repository on GitHub. To submit a pull request:
* Find the file to update and click on it to open

![image](https://user-images.githubusercontent.com/1430078/64454137-10199400-d09f-11e9-9d1a-b7fdca2c518e.png)
* Click on the edit button (✏️) to edit in your web browser

![image](https://user-images.githubusercontent.com/1430078/64454321-85856480-d09f-11e9-85a6-1c93bc6611e2.png)
* When done, add a title and description, then use the "Create a new branch and start a pull request." option.

![image](https://user-images.githubusercontent.com/1430078/64454455-dac17600-d09f-11e9-922b-0346117011f5.png)

(Note that you can also clone the docs repo and do a GitHub PR in the normal way.)

StingyJack:
How am I supposed to create a PR to fix docs that are missing, when I'm asking about those docs?

PR #3539, opened 2021-11-07
Improve testing docs

### LINKED ISSUE #1308: Document in-memory database is rooted in internal service provider
See https://github.com/aspnet/EntityFrameworkCore/issues/9613#issuecomment-464664863

---
#### Document Details

* ID: c95e5c7e-882f-76f7-5301-2a5d1ede39d3
* Version Independent ID: 8754aa99-67a7-98ac-a51e-0ba995f96414
* Content: [Testing with InMemory - EF Core](https://docs.microsoft.com/en-us/ef/core/miscellaneous/testing/in-memory)
* Content Source: [entity-framework/core/miscellaneous/testing/in-memory.md](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/miscellaneous/testing/in-memory.md)
* GitHub Login: @rowanmiller
* Microsoft Alias: **divega**

### LINKED ISSUE #2478: More information on testing strategies
We have some information on InMemory vs. Sqlite in-memory vs. read database, but it seems like a more general/strategic doc for testing on EF Core:

* Don't try mocking DbContext, which basically means mocking LINQ
* The repository is a right level at which to mock (show an example)
* When should you pay the extra layer of the repository pattern, as opposed to testing against inmemory/sqlite/real database
* Patterns for efficient testing against a real database (clean database, transactions...)

### LINKED ISSUE #2699: Sharing Databases example : Seed is out of order with SQL Lite - should this be mentioned?
When using SQL Lite for my test DB provider, I had to use `Connection.Open()` **before** calling seed. This is most likely because SQLLite will destroy the DB once a connection is closed and so this order is different than SQL Server which, when I tested it, does need the order mentioned in the sample/article.

``` csharp
        public MyDbContextFixture()
        {
            Connection = new SqliteConnection("Filename=:memory:");
            Connection.Open();
            Seed();
            
        }
```

The sample uses SQL Server for an example and likely is most important with non-memory dbs, but some test suites w/in memory DBs may still benefit from using a fixture so I believe that may be worth mentioning in the sample or article.

---
#### Document Details

* ID: abc24a97-4ba7-89d4-eb07-61695a9304cf
* Version Independent ID: b93a52b2-071b-f6f2-6f19-705ccd6464d8
* Content: [Sharing databases between tests - EF Core](https://docs.microsoft.com/en-us/ef/core/miscellaneous/testing/sharing-databases)
* Content Source: [entity-framework/core/miscellaneous/testing/sharing-databases.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/miscellaneous/testing/sharing-databases.md)
* GitHub Login: @ajcvickers

### LINKED ISSUE #2971: Provide more details on unit testing
One of the goals of a good unit test is to isolate external dependencies. The external dependency in this case _is_ the database/EF Core. When I'm testing my controller or service that consumes the database, I should be able to test ONLY my code, not unit test the database/EF Core. These three examples break that _fundamental_ principle by introducing a database into the unit tests and then expressing that it is correct and appropriate to do so.

The documentation says not to attempt to mock the DbContext or IQueryable because it is difficult and fragile. Trying to manage the state of the database while running concurrent unit tests and having to return the database back to a known state between test runs is, in my experience, MORE difficult, fragile, and error prone, assuming its possible at all in a CI/CD environment where the unit tests run may not be able to access a database.

---
#### Document Details

* ID: 33851e00-5992-c9c8-d9b6-428ba8bde6a9
* Version Independent ID: 793410c2-2e9f-e31e-7e2c-ac766eac451b
* Content: [Testing code that uses EF Core - EF Core](https://docs.microsoft.com/en-us/ef/core/testing/)
* Content Source: [entity-framework/core/testing/index.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/testing/index.md)
* GitHub Login: @ajcvickers

### LINKED ISSUE #3039: So, eh...what is an ItemsContext and where does it come from?
[Enter feedback here]

---
#### Document Details

* ID: 4ef36e12-dfc2-ac09-76b7-4af887bc39c1
* Version Independent ID: 87e16518-5ade-2159-ba56-11fbf6b2bf69
* Content: [EF Core testing sample - EF Core](https://docs.microsoft.com/en-us/ef/core/testing/testing-sample)
* Content Source: [entity-framework/core/testing/testing-sample.md](https://github.com/dotnet/EntityFramework.Docs/blob/master/entity-framework/core/testing/testing-sample.md)
* GitHub Login: @ajcvickers

roji:
Closes #2971
Closes #3039
Closes #2699
Closes #2478
Closes #1308


JonPSmith:
Happy to give feedback, but how do I see the updated docs, or should I check back when this goes live?

roji:
@JonPSmith unfortunately preview rendering of doc PRs isn't available externally... You can review directly on the markdown changes here if you want (that's what we usually do), or you can wait for this to go live if that's painful.

I'll definitely take a look at your blog posts and the link above!

JonPSmith:
I found the `testing` pages in the `roji:Testing` branch and looked at the three pages. I assume the `index` page replaces the current page and the other two pages are new. Here are my comments

## Index page (I assume this replaces the current index page)

To be honest I like the current index page with the 3 approaches more than your new index. Manning (the publisher of my books) were clear that you need to quickly explain the key points and then expand on it. Your new index page has some good stuff in it but its a bit complicated for the index page.

Also I wouldn't push the unit test / integration test differences. I find the unit test / integration test split on using a database is a bit false because accessing a test database is so easy. For me an integration test is uses an external service, or running up a complete ASP.NET Core to test end-to-end. 

What I really like is you comparison table, and I would definitely keep that and place it earlier. In my book I have a summary image - [This article](https://www.thereformedprogrammer.net/new-features-for-unit-testing-your-entity-framework-core-5-code/) starts with a similar image (in the book I changed "Mocking the database" to "Stubbing the database"). Things like your table and my figure helps people quickly understand and decide what they should do.

## integration-testing (new)

I really like this page because it has nice step-by-step explanation.

## unit-testing  (new)

Interesting, you consider SQLite in-memory as a unit test. I agree but then again I don't see the unit test / integration test divide when using a database.

I hope that helps.

roji:
> Also I wouldn't push the unit test / integration test differences. I find the unit test / integration test split on using a database is a bit false because accessing a test database is so easy. 

Interesting... I agree that accessing a test database (on the same database system used on production, e.g. SQL Server) is very easy, and should be the "default" thing to do. But we've been seeing a lot of users going down the InMemory path, precisely because they're under the impression that testing against SQL Server is tough. In fact, a lot of the new docs are oriented precisely to dispel that myth, both by sharpening the unit vs. integration testing divide, and by showing practical approaches to successful integration tests.

> For me an integration test is uses an external service [...]

I don't think that's part of the accepted definition, at least not the way I understand it... IMHO integration testing is simply about testing the actual components without swapping any of them with a test double.

Think of it this way: if my *production* database happens to be SQLite (not in-memory), I may still want to isolate my application code and unit-test it without the database (e.g. via a repository pattern). The fact that SQLite happens to be an in-process database rather than something running on some remote server shouldn't be relevant.

> [...] or running up a complete ASP.NET Core to test end-to-end.

There's indeed a distinction here between end-to-end and integration testing. End-to-end testing would indeed imply firing up the ASP.NET application and talking to it via HttpClient; or if it's a user-facing website rather than a web API, driving a browser instead, to include the client side (e.g. Javascript) in the test.

But that's another beast altogether. I do think it's useful to cut out certain parts of your application (like ASP.NET, client-side Javascript), while keeping other parts in and not swapping them with test doubles (your production database); it's not necessarily all-or-nothing.

> Interesting, you consider SQLite in-memory as a unit test [...]

Well, if your production database is SQL Server, then SQLite in-memory is a [testing fake](https://martinfowler.com/bliki/TestDouble.html) - it's something you use to remove SQL Server from the mix in order to focus on your application code. In that sense it's really no different from the EF Core InMemory provider - both are fake databases in the testing context; even if SQLite is a superior choice, in both cases you're excluding SQL Server, and have no guarantee that what worked on your fake will work in production.

entity-framework/core/managing-schemas/ensure-created.md
```diff
@@ -8,18 +8,18 @@ uid: core/managing-schemas/ensure-created
 ---
 # Create and Drop APIs
 
-The EnsureCreated and EnsureDeleted methods provide a lightweight alternative to [Migrations](xref:core/managing-schemas/migrations/index) for managing the database schema. These methods are useful in scenarios when the data is transient and can be dropped when the schema changes. For example during prototyping, in tests, or for local caches.
+The <xref:Microsoft.EntityFrameworkCore.Storage.IDatabaseCreator.EnsureCreated> and <xref:Microsoft.EntityFrameworkCore.Storage.IDatabaseCreator.EnsureDeleted> methods provide a lightweight alternative to [Migrations](xref:core/managing-schemas/migrations/index) for managing the database schema. These methods are useful in scenarios when the data is transient and can be dropped when the schema changes. For example during prototyping, in tests, or for local caches.
```
ajcvickers, line 11: Is it valid to do xrefs like this? They certainly don't look right when viewing on GitHub.,

entity-framework/core/managing-schemas/ensure-created.md
```diff
@@ -8,18 +8,18 @@ uid: core/managing-schemas/ensure-created
...
 
-Some providers (especially non-relational ones) don't support Migrations. For these providers, EnsureCreated is often the easiest way to initialize the database schema.
+Some providers (especially non-relational ones) don't support Migrations. For these providers, `EnsureCreated` is often the easiest way to initialize the database schema.
 
 > [!WARNING]
-> EnsureCreated and Migrations don't work well together. If you're using Migrations, don't use EnsureCreated to initialize the schema.
+> `EnsureCreated` and Migrations don't work well together. If you're using Migrations, don't use `EnsureCreated` to initialize the schema.
```
ajcvickers, line 16: This now sounds a bit harsh. See https://github.com/dotnet/efcore/issues/26870#issuecomment-985364245. Maybe it would be good to just document how to do this as part of these changes?

entity-framework/core/managing-schemas/ensure-created.md
```diff
@@ -40,7 +40,7 @@ dbContext.Database.EnsureCreated();
 
 ## SQL Script
 
-To get the SQL used by EnsureCreated, you can use the GenerateCreateScript method.
+To get the SQL used by `EnsureCreated`, you can use the GenerateCreateScript method.
```
ajcvickers, line 43: xref to the GenerateCreateScript method?

entity-framework/core/providers/in-memory/index.md
```diff
@@ -7,7 +7,10 @@ uid: core/providers/in-memory/index
 ---
 # EF Core In-Memory Database Provider
 
-This database provider allows Entity Framework Core to be used with an in-memory database. The in-memory database can be useful for testing, although the SQLite provider in in-memory mode may be a more appropriate test replacement for relational databases. The in-memory database is designed for testing only. The provider is maintained as part of the [Entity Framework Core Project](https://github.com/dotnet/efcore).
+This database provider allows Entity Framework Core to be used with an in-memory database. While some users use the in-memory database for testing, this is generally discouraged; the SQLite provider in in-memory mode is a more appropriate test replacement for relational databases. For more information on how to test EF Core applications, see the [testing documentation](xref:core/testing/index). The provider is maintained as part of the [Entity Framework Core Project](https://github.com/dotnet/efcore).
+
+> [!WARNING]
+> The In-Memory provider was not designed for production use and should never be used as such.
```
ajcvickers, line 13: I wonder if "production use" is clear here. Maybe, "The In-Memory provider was not designed for use outside of testing environments and should never be used as such?"

entity-framework/core/testing/in-memory.md
```diff
@@ -1,24 +0,0 @@
----
```
ajcvickers, line 1: Is it okay to just delete this? Won't that result in 404s for existing links?

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
-ms.date: 04/22/2020
+author: roji
+ms.date: 11/07/2021
 uid: core/testing/index
 ---
-# Testing code that uses EF Core
+# Testing EF Core Applications
 
-Testing code that accesses a database requires either:
+Testing is an important concern to almost all application types - testing allows you to be sure your application work correctly, and makes it instantly known if its behavior regresses in the future. Since testing may affect how your code is architected, it's highly recommended to plan for a testing early, and to ensure good coverage as your application evolves. This section provides an overview of various testing strategies for applications using EF Core, shows how each one can be implemented and discusses best practices.
```
ajcvickers, line 10: Grammar: "your application work correctly"

ajcvickers: Also, Oxford comma missing.

ajcvickers: This is a very opinionated statement about testing applications. Are we really the ones who should be saying. "you must test!" as opposed to saying, "if you want to test, this might help?"

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
 
-* Running queries and updates against the same database system used in production.
-* Running queries and updates against some other easier to manage database system.
-* Using test doubles or some other mechanism to avoid using a database at all.
+## Unit and integration testing
 
-This document outlines the trade-offs involved in each of these choices and shows how EF Core can be used with each approach.
+Although there are many kinds of testing, in this section we'll concentrate on two types of automated testing: unit and integration testing. Without going into formal definitions, unit tests are about isolating a single component and testing against it, with all related components (or dependencies) stubbed out. Importantly, unit tests never interact with a real database, but instead replace the database with a [test double](https://martinfowler.com/bliki/TestDouble.html) which fulfills the same contracts. Integration tests, in contrast, test actual components working together - including your production database system.
```
ajcvickers, line 14: I predict you're going to have a fun time arguing with people who disagree with these statements. Again, I wonder if we should be making these distinctions, as opposed to just showing techniques for different types of tests.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
 
-> [!TIP]
-> See [EF Core testing sample](xref:core/testing/testing-sample) for code demonstrating the concepts introduced here.
+Integration tests frequently involve the following difficulties:
 
-## All database providers are not equal
+1. In many cases, it's simply not possible or practical to test against the actual external dependency. For example, your application may to interact with some service that cannot be tested again (because of rate limiting, or the lack of a testing environment).
```
ajcvickers, line 18: Typo: again.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
+2. Even when it's possible to involve the real dependency, this may be exceedingly slow: running a large amount of tests against a cloud service may cause tests to take too long. Testing should be part of the developer's everyday workflow, so it's important that tests run quickly.
```
ajcvickers, line 19: More opinionated statements about how/when tests are run.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
 
-Fundamentally, this means that switching out the database provider will change EF Core behavior and the application can't be expected to function correctly unless it explicitly accounts for any differences in behavior.
-That being said, in many cases doing this will work because there is a high degree of commonality amongst relational databases.
-This is good and bad.
-Good because moving between database systems can be relatively easy.
-Bad because it can give a false sense of security if the application is not fully tested against the new database system.
+## Integration testing
 
-## Approach 1: Production database system
+Because of the above difficulties with integration tests, developers are frequently urged to concentrate on unit tests first, and have a robust unit test suite which they can run frequently on their machines; integration tests, in contrast, are supposed to be executed much less frequently, and in many cases also provide far less coverage. We recommend giving more thought to integration tests, and that that databases are actually far less affected by the above problems as much as people tend to think:
```
ajcvickers, line 26: "We recommend giving more thought to integration tests, and that that databases are actually far less affected by the above problems as much as people tend to think" doesn't parse for me.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
 
-As described in the previous section, the only way to be sure you are testing what runs in production is to use the same database system.
-For example, if the deployed application uses SQL Azure, then testing should also be done against SQL Azure.
+1. Most databases can nowadays be easily installed on the developer's machine. Container-based technologies such as Docker can make this very easy, and technologies such as [Github Workspaces](https://docs.github.com/en/codespaces/overview) and [Dev Container](https://code.visualstudio.com/docs/remote/create-dev-container) set up your entire development environment for you (including the database). When using SQL Server, it's also possible to test against [LocalDB](/sql/database-engine/configure-windows/sql-server-express-localdb).
```
ajcvickers, line 28: LocalDb is Windows-only.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
+Unlike integration tests, unit tests never involve the actual database system used in production, but rather swap it with a [test double](https://en.wikipedia.org/wiki/Test_double). Below are some common methods for unit testing EF Core applications with test doubles:
 
-All the major database systems have some form of "Developer Edition" for local testing.
-SQL Server also has a feature called [LocalDB](/sql/database-engine/configure-windows/sql-server-express-localdb).
-The primary advantage of LocalDB is that it spins up the database instance on demand.
-This avoids having a database service running on your machine even when you're not running tests.
+1. Use SQLite (in-memory mode) as a database fake, replacing your production database system.
+2. Use the EF Core InMemory provider as a database fake, replacing your production database system.
+3. Mocking or stubbing `DbContext` and `DbSet`.
+4. Introduce a repository layer between EF Core and your application code, and mocking or stubbing that layer.
```
ajcvickers, line 41: Grammar.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
 
-LocalDB is not without its issues:
+Below, we'll explore what each method means, and compare it with the others. We recommend reading through the different methods to gain a full understanding, but if you're looking for a quick suggestion, we highly recommend the 4th method: introducing a repository layer.
```
ajcvickers, line 43: Do we highly recommend using a repository? I don't. I highly recommend not writing the kind of unit tests that require a repository.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
+* Raw SQL may work, or it may fail or return different results, depending on exactly what is being done. SQL dialects are different in many ways across databases.
 
-EF Core tests the SQL Server provider primarily by running it against a local SQL Server instance.
-These tests run tens of thousands of queries in a couple of minutes on a fast machine.
-This illustrates that using the real database system can be a performant solution.
-It is a myth that using some lighter-weight database is the only way to run tests quickly.
+Compared to running integration tests against your production database system, it's relatively easy to get started with SQLite, and so many users do. Unfortunately, the above limitations tend to eventually become problematic when unit testing EF Core applications, even if they don't seem to be at the beginning. As a result, we recommend either concentrating on integration tests, or implementing unit tests via the repository pattern as discussed below.
```
ajcvickers, line 56: I disagree with this. I don't think using SQLite is any easier than using your production database system. In fact, it can be harder because a) you need your tests to configure a whole different provider, rather than just a connection string, and b) it is hard to understand and control the lifetime of the SQLite in-memory mode.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
 
-However, remember that:
+* The InMemory provider generally supports less query types than the SQLite provider, since it isn't a relational database. More queries will fail or behave differently in comparison to your production database.
+* Transactions are not supported.
+* Raw SQL is completely unsupported. Compare this with SQLite, where it's possible to use raw SQL, as long as that SQL works in the same way on SQLite and your production database.
+* The InMemory provider has not been optimized for performance, and will generally work slower than SQLite in in-memory mode.
 
-* SQLite inevitably doesn't support everything that your production database system does.
-* SQLite will behave differently than your production database system for some queries.
+In summary, InMemory has all the disadvantages of SQLite, along with a few more - and offers no advantages in return. If you are looking for a simple, in-memory database fake, use SQLite instead of the InMemory provider; but consider using the repository pattern instead as described below.
```
ajcvickers, line 69: I think it has the advantage that its very easy to use.

It's interesting, but after reading these sections my conclusions are:
- Don't use another database as a fake unless what the database does is secondary to what you are testing. (In other words, it can be fine to easily get a list of customers and products that are then used to test some business logic.)
- If you're going to use a fake, then understand that is a fake and use the easiest thing possible, which might be in-memory.

Fundamentally, neither SQLite not the in-memory provider will behave the same as your production database, so _it doesn't make sense to use either of these_ for any testing where that is important. Further, it _is easy_ to use some form of your production database system, and _it doesn't have these issues_, so do that. It will always be better.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
-On the other hand, these things tend to be very important when testing an application that uses EF Core.
+### Repository pattern
 
-## Unit testing
+The approaches above attempted to either swap EF Core's production database provider with a fake testing provider, or to stub out `DbSet` itself with an in-memory collection. These techniques are similar in that they evaluate the program's LINQ queries - either in SQLite or in memory - and this is ultimately the source of the difficulties outlined above: a query designed to run against a specific production database cannot run elsewhere without issues.
+
+For proper, reliable unit testing, consider introducing a [repository layer](https://martinfowler.com/eaaCatalog/repository.html) which mediates between your application code and EF Core. The production implementation of the repository contains the actual LINQ queries and executes them via EF Core. In testing, the repository abstraction is directly stubbed or mocked without being needing any actual LINQ queries, effectively removing EF Core from your testing stack altogether and allowing tests to focus on application code alone.
```
ajcvickers, line 85: "being needing"

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,116 @@
...
+![Comparison of fake provider with repository pattern](_static/fake-provider-and-repository-pattern.png)
+
+Since LINQ queries are no longer part of testing, you can directly provide query results to your application. Put another way, the previous approaches roughly allow stubbing out *query inputs* (e.g. replacing SQL Server tables with an in-memory one), but then still execute the actual query operators in-memory; the repository pattern, in contrast, allows you to stub out *query outputs* directly, allowing for far more powerful and focused unit testing.
+
+One possible disadvantage of the repository pattern is that it requires you to architect your application accordingly: application code can no longer execute LINQ queries directly, but must instead call into a repository method, which executes the query (and is stubbed in testing). This can make for a somewhat heavier application architecture, but can also have other advantages: all data access code is concentrated in one place rather than being spread around the application, and if your application needs to support more than one database, the repository abstraction can be very helpful for tweaking queries across providers.
```
ajcvickers, line 93: I think it's important to point out that not making IQueryable available to your application has _very significant_ disadvantages. IQueryable, and that way it can be used to compose complex queries, is, I think, one of the things that makes LINQ and C#/.NET great. If you hide all that behind a repository, then you have introduce your own abstractions for creating complex queries. This is cleaner in terms of separation of concerns and not letting your persistence layer leak into your domain layer, but is also consequentially _much more limiting_ (because by-design a repository limits the queries that can be constructed) and _much more work_. If you're doing this only for the dubious benefit of unit-testing your persistence layer, then I think very often you will be making the wrong tradeoff.

It's also very important to emphasize that repository is only beneficial if it really does abstract way queries into well-defined, domain specific requests of the repository. This is why the "generic repository" pattern is generally pointless--especially if exposes IQueryables directly, but even if it just replaces IQueryable with another way to build general-purpose queries that may behave differently on different providers.

ajcvickers:
@roji Going to hold off and let your respond to some of my comments before reviewing further.

entity-framework/core/testing/index.md
ErikEJ: But SQL Server Express is xplat, and available as a docker image

entity-framework/core/testing/index.md
JeremyLikness: I may want the repo to handle business logic for C.U.D. operations but still have the flexibility to build a query based on criteria. The repo can still manage the connection, context, etc. but if it exposes queries like this, you can still test without a live database and build logic into your repo to approve the expression tree before it's used.

```csharp
// signature 
Task<ICollection<T>> Query<T>(Func<IQueryable<T>, IQueryable<T>> filter);

// repo
using var ctx = new Context();
return await filter(ctx.Set<T>()).ToListAsync();

// consumer
var sold = await repo.Query<Item>(q => q.Where(i => i.IsSold));
```

entity-framework/core/managing-schemas/ensure-created.md
roji: We do this in many places, e.g. https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/change-tracking/change-detection.md?plain=1#L129 (which looks fine [here](https://docs.microsoft.com/en-us/ef/core/change-tracking/change-detection#methods-that-automatically-detect-changes)). It's definitely not standard markdown (and so looks bad on github markdown rendering), but...

Am I missing something?

entity-framework/core/managing-schemas/ensure-created.md
roji: Do you mean specifically wrt testing, e.g. use EnsureCreated when testing but Migrations in production? We could amend the warning for that, though I still think it's generally valid, at least until we do #2167, no?

entity-framework/core/providers/in-memory/index.md
roji: Better

entity-framework/core/testing/in-memory.md
roji: I added a redirection here to the testing index here: https://github.com/dotnet/EntityFramework.Docs/pull/3539/files#diff-3fa6f2a8672a06f5c881c8621af9e49f935e65eb99e06ec73d42afe0b1323d7aR459, is that enough?

entity-framework/core/testing/index.md
roji: I agree it's "programmatic", though I wouldn't say this is very controversial :) I think it's OK as a general introductory blurb, but if you believe this is too much I can tone it down.

entity-framework/core/testing/index.md
roji: Interesting - these seem like pretty uncontroversial definitions to me... and I even explicitly say we're avoiding formal definitions (and link to Martin Fowler, who specifically talks about replacing real databases)... Do you think it's controversial to say that integration tests include your real database, whereas unit tests do not?

The thing is that within these docs, it's important to make the distinction between these two testing types, since the whole docs are split along those lines. I also think lots of users are a bit blurry on what kind of testing they want (and also what it is they're trying to test), so making this clearer may help.

entity-framework/core/testing/index.md
roji: Here I'm trying to list reasons which *may* lead users to abandon the idea of testing against their real database - (speed, isolation...), precisely in order to disprove them later and show that integration testing isn't as bad as they think. I think the "may", "in many cases" make this non-opinionated - just a list of possible reasons why someone would avoid it... 

Any suggestions on making this better?

entity-framework/core/testing/index.md
roji: You're right. Changed to "We recommend giving more thought to integration tests, and suggest that databases may actually be far less affected by the above problems as much as people tend to think", better?

entity-framework/core/testing/index.md
roji: Changing last sentence to: "When using SQL Server, it's also possible to test against [LocalDB](/sql/database-engine/configure-windows/sql-server-express-localdb) on Windows, or easily set up a Docker image on Linux."

entity-framework/core/testing/index.md
roji: Made consistent and corrected.

entity-framework/core/testing/index.md
roji: That's true, though on the other hand:

1. You don't have to set it up (e.g. Docker, PG installation), which some people may consider problematic
2. You don't have to deal with test isolation issues (run tests in transactions, use respawn to reset and reseed database state...).

Note that I don't disagree with you... maybe "it may be easier to get started with SQLite" instead of "it's relatively easy ..."?

entity-framework/core/testing/index.md
roji: We briefly discussed this offilne. Here are some points:

* I agree with yuor first bullet point: if you're using only very basic/universal database capabilities - simple read/write, no EF.Functions - using a fake (either InMemory or SQLite) may be sufficient; it's certainly easier to set up testing (e.g. in a CI pipeline) against a fake, and unlike repository doesn't impact your program architecture in an extreme way.
* However, on the second point, I do believe SQLite is advantageous because of the 4 points listed above: you get transactions, raw SQL (where the SQL works the same way as on your production database!), more queries translate and in a more similar way (because it's relational), and possibly speed. So within the (necessarily limited!) world of fake databases, SQLite does widen the scope of stuff you can unit-test without adopting a repository.
* Sure, there's that lifetime issue of SQLite in-memory, but that's a one-time startup thing you need to grok, and then it's OK. I wouldn't recommend against SQLite solely because of that.

> Fundamentally, neither SQLite not the in-memory provider will behave the same as your production database, so it doesn't make sense to use either of these for any testing where that is important. Further, it is easy to use some form of your production database system, and it doesn't have these issues, so do that. It will always be better.

So again, I think it's a matter of degree. For full testability/fidelity, I agree you need your real database, or a repository layer. But if you're willing to compromise on that to avoid real database setup, isolation issues, etc., and your application doesn't need EF.Functions.Whatever, then a fake can be reasonable, and some fakes (SQLite) are closer to your real database than others (InMemory).

entity-framework/core/testing/index.md
roji: That's an interesting point. We're now discussing certain "dynamic" scenarios where e.g. the web API itself exposes flexible querying, right?

First, my unfounded intuition is that flexible, dynamic query APIs are on the more rare/advanced side. I think a lot of scenarios have fixed, preset query shapes which can easily be abstracted behind repository methods with regular parameters. Limited dynamic scenarios can be implemented without full-blown IQueryable by having the repository can expose a simplified, non-queryable API (e.g. a string parameter with the name of the property to order by).

Now, for cases where a full IQueryable really is needed (totally arbitrary filterable/sortable/projectable web API, again - seems rare)... Users could expose IQueryable here, and have their repository mock perform verification that the given query tree has the expected shape. I don't think this is fundamentally different from mock-style verification of other parameter types - just more elaborate/complex. So I don't think the interdiction against exposing IQueryable from repositories has to be absolute; it's just good practice to avoid it unless really needed.

> It's also very important to emphasize that repository is only beneficial if it really does abstract way queries into well-defined, domain specific requests of the repository [...]

I completely agree. I was mainly trying to avoid going into the *general* repository discussion here, and to only talk about repositories as they relate to testing (this paragraph goes a bit beyond that).

roji:
@ajcvickers answered your comments above. If you want we can have a talk about this together in the coming days, it may be a more efficient way to move forward.

atrauzzi:
@roji - Got here from your link in our twitter conversation.  Overall, I'm obviously super excited to see as much advice as possible for testing with EF. I'm looking forward to seeing the fully formatted version. 

I did scan through to see the section about using containers in test infrastructure.  I want to put forward that the explanation - while doable - doesn't necessarily prove that it's any better or any easier than what people are looking for by using the in-memory provider. It might just be sidestepping the objective by presenting containers (and a load of ops effort) as equivalent, which - it pains me to say - they are not.

(I agree that you're "not testing the real thing" without using an official implementation, but there are many cases of "vanilla LINQ" that are perfectly eligible for the more unit-y tests using the in-memory driver.)

I'll also reiterate here that it would be really nice to see these docs mention libraries like AutoFixture and NSubstitute.  It would also be nice to see advice on how to structure scenario setup code.

:heart:

roji:
Thanks for the feedback @atrauzzi!

Yeah, I'm planning to revisit this PR (which was originally done a while ago already), and refine certain things. It's also a basis for further improvement - we can add more content gradually rather than trying to get everything in at the get-go. I'll try to get this merged and online in the coming days, and then we can further reiterate on it.

As I wrote on twitter (and in these docs), I generally have the feeling that the fear of the ops overhead around testing against a real database are somewhat exaggerated/overblown. This is among other things a result of my own experience working with Npgsql and EF, which very heavily lean on integration tests against real databases - and do so quite reliably.

Having said that, my intention in these docs was also to provide guidance for whatever path the user chooses, so InMemory and SQLite are definitely covered (even if slightly discouraged).

entity-framework/core/testing/index.md
roji: I've tweaked things a bit around this. I'm still using *integration* and *unit* tests in the same way (i.e. with or without the production database system), but I introduce these terms as definitions for simplicity's sake in the new overview page. Hope that helps.

roji:
@ajcvickers I've pushed some tweaking, especially to have a short, quick summary/overview page, and to keep the much longer information in a separate "Choosing a testing strategy" page.

I'm ready to quickly iterate over this if you are - it would be good to finally get this in. Note that we also have the testing community standup with @JonPSmith next week, ideally these docs will be live by then (we can always continue refining afterwards).

entity-framework/core/testing/index.md
JonPSmith: Hi @roji,

I also use the phase _automated testing_ when talking about testing EF Core code. The problem with the term _integration testing_ is people thinks that hard when in fact using EF Core is much easier that mocking or subbing a repository.

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,26 @@
...
-* Running queries and updates against some other easier to manage database system.
-* Using test doubles or some other mechanism to avoid using a database at all.
+## Involving the database (or not)
 
-This document outlines the trade-offs involved in each of these choices and shows how EF Core can be used with each approach.
+When writing tests for your EF Core application, one basic decision you need to make is whether your tests will involve your production database system - just as your application does - or whether your tests will run against a [test double](https://martinfowler.com/bliki/TestDouble.html), which replaces your production database system. Two prominent examples of test doubles in the EF Core context are [SQLite in-memory mode](xref:core/testing/choosing-a-testing-strategy#sqlite-as-a-database-fake), and the [InMemory provider](xref:core/testing/choosing-a-testing-strategy#inmemory-as-a-database-fake). For simplicity's sake, we'll use the term *integration tests* to refer to tests which involve your production database systems, and *unit tests* to refer to tests which use a test double.
```
ajcvickers, line 14: "we'll use the term *integration tests* to refer to tests which involve your production database systems, and *unit tests* to refer to tests which use a test double."

We should start a pool on how long it is before somebody challenges this definition. ;-)

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,26 @@
...
 
-> [!TIP]
-> See [EF Core testing sample](xref:core/testing/testing-sample) for code demonstrating the concepts introduced here.
+For an in-depth comparison and analysis of integration and unit testing approaches in EF Core applications, see [Choosing a testing strategy](xref:core/testing/choosing-a-testing-strategy). Below is a short point-by-point summary to help you get up to speed with the different options:
 
-## All database providers are not equal
+* Developers frequently turn to unit testing because they believe testing against their production database system is difficult or slow. This isn't always true in our experience, and we suggest giving this approach a chance: [Integration testing](xref:core/testing/integration-testing) provides techniques for reliable, fast integration testing. Writing at least some integration tests is usually necessary in any case - to make sure your application actually works against your production database - and unit testing can be limited in what it allow you to test (see below).
```
ajcvickers, line 18: allow -> allows

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,26 @@
...
+* The [InMemory provider](xref:core/testing/choosing-a-testing-strategy#inmemory-as-a-database-fake) will not behave like your real database in many important ways. Some features cannot be tested with it at all (e.g. transactions, raw SQL..), some others may behave differently than your production database (e.g. case-sensitivity in queries). While InMemory can work for simple, constrained query scenarios, it is highly limited and we discourage its use.
+  * Mocking `DbSet` for querying suffers from the same disadvantages as the InMemory approach; we discourage this as well.
```
ajcvickers, line 20:
```suggestion
  * Mocking `DbSet` for querying is complex and difficult, and suffers from the same disadvantages as the InMemory approach; we discourage this as well.
```

entity-framework/core/testing/index.md
ajcvickers, line 18:
```suggestion
* Developers frequently turn to unit testing because they believe testing against their production database system is difficult or slow. This isn't always true in our experience, and we suggest giving this approach a chance: [Integration testing](xref:core/testing/integration-testing) provides techniques for reliable, fast integration testing. Writing at least some integration tests is usually necessary in any case - to make sure your application actually works against your production database - and unit testing can be limited in what it allows you to test (see below).
```

entity-framework/core/testing/index.md
ajcvickers, line 14: "InMemory" is usually styled as "in-memory".

entity-framework/core/testing/index.md
```diff
@@ -1,117 +1,26 @@
...
+* [SQLite in-memory mode](xref:core/testing/choosing-a-testing-strategy#sqlite-as-a-database-fake) offers better compatibility with production relational databases, since SQLite is itself a full-fledged relational database. However, there will still be some important discrepancies between SQLite and your production database, and some features cannot be tested at all (e.g. provider-specific methods on EF.Functions).
+* For a unit testing approach that allows you to use a reliable test double for all the functionality of your production database system, it's possible to introduce a [repository layer](xref:core/testing/choosing-a-testing-strategy#repository-pattern) in your application. This allows you to exclude EF Core entirely from testing and to fully mock the repository; however, this alters the architecture of your application in a way which could be significant.
```
ajcvickers, line 22: 👍

entity-framework/core/testing/choosing-a-testing-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Integration testing
+
+Because of the above difficulties with integration tests, developers are frequently urged to concentrate on unit tests first, and have a robust unit test suite which they can run frequently on their machines; integration tests, in contrast, are supposed to be executed much less frequently, and in many cases also provide far less coverage. We recommend giving more thought to integration tests, and suggest that databases may actually be far less affected by the above problems as much as people tend to think:
```
ajcvickers, line 22:
```suggestion
Because of the above difficulties with integration tests, developers are frequently urged to concentrate on unit tests first, and have a robust unit test suite which they can run frequently on their machines; integration tests, in contrast, are supposed to be executed much less frequently, and in many cases also provide far less coverage. We recommend giving more thought to integration tests, and suggest that databases may actually be far less affected by the above problems than people tend to think:
```

entity-framework/core/testing/choosing-a-testing-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+## Unit testing
+
+Unlike integration tests, unit tests never involve the actual database system used in production, but rather swap it with a [test double](https://en.wikipedia.org/wiki/Test_double). Below are some common methods for unit testing EF Core applications with test doubles:
+
+1. Use SQLite (in-memory mode) as a database fake, replacing your production database system.
+2. Use the EF Core InMemory provider as a database fake, replacing your production database system.
+3. Mock or stub out `DbContext` and `DbSet`.
+4. Introduce a repository layer between EF Core and your application code, and mock or stub that layer.
+
+Below, we'll explore what each method means, and compare it with the others. We recommend reading through the different methods to gain a full understanding, but if you're looking for a quick suggestion, we highly recommend the 4th method: introducing a repository layer.
```
ajcvickers, line 39: I still don't think we should say we highly recommend a repository layer.

entity-framework/core/testing/choosing-a-testing-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+* Some queries which work on SQL Server simply aren't supported on SQLite, because the exact SQL support in these two database differs.
+* If your query happens to use a provider-specific method such as SQL Server's [`EF.Functions.DateDiffDay`](xref:core/providers/sql-server/functions#date-and-time-functions), that query will fail on SQLite, and cannot be tested.
+* Raw SQL may work, or it may fail or return different results, depending on exactly what is being done. SQL dialects are different in many ways across databases.
+
+Compared to running integration tests against your production database system, it's relatively easy to get started with SQLite, and so many users do. Unfortunately, the above limitations tend to eventually become problematic when unit testing EF Core applications, even if they don't seem to be at the beginning. As a result, we recommend either concentrating on integration tests, or implementing unit tests via the repository pattern as discussed below.
```
ajcvickers, line 52:
```suggestion
Compared to running integration tests against your production database system, it's relatively easy to get started with SQLite, and so many users do. Unfortunately, the above limitations tend to eventually become problematic when unit testing EF Core applications, even if they don't seem to be at the beginning. As a result, we recommend either concentrating on integration tests, or if unit tests for these things are an absolute necessity, then taking onboard the cost of a repository pattern as discussed below.
```

entity-framework/core/testing/choosing-a-testing-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Mocking or stubbing DbContext and DbSet
+
+This approach typically uses a mock framework to create a test double of `DbContext` and `DbSet`, and tests against those doubles. Mocking `DbContext` can be a good approach for testing various *non-query* functionality, such as calls to <xref:Microsoft.EntityFrameworkCore.DbContext.Add%2A> or <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges>, allowing you to verify that your code called them in write scenarios.
+
+However, properly mocking `DbSet` *query* functionality is not possible, since queries are expressed via LINQ operators, which are static extension method calls over `IQueryable`. As a result, when some people talk about "mocking `DbSet`", what they really mean is that they stub `DbSet` out with an in-memory collection, and then evaluate query operators against that collection in memory, just like a simple `IEnumerable`. This isn't the same as actually mocking or stubbing the database, which would involve stubbing out query **end results**, rather than query inputs (i.e. the database table, as represented by a `DbSet`).
```
ajcvickers, line 73:
```suggestion
However, properly mocking `DbSet` *query* functionality is extremely difficult, since queries are expressed via LINQ operators, which are static extension method calls over `IQueryable`. As a result, when some people talk about "mocking `DbSet`", what they really mean is that they stub `DbSet` out with an in-memory collection, and then evaluate query operators against that collection in memory, just like a simple `IEnumerable`. This isn't the same as actually mocking or stubbing the database, which would involve stubbing out query **end results**, rather than query inputs (i.e. the database table, as represented by a `DbSet`).
```

ajcvickers: Or maybe, "practically impossible". You can do it--you basically write your own LINQ provider for your tests.

entity-framework/core/testing/choosing-a-testing-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+![Comparison of fake provider with repository pattern](_static/fake-provider-and-repository-pattern.png)
+
+Since LINQ queries are no longer part of testing, you can directly provide query results to your application. Put another way, the previous approaches roughly allow stubbing out *query inputs* (e.g. replacing SQL Server tables with an in-memory one), but then still execute the actual query operators in-memory; the repository pattern, in contrast, allows you to stub out *query outputs* directly, allowing for far more powerful and focused unit testing.
+
+One possible disadvantage of the repository pattern is that it requires you to architect your application accordingly: application code can no longer execute LINQ queries directly, but must instead call into a repository method, which executes the query (and is stubbed in testing). This can make for a somewhat heavier application architecture, but can also have other advantages: all data access code is concentrated in one place rather than being spread around the application, and if your application needs to support more than one database, the repository abstraction can be very helpful for tweaking queries across providers.
```
ajcvickers, line 89: I think it's important to call out even more explicitly the disadvantages of a repository pattern. Maybe something like:

> LINQ is a very powerful feature of .NET. It allows applications to compose complex queries that are translated and executed against the database without the need to shape the data layer to explicitly support these queries. In addition to being powerful, this also provides a great deal of flexibility for the evolution of applications over time.

> Using a repository pattern to facilitate unit testing as described here necessarily prevents applications from using LINQ in this way, since if the repository itself exposes LINQ, then the testing problem has not been solved. This means that there is a significant cost involved in implementing and maintaining a repository. This cost should not be discounted when making a choice on how to test an application, especially given that integration tests for the queries generated by the repository will still be needed anyway.

> It's worth noting that repositories do have advantages outside of just testing. They ensure all data access code is concentrated in one place rather than being spread around the application, and if your application needs to support more than one database, then the repository abstraction can be very helpful for tweaking queries across providers.

entity-framework/core/testing/choosing-a-testing-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+The following table provides a quick, comparative view of the different testing techniques, and shows which functionality can be tested under which approach:
+
+Feature                             | InMemory     | SQLite in-memory          | Mock DbContext | Repository pattern | Integration testing
+----------------------------------- | ------------ | ------------------------- | -------------- | ------------------ | -------------------
+Test double type                    | Fake         | Fake                      | Mock/stub      | Mock/stub          | Real, no double
+Raw SQL?                            | No           | Depends                   | No             | Yes                | Yes
+Transactions?                       | No (ignored) | Yes                       | Yes            | Yes                | Yes
+Provider-specific translations?     | No           | No                        | No             | Yes                | Yes
+Exact query behavior?               | Depends      | Depends                   | Depends        | Yes                | Yes
+Requires special code architecture? | No           | No                        | No             | Yes                | No
```
ajcvickers, line 104:
```suggestion
Requires special code architecture? | No           | No                        | No             | Yes                | No
Restricts the use of LINQ | No | No | No | Yes | No
```

entity-framework/core/testing/choosing-a-testing-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Summary
+
+* We highly recommend that developers have good test coverage of their application running against their actual production database system. This provides confidence that the application actually works in production, and with proper design, tests can execute reliably and quickly. Since integration tests are required in any case, it's a good idea to start there, and add unit tests later, based on need.
+* For unit testing, we recommend implementing the repository pattern, allowing you to properly stub or mock out your data access layer above EF Core, rather than via a fake EF Core provider (Sqlite/InMemory) or by mocking `DbSet`.
```
ajcvickers, line 109:
```suggestion
* If unit tests for queries are absolutely required, then we recommend implementing the repository pattern, allowing you to properly stub or mock out your data access layer above EF Core, rather than via a fake EF Core provider (Sqlite/InMemory) or by mocking `DbSet`.
```

entity-framework/core/managing-schemas/ensure-created.md
ajcvickers: If you only ever use the migrations as scaffolding (which seems common), then EnsureCreated should be fine. Its only if you edit the migrations that EnsureCreated is no longer a good choice. Contrast this to the case for EF6, where using the equivalent of EnsureCreated would (attempt to) discover your migrations and execute those, so the database created would be the same either way.

entity-framework/core/testing/index.md
ajcvickers: I asked a scholar from Oxford, and they said the comma must be used. I then asked a scholar from Cambridge, and they agreed. If scholars from Oxford and Cambridge agree on something, then it must be right.

samples/core/Testing/BusinessLogic/IBloggingRepository.cs
```diff
@@ -0,0 +1,15 @@
+using System.Collections.Generic;
+
+namespace EF.Testing.BusinessLogic
+{
+    #region IBloggingRepository
+    public interface IBloggingRepository
+    {
+        Blog GetBlogByName(string name);
+        IEnumerable<Blog> GetAllBlogs();
+        void AddBlog(Blog blog);
```
ajcvickers, line 10: Consider blank lines between methods to improve readability.

entity-framework/core/testing/unit-testing.md
```diff
@@ -0,0 +1,60 @@
...
+
+[!code-csharp[Main](../../../samples/core/Testing/BusinessLogic/IBloggingRepository.cs?name=IBloggingRepository)]
+
+... and here's a partial sample implementation for production use:
+
+[!code-csharp[Main](../../../samples/core/Testing/BusinessLogic/BloggingRepository.cs?name=BloggingRepository)]
+
+There's not much to it: the repository simply wraps an EF Core context, and exposes methods which execute the database queries and updates on it. A key point to note is that our `GetAllBlogs` method returns `IEnumerable<Blog>`, and not `IQueryable<Blog>`. Returning the latter would mean that query operators can still be composed over the result, requiring EF Core to still involved in translating the query; this would defeat the purpose of having a repository in the first place. `IEnumerable<Blog>` allows us to easily stub or mock what the repository returns.
```
ajcvickers, line 22:
```suggestion
There's not much to it: the repository simply wraps an EF Core context, and exposes methods which execute the database queries and updates on it. A key point to note is that our `GetAllBlogs` method returns `IEnumerable<Blog>`, and not `IQueryable<Blog>`. Returning the latter would mean that query operators can still be composed over the result, requiring that EF Core is still involved in translating the query; this would defeat the purpose of having a repository in the first place. `IEnumerable<Blog>` allows us to easily stub or mock what the repository returns.
```

entity-framework/core/testing/unit-testing.md
```diff
@@ -0,0 +1,60 @@
...
+
+## SQLite in-memory
+
+SQLite can easily be configured as the EF Core provider for your test suite instead of your production database system (e.g. SQL Server); consult the [SQLite provider docs](xref:core/providers/sqlite/index) for details. However, it's usually a good idea to use SQLite's [in-memory database](https://sqlite.org/inmemorydb.html) feature when unit testing, since it provides easy isolation between tests, and does not require dealing with actual SQLite files.
+
+To use in-memory SQLite, it's important to understand that a new database is created whenever a low-level connection is opened, and that it's deleted what that connection is closed. In normal usage, EF Core's `DbContext` opens and closes database connections as needed - every time a query is executed - to avoid keeping connection for unnecessarily long times. However, with in-memory SQLite this would lead to resetting the database every time; so as a workaround, we open the connection before passing it to EF Core, and arrange for it to be closed only when the test completes:
```
ajcvickers, line 42: Has this behavior changed in any way due to pooling? /cc @bricelam

ajcvickers (approved):
Great stuff!

entity-framework/core/testing/choosing-a-testing-strategy.md
atrauzzi: Also worth mentioning that leaking `IQueryable` while sometimes frowned upon, allows querying to become a minor cross-cutting concern.  To me, this is actually quite powerful, especially given that `IQueryable` is part of the language itself, and so leaking it doesn't create any unwanted dependencies.

( https://chillicream.com/blog/2020/03/18/entity-framework )

entity-framework/core/testing/choosing-a-testing-strategy.md
atrauzzi: Agreed. Especially with how much the ASP.NET team is trying to simplify application boilerplate and bootstrap.

It would for sure be jarring to someone getting started with .NET server side development.  Consider from their perspective:  Going from a tutorial on ASP.NET telling them how they've lowered the ceremony and then arrive at Entity Framework docs which tell them to strap in for a lot of repository pattern boilerplate.

Most people just want to write LINQ.  The state of the art here is LINQ in mediator listeners/handlers.

entity-framework/core/testing/integration-testing.md
```diff
@@ -0,0 +1,134 @@
...
+
+> [!TIP]
+> This page shows [xUnit](https://xunit.net/) techniques, but similar concepts exist in other testing frameworks, including [NUnit](https://nunit.org/).
+
+## Setting up your database system
+
+Most database systems nowadays can be easily installed, both in CI environments and on developer machines. While it's frequently easy enough to install the database via the regular installation mechanism, ready-to-use Docker images are available for most major databases and can make installation particularly easy in CI. For the developer environment, [Github Workspaces](https://docs.github.com/en/codespaces/overview), [Dev Container](https://code.visualstudio.com/docs/remote/create-dev-container) can set up all needed services and dependencies - including the database. While this requires an initial investment in setup, once that's done you have a working integration testing enviropnment and can concentrate on more important things.
```
atrauzzi, line 19: > enviropnment

:heart:

entity-framework/core/testing/index.md
roji: I'm sure that will happen :) I did add "for simplicity's sake" to hopefully head that off, but hopefully it shouldn't be that controversial. It's certainly more practical than having to say "tests that involve your production database system" and "tests which replace your production database with a test double"...

If you feel really strongly about this, I can try changing, but we can see how people react and possibly change it later if needed.

entity-framework/core/testing/index.md
roji: OK.

Though if by "mocking DbSet" we mean making DbSet backed by an in-memory collection, and applying LINQ operators as Enumerable ones, then I did this very easily with one of the ready-made EF-specific mocking packages out there (don't remember which one). It really was easy and simple, but doesn't really get you much beyond what InMemory provides.

Otherwise, mocking DbSet could mean creating a whole other LINQ provider (non-EF!), which really doesn't seem worth discussing.

entity-framework/core/testing/index.md
roji: Corrected everywhere.

entity-framework/core/testing/choosing-a-testing-strategy.md
roji: Yeah, as I wrote above - I think we're jumping between two definitions of what it means to mock DbSet/query.

I think most people understand this to mean having a DbSet backed by an in-memory collection, and applying Enumerable operators to it (technically faking rather than mocking). That's easy to do, but not really more useful than the in-memory provider.

The 2nd meaning (use a mock/fake LINQ provider) is sort-of what we provide with the in-memory provider; sure, we provide a fake EF provider rather than non-EF LINQ provider, but that's quite similar. Obviously users writing their own LINQ provider makes no sense, and I'm not sure what benefits that would bring over our in-memory provider.

entity-framework/core/testing/choosing-a-testing-strategy.md
roji: @atrauzzi the moment you leak IQueryable, you lose the ability to stub/mock your repository layer - which at least in this context is the main reason for having a repository in the first place... I get what you're saying: if you have a complex query used in multiple places, and you want to compose further operators over it in those places, then yeah, you can wrap it in a method returning IQueryable for further composition. But if you want to preserve the ability to stub out your repository, each use of that query itself must be implemented in a method which *doesn't* return IQueryable, so that it can be stubbed/mocked.

Bottom line, I think this would be more confusing than helpful in this specific conversation, which is about testing/stubbing/mocking (GraphQL is a very specific case which again is somewhat orthogonal to testing).

roji: @ajcvickers I get what you're saying: it's indeed important to stress that testable repositories can't return IQueryable. But  I wouldn't say a repository pattern prevents composing complex queries with different result shapes, or any other composable operators (I think that's what you're saying). The basic limitation is that whatever API your product code uses must itself not return IQueryable; but a repository can have a private Queryable-returning method representing a complex query, and multiple public methods which compose over that (projections, filters, whatever) to return IEnumerable.

You could definitely argue that this is more laborious, as each data shape requires another IEnumerable-returning repository method, but that's just the regular price of the pattern (and different from saying that a repository prevents anything, no?)

I've integrated parts of your suggested text above, and hopefully made it clearer that the repository pattern has a significant implementation/maintenance cost - let me know what you think, am open for further changes.

entity-framework/core/testing/choosing-a-testing-strategy.md
roji: That's an extreme formulation :) I think we're trying to say that all LINQ must be encapsulated behind Enumerable-returning methods (which is what I tried to convey with "Requires special code architecture").

I've replaced "Requires special code architecture" (which was supposed to convey that) with "Can use LINQ anywhere in the application?", where the repository pattern is the only one where the answer is "no" (also added a footnote). Let me know if that seems sufficient.

entity-framework/core/testing/choosing-a-testing-strategy.md
roji: That's really strong - I tried to not be that categorical about rejecting unit testing. Can we compromise and remove the "absolutely"? :rofl:

entity-framework/core/testing/choosing-a-testing-strategy.md
roji: This statement is made within the context of the "Unit testing" section, but I agree and will tweak to make this more explicit and nuanced.

@atrauzzi I don't disagree, but that's the reality of the situation. If you (a) insist on doing unit tests (no database involved), and (b) do anything beyond a very restricted subset of queries, then the only way to do this reliably and comprehensively is via a repository layer. If that's unacceptable because it's too complicated/ceremonious, you can test against your real database - but within the above two constraints there's not much else we can say.

roji: Here's what I ended up with for now:

> Below, we'll explore what each method means, and compare it with the others; We recommend reading through the different methods to gain a full understanding of each one. If you've decided that unit tests are required (as opposed to integration tests), than a repository layer is the only approach allowing the comprehensive and reliable stubbing/mocking of the data layer. However, that approach has a significant cost in terms of implementation and maintenance.

roji:
@ajcvickers pushed fixes and tweaks - as there are some unresolved discussions above, we can briefly discuss tomorrow during the day and merge in the evening?

entity-framework/core/testing/choosing-a-testing-strategy.md
atrauzzi: Makes sense, and I've come to agree. Using the real database is essential without a repository layer. 👍

ajcvickers:
I think the fundamental problem here is that when I read this document end-to-end and then use principled reasoning, I come to the conclusion that I _must use a repository layer_. More specifically:

* Principle: Unit tests are extremely important and I must have them.
* Principle: I want to follow best-practice guidance from the team.
* Assertion from this document: To unit test an EF Core application, it is highly recommended I use a repository layer.
* Conclusion: I must use a repository layer in my EF Core application.

But I don't think this is the correct conclusion. For most applications, I don't think adding a repository layer is necessary or beneficial. So either I'm wrong about this, or this document is leading me to the wrong conclusion. I think it would be a good idea to discuss with @bricelam @AndriySvyryd, @maumar, @smitpatel, and @JeremyLikness whether or not my view (and my understanding of the team view) is wrong. But assuming I am not wrong, then why does this document lead me to the wrong conclusion? Some possibilities:

1. My principle is at odds with the intent of this document. Specifically, I would use a real database for my tests, but I don't know that I would call those tests "integration tests". This is why I am somewhat uncomfortable using the terms "integration testing" and "unit testing". They are highly-charged and yet ambiguous terms, and I don't think using them here necessarily results in clarity.
2. My principle is wrong. What I really mean is, "~~Unit~~ Tests are extremely important and I must have them." This is essentially the same as the first point, but the principle I stated _including the word "unit"_ is very widely and very strongly held, and we're not going to change people's mind on it.
3. My principle is too strong. I need unit tests for everything _except_ interactions with the database. Again, a hard sell for a widely and strongly held belief.

So, fundamentally, _if_ we're going to make the distinction between "integration testing" and "unit testing", then I think we have to make it absolutely clear that we're saying what we define as "unit testing" is not something that is always needed. We have to be very strong on this point, because we're going against a widely and strongly held belief.

This means we could resolve this by simply removing the distinction and labeling of the different kinds of tests. This will leave the door open for questions like, "but how do I unit test?" I think this is okay; the answer is that you don't need to, or that, if you prefer, you can still think of your tests as unit tests even though they use a real database. But, if we are going to make this distinction, then we also have to be absolutely clear that we are _not_, in general, recommending the repository pattern.

roji:
OK, it seems like the best course of action is to get rid of the terms "unit" and "integration" tests. I'll change the text.

> Re: the LINQ/IQueryable issue. When we talk about a repository pattern, what we are talking about is not exposing LINQ/IQueryable to most of the application. I think it's very important that we emphasize this because:

I agree, and I think that's already mentioned both in the [testing strategies page](https://github.com/dotnet/EntityFramework.Docs/pull/3539/files#diff-7d43ce68bd8bb4144e37aba924a13e3899cc5761fa173dc7e7637f14a435072dR87) and in the [implementation page](https://github.com/dotnet/EntityFramework.Docs/pull/3539/files#diff-2ee94baef85c45fbd2587d5000661c9be1dd16965f91e0cbfb84dc899b9de168R22). If this still seems insufficient to you I can make this into a more sizable (and scary) paragraph.

entity-framework/core/testing/index.md
ajcvickers: What I am saying is that writing a test double that understands/interprets the parameter (i.e. the expression tree) and then returns appropriate test data based on that is very complex/hard. You say this isn't worth discussing, but we have had people try to do just that. (I tried to do it many years ago!)

entity-framework/core/testing/choosing-a-testing-strategy.md
ajcvickers: > Users writing their own LINQ provider makes no sense

Exactly.

> Obviously.

To you... :-D

entity-framework/core/testing/choosing-a-testing-strategy.md
ajcvickers: @atrauzzi I'm in favor of using IQueryable throughout application code. Hence I'm not in favor of the repository pattern. Something that looks like a repository but exposes IQueryable...is DbContext/DbSet. :-)

@roji

> But I wouldn't say a repository pattern prevents composing complex queries with different result shapes, or any other composable operators

I'm explicitly not saying this. I'm saying you have to write and maintain code in the repository to do this. There are good ways to do this. However, they are not IQueryable, and are (necessarily and by-design) more restrictive than IQueryable. This is what you lose/gain from having the repository; not the ability to have parameterized/extendable queries.

ajcvickers (approved):
Perfect! With the change not to say "unit" or "integration" I think this is now excellent and clear guidance. Bravo!

roji:
Thank you! I even grudgingly agree that this will head off some worthless discussions and possible confusion.

PR #3549, opened 2021-11-10
Correct SQL in many-to-many scaffold sample

simon-reynolds:
Constraints on PostTag where referencing wrong tables based on name
Posts table also included a reference to Blog table not listed in example


ajcvickers (approved):

ajcvickers:
Thanks!

PR #3558, opened 2021-11-14
Fix typo in core/index.md

oreze:
fix typo (i with diaeresis)


roji:
This doesn't seem like an error: naive and `naïve` are both OK, no? In fact, `naïve` is slightly move accurate, since this French-originating word isn't pronounced according to standard rules (that would sound like `knave`).

PR #3568, opened 2021-11-16
Basic Save title does not encompass information provided

justinbourb:
This document covers create, save and delete.  Basic save is a misleading title.

### DIFF
--- a/entity-framework/core/saving/basic.md
+++ b/entity-framework/core/saving/basic.md
@@ -1,11 +1,11 @@
 ---
-title: Basic Save - EF Core
+title: Create Update Delete - EF Core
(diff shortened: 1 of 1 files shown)


roji:
@justinbourb "save" is used to encompass all modification kinds, including adding, modifying and removing data (as is explained in the first sentence of the page). This also corresponds to the EF Core API which is used to persist (save) all scheduled changes - SaveChanges.

ajcvickers:
@roji Agreed, although Create, Update, and Delete might be clearer?

@JeremyLikness?

ajcvickers:
Note from triage: move Basic Save section into Overview, which is currently not useful.

See https://github.com/dotnet/EntityFramework.Docs/issues/3629

PR #3573, opened 2021-11-19
Fix typo in disconnected-entities.md

nbarbettini:
It looks like this was meant to read "implemented **by** using query filters" or simply "implemented using query filters". I opted to fix it to the latter.


roji (approved):
Thanks!

PR #3580, opened 2021-11-24
Add documentation for EF6 value converter

vanillajonathan:
https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-6.0/whatsnew#pre-convention-model-configuration


roji:
Duplicate of #3278

PR #3584, opened 2021-11-26
Update TemporalAll method description

VahidFarahmandian:
TemporalAll method returns historical data and the current state of each record.


ajcvickers:
@VahidFarahmandian I'm not sure this makes things clearer, even though it is technically correct. If we are going to go in this direction, then all the LINQ operator descriptions should be changed to reflect this, or possibly the whole section should be re-worded differently.

VahidFarahmandian:
> @VahidFarahmandian I'm not sure this makes things clearer, even though it is technically correct. If we are going to go in this direction, then all the LINQ operator descriptions should be changed to reflect this, or possibly the whole section should be re-worded differently.

@ajcvickers what if we add our definition for "historical data" term? Because while reading this document, the reader faced with HistoryTable and its usage, and then he/she faced with the term "historical data", which at first glance it seems to be the data inside History table. By providing our definition about "historical data" we can at least provide a better clarification and description for the given methods.

ajcvickers:
@VahidFarahmandian Sounds reasonable.

ajcvickers:
Superseded by #3593.

PR #3607, opened 2021-12-10
Fix CosmosModelConfigurationSample

okyrylchuk:
Hi,
I believe that in the CosmosModelConfigurationSample should be the HasManualThroughput method instead of two HasAutoscaleThroughput methods.
If it's not correct, please reject the PR.


ajcvickers (approved):

ajcvickers:
Thanks!

PR #3621, opened 2021-12-14
Fix links from Breaking changes to What's new

danielkraut:
Links which did not use xref were broken (lines 136 and 159). Line 103 worked but I wanted to unify it with the rest. There are two options:

1. use xref - longer path, doesnt require .md suffix
2. use just file name - required .md suffix to work in github

I went with option n.2. Not adding the .md suffix would probably work on docs.microsoft.com website but doesn't work when browsing on github.

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-6.0/breaking-changes.md
+++ b/entity-framework/core/what-is-new/ef-core-6.0/breaking-changes.md
@@ -133,7 +133,7 @@ Instead of assigning the same owned instance to a new owner you can assign a cop
 ### Cosmos: Related entity types are discovered as owned
 
 [Tracking Issue #24803](https://github.com/dotnet/efcore/issues/24803)
-[What's new: Default to implicit ownership](whatsnew.md#default-to-implicit-ownership)
+[What's new: Default to implicit ownership](xref:core/what-is-new/ef-core-6.0/whatsnew#default-to-implicit-ownership)
 
 #### Old behavior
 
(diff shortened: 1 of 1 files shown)


ajcvickers:
@smitpatel Assigning to you as the 🔗Link Meister 🔗.

entity-framework/core/what-is-new/ef-core-6.0/breaking-changes.md
```diff
@@ -133,7 +133,7 @@ Instead of assigning the same owned instance to a new owner you can assign a cop
 ### Cosmos: Related entity types are discovered as owned
 
 [Tracking Issue #24803](https://github.com/dotnet/efcore/issues/24803)
-[What's new: Default to implicit ownership](/core/what-is-new/ef-core-6.0/whatsnew#default-to-implicit-ownership)
+[What's new: Default to implicit ownership](whatsnew.md#default-to-implicit-ownership)
```
smitpatel, line 136: Should use xref: syntax in all 3.

danielkraut: I changed links to xref syntax. Altough I have to say that I don't understand the reason why use xref syntax when the other syntax is usually chosen in markdown. And article [How to use links in documentation](https://docs.microsoft.com/en-us/contribute/how-to-write-links#xref-cross-reference-links) says that xref links are recommended when linking to APIs. And xref links don't work when browsing markdowns directly on github.

smitpatel: They are validated at runtime and allow to reference other docsets. Also they are protected against file renames. We have decided to use xref syntax rather than ".md" files specifically for the latter. A lot of things in the doc page cannot be rendered in markdown mode (API docs/ code samples/ tabs), so don't see a reason to try hard to preserve these links.

smitpatel (approved):

PR #3632, opened 2021-12-17
Add multitenancy

### LINKED ISSUE #3589: Port multi-tenancy guidance to docs
Add a section addressing [this blog post](https://blog.jeremylikness.com/blog/multitenancy-with-ef-core-in-blazor-server-apps/) to the docs.

JeremyLikness:
Add document describing how to support multi-tenant databases. Fixes #3589.


entity-framework/core/miscellaneous/multitenancy.md
```diff
@@ -0,0 +1,118 @@
...
+# Multi-tenancy
+
+Many line of business applications are designed to work with multiple customers. It is important to secure the data so that customer data isn't "leaked" or seen by other customers and potential competitors. These applications are classified as "multi-tenant" because each customer is considered a tenant of the application with their own set of data.
+
+> [!IMPORTANT]
+> This document provides examples and solutions "as is." These are not intended to be "best practices" but rather "working practices" for your consideration.
+
+## Supporting multi-tenancy
+
+There are many approaches to implementing multi-tenancy in applications. One common approach (that is sometimes a requirement) is to keep data for each customer in a separate database. The schema is the same but the data is customer-specific. Another approach is to partition the data in an existing database by customer.
```
roji, line 18: When partitioning multiple tenants in the same database, there's the single-schema approach with a tenant ID (which is where global query filters come into play), but there's also the schema-per-tenant approach, where schemas can also diverge between tenants (big headache). It may be worth at least making that distinction and clarify which one we're talking about here, etc.

Ultimately we may want multiple nested pages showing each method: multiple databases, multiple schemas, single schema with tenant ID; it may be slightly confusing to have the two approaches described below in the same page. But of course we don't need to cover everything right away (as it's a bit subject), I'd be happy to contribute to this as well in the future.

JeremyLikness: Great feedback. I added an example.

roji:
BTW consider linking to this page from [the global query filters doc page](https://docs.microsoft.com/en-us/ef/core/querying/filters).

entity-framework/core/miscellaneous/multitenancy.md
```diff
@@ -0,0 +1,148 @@
...
+|**Scenario**|**Single database**|**Multiple databases**|
+|:--|:--|:--|
+|_User stays in a single tenant_|`Scoped`|`Scoped`|
+|_User can switch tenants_|`Scoped`|`Transient`|
+
+The default of `Singleton` still makes sense if your database does not take on user-scoped dependencies.
+
+## Performance notes
+
+It is highly unlikely that a single user session will require hundreds or thousands of `DbContext` instances during a given session. In local benchmarks it only takes 3 microseconds on a laptop to go from requesting the factory to creating a usable `DbContext`. That translates to over 300,000 requests per second.
```
roji, line 144: If it really does take only 3 microseconds, I'll bet that's the case only with DbContext pooling - or did you see something else?

In any case, it may be good to clarify what exactly you're advising users, and how that's specific to multitenancy.

roji (commented):
Hey @JeremyLikness, here are some comments.

JeremyLikness:
@roji I added the samples and will do the document updates once we are agreed the samples are correct. Please let me know. /cc @dotnet/efteam

samples/core/Multitenancy/exampleschema/Program.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        public ContactContext(
+            TenantProvider provider,
+            DbContextOptions<ContactContext> opts)
+            : base(opts) => _tenant = provider.Tenant;
+
+        public DbSet<TenantAContact> TenantAContacts { get; set; } = null!;
+        public DbSet<TenantBContact> TenantBContacts { get; set; } = null!;
+
+        public IQueryable<Contact> Contacts => _tenant == "TenantA" ?
```
roji, line 125: I'm not too sure about this approach... This requires modifying application code each time a tenant is added to the system, which seems problematic.

It's probably possible to do something fancy, i.e. get the full tenant list from some config file and then dynamically configure all the right tables in the context - but that's more complex than what's being shown here...

Are we sure we want to show this? The schema-per-tenant approach below is probably preferable in almost all cases anyway (unless maybe when there's no schemas like in SQLite)

samples/core/Multitenancy/exampleschema/Program.cs
```diff
@@ -0,0 +1,182 @@
...
+        public AlternateContactContext(
+            TenantProvider provider,
+            DbContextOptions<AlternateContactContext> opts)
+            : base(opts) => Tenant = provider.Tenant;
+
+        public DbSet<Contact> Contacts { get; set; } = null!;
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+         =>
+            // SQLite doesn't handle multiple schemas, so we use different
```
roji, line 158: Yeah, I'd definitely do this sample on SQL Server (LocalDB) so it actually shows what we want to show...

samples/core/Multitenancy/exampleschema/Program.cs
```diff
@@ -0,0 +1,182 @@
...
+            // table names but the same API can be used with SQLServer for
+            // same table name, different schema
+            modelBuilder.Entity<Contact>().ToTable($"{Tenant}Contacts");
+
+        /// <summary>
+        /// Tell EF Core to cache the model by tenant.
+        /// </summary>
+        /// <param name="optionsBuilder">The options builder.</param>
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        => optionsBuilder
```
roji, line 168: indentation

samples/core/Multitenancy/exampleschema/Program.cs
```diff
@@ -0,0 +1,182 @@
...
+            .ReplaceService<IModelCacheKeyFactory, TenantModelCacheKeyFactory>();
+    }
+
+    /// <summary>
+    /// Implemented to support a model per tenant in the same DbContext.
+    /// </summary>
+    public class TenantModelCacheKeyFactory : IModelCacheKeyFactory
```
roji, line 175: See the new EF Core 6.0 note in the [docs](https://docs.microsoft.com/en-us/ef/core/modeling/dynamic-model#imodelcachekeyfactory) on the design-time bit.

samples/core/Multitenancy/examplesingle/Program.cs
```diff
@@ -0,0 +1,123 @@
+﻿// See https://aka.ms/new-console-template for more information
+using ExampleSingle;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+
+Console.WriteLine("Single database and table example.");
+
+// grab some types as sample data
+#pragma warning disable CS8601 // Where clause guarantees non-nullable
+var types = AppDomain.CurrentDomain.GetAssemblies()
```
roji, line 10: Uhh... This seems like it goes over all loaded assemblies and create MultitenantTypes for all types which have have a namespace - that's a whole lot of types :) You'd probably want to narrow that down via some sort of attribute or interface - but even then I'd recommend against this kind of assembly exploration (especially of the entire domain).

If what we're looking for is a full list of tenants (though not sure that's a good idea), that should probably come from user config.

samples/core/Multitenancy/examplesingle/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+    .ToList();
+#pragma warning restore CS8601 // Possible null reference assignment.
+
+// unique "tenants" = namespaces
+var tenants = types.Select(t => t.Tenant).Distinct()
+    .OrderBy(t => t).ToList<string>();
+
+var provider = new TenantProvider { Tenant = tenants[0] };
+
+// using this so the sample is relevant to ASP.NET Core apps
```
roji, line 28: As above, we may just want to do an ASP.NET Core sample then

samples/core/Multitenancy/examplesingle/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        public SingleTenantContext(
+            DbContextOptions<SingleTenantContext> opts,
+            TenantProvider provider)
+            : base(opts) => _tenant = provider.Tenant;
+        public DbSet<MultitenantType> MultitenantTypes { get; set; } = null!;
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+            => modelBuilder.Entity<MultitenantType>()
+                    .HasQueryFilter(mt => mt.Tenant == _tenant);
```
roji, line 121: OK, so this is the query filter sample. In that case I'd rename the project from `examplesingle` to something like `QueryFilterMultitenancy`.

roji, line 121: indentation

samples/core/Multitenancy/examplesingle/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+{
+    /// <summary>
+    /// Simple tenant service.
+    /// </summary>
+    public class TenantProvider
+    {
+        public string Tenant { get; set; } = string.Empty;
+    }
+
+    public class MultitenantType
```
roji, line 101: Maybe just use Contact like in the other sample? MultitenantType is slightly confusing, and it's good to be consistent across these samples so users can easily compare the alternatives

samples/core/Multitenancy/examplesingle/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+{
+    ctx.Database.EnsureDeleted();
+    ctx.Database.EnsureCreated();
+    ctx.MultitenantTypes.AddRange(types);
+    ctx.SaveChanges();
+}
+
+var done = false;
+
+while (!done)
```
roji, line 54: I'm not sure the loop with the int parsing helps - maybe keep it minimal with a quick code sample like in the other samples?

roji (changes requested):
Hey @JeremyLikness, here's a first round of comments.

JeremyLikness:
@roji I updated the samples, there are two (single db, multi-db). Please review and I'll update the markdown when the code is ready. Thanks!

JeremyLikness:
Made a pass at all recommendations. This is ready for review.

bricelam (approved):
LGTM. It's great to finally see this guidance finally included in our docs!

roji (approved):
sheep eeeeeet

PR #3635, opened 2021-12-18
Document UseQueryTrackingBehavior

### LINKED ISSUE #3634: Document UseQueryTrackingBehavior

roji:
Closes #3634

### DIFF
--- a/entity-framework/core/querying/tracking.md
+++ b/entity-framework/core/querying/tracking.md
@@ -43,9 +43,11 @@ Starting with EF Core 5.0, you can combine both of the above behaviors in same q
 
 ## Configuring the default tracking behavior
 
-In some scenarios, most (or even all) queries in your application should be non-tracking queries. Rather than adding `AsNoTracking()` on each and every query in the codebase, you can configure the default tracking behavior at the context level:
+If you find yourself changing the tracking behavior for many queries, you may want to change the default instead:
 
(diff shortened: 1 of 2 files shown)


entity-framework/core/querying/tracking.md
```diff
@@ -41,6 +41,12 @@ Starting with EF Core 5.0, you can combine both of the above behaviors in same q
 
 [!code-csharp[Main](../../../samples/core/Querying/Tracking/Program.cs#NoTrackingWithIdentityResolution)]
 
+## Configuring the default tracking behavior
+
+In some scenarios, most (or even all) queries in your application should be non-tracking queries. Rather than adding `AsNoTracking()` on each and every query in the codebase, you can configure the default tracking behavior at the context level:
```
smitpatel, line 46: > In some scenarios, most (or even all) queries in your application should be non-tracking queries.

I find the statement putting out an opinion without any details why so. Further detail of the behavior doesn't belong in this API. If you want to write about why queries should be non-tracking then utilize some other space. This section should only describe how to specify default behavior and all the tracking behavior can be specified as default.

roji: Would "need to be" be better than "should be"?

smitpatel: Already answered above.

roji: Take a look at my latest commit, what do you think?

smitpatel (approved):

PR #3643, opened 2021-12-29
Fix command execution on wrong table

virzak:
Also this code is never triggered from the Main(). Not sure if this is by design.


roji (approved):
Thanks!

Issue #681, opened 2018-04-16
Improve documentation on how to write an EF Core provider 


divega:
Currently we have something very brief at https://docs.microsoft.com/ef/core/providers/writing-a-provider pointing to @ajcvickers's blog.  

It would be good to update this to cover:
- Support for EF Core 2.1 features
- The mechanics of setting up spec tests (e.g. https://github.com/dotnet/efcore/issues/27107#issuecomment-1004923695)
- The mechanics of adding provider specific extension methods

Whether this should be in a blog post or if it needs to be part of the documentation is something we can discuss, but I would be inclined to have it in the actual documentation.

TomasHubelbauer:
I'd like to express my preference for including the guide on how to create one's own EF Core database provider in the documentation as well. It's way more authoritative than a blog post. I'm just trying to figure out going by the old version of the docs, can you provide a rundown of the changes or a super rough estimate for when 3.1.0 might land?

omatrot:
I'd like to write a scaffolding only provider around edmx files in order to fill the gap for all of us having existing edmx files that we want to keep around in order to create entities and DbContexts.

Any information that could allow me to start would be very much appreciated.

ajcvickers:
Suggestion from @ErikEJ on https://github.com/dotnet/efcore/issues/20793#issuecomment-621943308

> It would lower the barrier to have a "template/null" provider to use as a starting point. But of course it can quickly become stale.

roji:
Ideally this would be built as part of CI, to keep it in at least minimal sync with EF...

ErikEJ:
@roji do you think there would be any value in this?

roji:
Yeah, I think it would... Our docs repo has a set of samples - which now even get built in CI - so it could live there. On the other hand, if we put it in the normal repo it would naturally evolve as we change EF Core...

ErikEJ:
I think it is better to fix it to depend on a released version of EF Core, and then just update on new release (if needed at all)

smitpatel:
Releases to consider
- LTS release
- Current release (stable)
- Current preview

I think current preview would provide slightly lesser value. If I am writing a provider from scratch, I would want to depend on stable release so I can build my provider and release it. It would be easy after 3.1 to move to next major release for providers even with breaking changes. The issue with targeting previews would be constantly moving target. If EF Core added some API in one preview but changed it in another preview (we love API reviews), then provider has unnecessary effort to keep changing things when writing the core provider at the same time.

[...]

PR #3644, opened 2021-12-29
Add plugin sample

virzak:


samples/core/SqlServer/Plugin/PluginSetup.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        public override void PopulateDebugInfo(IDictionary<string, string> debugInfo)
+            => debugInfo["SqlServer:" + nameof(SqlServerExtensions.UseAugmentation)] = "1";
+
+        public override string LogFragment
+            => "using Augmentation";
+    }
+}
+
+public static class SqlServerNetTopologySuiteServiceCollectionExtensions
```
roji, line 47: The type name seems wrong

samples/core/SqlServer/Plugin/DbFunctionsExtensions.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+
+namespace SqlServer.Plugin;
+public static class DbFunctionsExtensions
```
roji, line 6: The class should be named something like AugmentationDbFunctionsExtensions

roji (commented):
Thanks for this PR.

We have plans to improve documentation on writing providers (#681, for [this page](https://docs.microsoft.com/en-us/ef/core/providers/writing-a-provider)), we could also have a page on how to write a plugin.

We usually add samples with the corresponding conceptual documentation; a code sample alone (such as this) isn't easily discoverable by users looking at the docs. If the goal is simply to have a working plugin source to look at, then simply pointing to e.g. the NetTopologySuite source code may be enough.

Note also that  this sample shows only IMethodCallTranslatorPlugin, but there are other plugin types (e.g. ITypeMappingSourcePlugin, IEvaluatableExpressionFilterPlugin, IConventionSetPlugin...).

samples/core/SqlServer/Plugin/PluginSetup.cs
virzak: Fixed

samples/core/SqlServer/Plugin/DbFunctionsExtensions.cs
virzak: Fixed

virzak:
Hi @roji,

This PR was started in an attempt to reproduce a possible bug in EF Core PawelGerr/Thinktecture.EntityFrameworkCore#17

I didn't find an existing test case to modify in [dotnet/efcore](https://github.com/dotnet/efcore), because I'm either not familiar with it or because it wasn't there.

Still hoping to reproduce the issue, but in the meantime the PR was a good exercise for me to isolate the plugin concept. If the PR gets merged and helps someone else - that's a bonus.

virzak:
Issue dotnet/efcore#27075 was the original reason for this PR

roji:
@virzak we've discussed this internally, and the decision is that we prefer not to add standalone code samples that aren't part of corresponding conceptual documentation. In other words, something like this would definitely make sense as part of #681, but we don't think that having the code without any docs is very helpful.

Good luck with your bug and let us know if you need any help.

virzak:
@roji Yeah, that's cool. You mentioned that you're pointing developers to NetTopologySuite source code. Perhaps a plugin/README.md could have that link until #681 is complete. It would make sense for me, as I am who tends to look at the source code a lot more than reading documentation.

roji:
I'm not sure a plugin/README.md would be very discoverable either - I think at some point we need to do #681...

PR #3655, opened 2022-01-04
Update ControllingTransaction.cs

riddlemd:
Fixed typo, DatabaseFacade does not have a Commit method, it has a CommitTransaction method.


roji:
@riddlemd the call is to Commit() on IDbContextTransaction, which is returned from BeginTransaction() above, and not on DatabaseFacade. Your change doesn't compile.

PR #3656, opened 2022-01-04
Remove DbContext Pooling from Miscellaneous TOC

roji:
DbContext Pooling used to be a page under Miscellaneous, but was moved to the Advanced Performance Topics page under Performance. A redirection is already in place, but the old TOC entry was left in place.

### DIFF
--- a/entity-framework/toc.yml
+++ b/entity-framework/toc.yml
@@ -112,6 +112,8 @@
       items:
       - name: Overview
         href: core/dbcontext-configuration/index.md
+      - name: Context pooling
+        href: core/performance/advanced-performance-topics#dbcontext-pooling
 
     - name: Create a model
       items:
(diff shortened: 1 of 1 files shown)


ajcvickers:
Maybe we should leave it in both places? For discoverability.

roji:
I guess... I like to leave Miscellaneous as small as possible, and DbContext pooling really is a perf feature. If anything, it should probably be mentioned somewhere under "DbContext Lifetime, Configuration, and Initialization"...

ajcvickers:
I like putting it under  "DbContext Lifetime, Configuration, and Initialization"

roji:
I can do that. What do you think, an additional short section in the existing overview page? Because there's only that one page in the section, it would be weird to have a second DbContext Pooling TOC item...

ajcvickers:
There's going to be many more pages. As soon as I write them. :-) I think just move the second TOC item there.

roji:
@ajcvickers done

ajcvickers (approved):

PR #3669, opened 2022-01-10
Update supported versions of EFCore for EFCore.SoftDelete package.

AshkanAbd:
In recent version of EFCore.SoftDelete (1.3.x), I added dotnet 6 support.


roji (approved):
Thanks.

Issue #3671, opened 2022-01-11
Invalid Links - missing `/ef/` prefix in produced href


Kissaki:
The following links are invalid:

* [porting EDMX](https://docs.microsoft.com/en-us/efcore-and-ef6/porting/port-edmx#other-considerations) (missing `/ef/` prefix)
* [Review behavior changes between EF6 and EF Core](https://docs.microsoft.com/en-us/efcore-and-ef6/porting/port-behavior) (missing `/ef/` prefix)
* [here](https://docs.microsoft.com/en-us/efcore-and-ef6/porting/port-detailed-cases#required-and-optional) (missing `/ef/` prefix)
* [port from EDMX](https://docs.microsoft.com/en-us/efcore-and-ef6/porting/port-edmx) (missing `/ef/` prefix)
* Code as source of truth
* Database as source of truth
* Hybrid model
* detailed differences between EF6 and EF Core

The sources link path do have the `/ef/` prefix, like other links that work on this page. So I do not see what is wrong.

I also tried running it locally with DocFX, but could not really work with it or find out more. (missing CSS, wrong href paths)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9cbb86f3-378b-e83d-56c3-c0532192bd13
* Version Independent ID: c96b2fcb-8af0-f05b-e188-a6c59dd9f428
* Content: [Port from EF6 to EF Core - EF](https://docs.microsoft.com/en-us/ef/efcore-and-ef6/porting/)
* Content Source: [entity-framework/efcore-and-ef6/porting/index.md](https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/efcore-and-ef6/porting/index.md)
* Product: **entity-framework**
* Technology: **entity-framework-core**
* GitHub Login: @JeremyLikness
* Microsoft Alias: **avickers**

Kissaki:
Subpages under this node have the same problem.

ajcvickers:
Fixed by #3672.

PR #3672, opened 2022-01-11
Fix broken links in porting guide

JeremyLikness:
Includes conversion from relative path to xref: for better validation.


JeremyLikness:
Will fix #3671

AndriySvyryd (approved):

PR #3681, opened 2022-01-16
using System.ComponentModel.DataAnnotations.Schema

yecril71pl:
This is the first and only place in the hand-out where `using` declarations are explicitly given.
`[Table]` attribute is used later on, so I added the namespace for it here.
I also added the `[Table]` attribute, otherwise the namespace would be unneeded for the given snippet.

### DIFF
--- a/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
+++ b/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
@@ -1,6 +1,3 @@
-using System.ComponentModel.DataAnnotations;
-using Microsoft.EntityFrameworkCore;
-
 namespace EFModeling.EntityProperties.DataAnnotations.Required
 {
     internal class MyContext : DbContext
--- a/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
+++ b/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
@@ -8,7 +8,7 @@ namespace EFModeling.EntityProperties.DataAnnotations.Required
         public DbSet<Blog> Blogs { get; set; }
     }
 
-#region Required
+    #region Required
     public class Blog
     {
         public int BlogId { get; set; }
@@ -16,5 +16,5 @@ namespace EFModeling.EntityProperties.DataAnnotations.Required
         [Required]
         public string Url { get; set; }
     }
-#endregion
+    #endregion
 }
--- a/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
+++ b/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
@@ -1,0 +1,18 @@
+using System.ComponentModel.DataAnnotations;
+using Microsoft.EntityFrameworkCore;
+
+namespace EFModeling.EntityProperties.DataAnnotations.Required
+{
+    internal class MyContext : DbContext
+    {
+        public DbSet<Blog> Blogs { get; set; }
+    }
+
+    public class Blog
+    {
+        public int BlogId { get; set; }
+
+        [Required]
+        public string Url { get; set; }
+    }
+}
--- a/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
+++ b/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
@@ -1,4 +1,10 @@
-﻿using System.ComponentModel.DataAnnotations;
+// RequiredAttribute
+using System.ComponentModel.DataAnnotations;
+
+// TableAttribute
+using System.ComponentModel.DataAnnotations.Schema;
+
+// DbContext
 using Microsoft.EntityFrameworkCore;
 
 namespace EFModeling.EntityProperties.DataAnnotations.Required
@@ -9,6 +15,7 @@ namespace EFModeling.EntityProperties.DataAnnotations.Required
     }
 
     #region Required
+    [Table("Blogs")]
     public class Blog
     {
         public int BlogId { get; set; }
--- a/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
+++ b/samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
@@ -1,3 +1,6 @@
+using System.ComponentModel.DataAnnotations;
+using Microsoft.EntityFrameworkCore;
+
 namespace EFModeling.EntityProperties.DataAnnotations.Required
 {
     internal class MyContext : DbContext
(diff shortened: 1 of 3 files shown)


ajcvickers:
@yecril71pl Why is `Table` needed for this code?

yecril71pl:
> @yecril71pl Why is `Table` needed for this code?

This code shows, in particular, how to import annotations for entities using `using`.  There are two namespaces we should be `using` for annotations.  Importing a namespace requires adding some namespace member to the code for exposition (this is sample code, not real code).  I chose `Table` as an example member.

roji:
@yecril71pl the point of this source file is to demonstrate the use of the `[Required]` attribute, nothing more. There's already an example for using the `[Table]` attribute in [TableName.cs](https://github.com/dotnet/EntityFramework.Docs/blob/main/samples/core/Modeling/EntityTypes/DataAnnotations/TableName.cs).

yecril71pl:
> @yecril71pl the point of this source file is to demonstrate the use of the `[Required]` attribute, nothing more. There's already an example for using the `[Table]` attribute in [TableName.cs](https://github.com/dotnet/EntityFramework.Docs/blob/main/samples/core/Modeling/EntityTypes/DataAnnotations/TableName.cs).

That example does not contain a `using` statement, nor does any example other than the first.  And rightfully so because repeating `using` statements in every sample would just be noise. It can reasonably be assumed that all relevant `using` statements should be in the 1st sample.

roji:
@yecril71pl maybe I'm missing something, but how is Required.cs special as compared to, say, TableName.cs? Both show annotations, both have `using`.

As you've stated, we don't include `using` directives in the quoted code samples in our conceptual docs; in other words, [this section](https://docs.microsoft.com/en-us/ef/core/modeling/entity-properties?tabs=data-annotations%2Cwithout-nrt#explicit-configuration) does not show any `using`.

So I'm still not understanding what this proposed change improves etc.

yecril71pl:
> @yecril71pl maybe I'm missing something, but how is Required.cs special as compared to, say, TableName.cs? Both show annotations, both have `using`.
> 
> As you've stated, we don't include `using` directives in the quoted code samples in our conceptual docs; in other words, [this section](https://docs.microsoft.com/en-us/ef/core/modeling/entity-properties?tabs=data-annotations%2Cwithout-nrt#explicit-configuration) does not show any `using`.

[That one](https://docs.microsoft.com/en-us/ef/core/modeling/#use-fluent-api-to-configure-a-model), however, does.  We have to put the necessary `using` statements somewhere and the 1st sample is the best place to do it.

roji:
Right - that's the overview. And the [data annotations sample just below](https://docs.microsoft.com/en-us/ef/core/modeling/#use-data-annotations-to-configure-a-model) shows the appropriate using directive for the [Required] attribute being shown. The fact that [Table] is in another namespace doesn't mean we need to show that - there may be other attributes (or Fluent APIs) which require yet other namespaces.

yecril71pl:
> Right - that's the overview. And the [data annotations sample just below](https://docs.microsoft.com/en-us/ef/core/modeling/#use-data-annotations-to-configure-a-model) shows the appropriate using directive for the [Required] attribute being shown. The fact that [Table] is in another namespace doesn't mean we need to show that - there may be other attributes (or Fluent APIs) which require yet other namespaces.

We need to show that because we explicitly use the `[Table]` attribute later on.

roji:
@yecril71pl we show many APIs throughout our docs, without necessary showing any using directives first; users just figure it out with the help of their IDEs, which suggest the proper using statements. If we go down this path we have to start putting namespaces in all samples, which would be a bad idea.

yecril71pl:
> @yecril71pl we show many APIs throughout our docs, without necessary showing any using directives first; users just figure it out with the help of their IDEs, which suggest the proper using statements. If we go down this path we have to start putting namespaces in all samples, which would be a bad idea.

Yes, having the necessary statements in the 1st sample is both necessary and sufficient.  And do not get me started on how unhelpful VS Code on Linux is.

ajcvickers:
The problem here is that the same code example is used for both the [general "Data Annotations" section](https://docs.microsoft.com/en-us/ef/core/modeling/#use-data-annotations-to-configure-a-model) and the [section specifically about "Required"](https://docs.microsoft.com/en-us/ef/core/modeling/entity-properties?tabs=data-annotations%2Cwithout-nrt#explicit-configuration). I think it does make sense to show the `Table` attribute and `Schema` namespace in the general section, it certainly is not appropriate in the `Required` section.

So, to fix this, there needs to be a separate example for the general section, and the namesapce and attribute need to be added to that section only. Also, there should not be any comments on the namespace declarations, and the region for `Required` should be removed.

samples/core/Modeling/EntityProperties/DataAnnotations/Required.cs
```diff
@@ -8,13 +8,11 @@ internal class MyContext : DbContext
         public DbSet<Blog> Blogs { get; set; }
     }
 
-    #region Required
```
ajcvickers, line 11: I didn't mean to remove it here. I meant it should not be in the new sample. It is required here (no pun intended) to define which part of the code to show in the docs.

smitpatel:
We have discussed several times in the past about showing using statements in our code samples. (Search github issue tracker with query "using namespace"). We arrived at a conclusion that using statements causes code samples to expand too much compared to value they provide. Plus with all modern IDEs, it is easy to import required using when copy pasting code. For non-IDE users, there is [.NET API browser](https://docs.microsoft.com/dotnet/api/) to figure out where does the class exist if not already known. I haven't seen any information above which requires us to change our decision.

As for the particular page which is being edited here, it is the very basic page of model building explaining concept of ways to create/configure model which doesn't have correlation to required attribute or any particular aspect of model configuration. Since this is basic page, just like Getting started page, it does mention using statements to make it easy to copy paste when understanding the concept. Based on our decision earlier we can argue that may be this code samples shouldn't have using statements either. I am fine either way though for this page, given it is at the bottom of model building blocks, it is better to have them rather than remove it.

Given the using statements on the page are not really about particular model configuration (e.g. if rather than required, page showed table configuration then using for TableAttribute would be in sample but not for Required) and there could be variety of different using statements depending on aspect of model being configured, we cannot add all of them. So we are going to keep the way it is. Unless there are multiple customers are finding issue with the particular page. As for customer issues relating to not showing usings in all code snippets, we have already decision for that.

ajcvickers:
@smitpatel After taking a second look at this, I think we should re-consider.

yecril71pl:
> Given the using statements on the page are not really about particular model configuration (e.g. if rather than required, page showed table configuration then using for TableAttribute would be in sample but not for Required) and there could be variety of different using statements depending on aspect of model being configured, we cannot add all of them. 

We should only mention namespaces for attributes mentioned further in the guide.  AFAIK, this amounts to adding 1 namespace.

ajcvickers (approved):

PR #3685, opened 2022-01-19
Warnings for deterministic ordering

### LINKED ISSUE #3242: Document behavior of AsSplitQuery() when subquery has unstable ordering
I'm sorry I cannot fully reproduce this error yet, but as I already described in a just created stackoverflow (https://stackoverflow.com/questions/67169707/azure-sql-queries-with-same-order-returning-different-resultsets) Azure SQL _somehow_ seams to have problem _keeping_ the order for some queries. This sure breaks the logic of your "AsSplitQuery()" so that I needed to remove it completely and live with the performance decrease.

Is this maybe an already known misbehavior? Is there something I can do to make it work? Anyways, I thought you should know this.

Edit: I now know that SQL Server is not guaranteed to keep the order between queries for not unique columns! The order is determined based on indexes, columns etc.. But what to do in these cases? Guess I'm not the last one to fall in this pit. I now append a `.ThenBy(e => e.UniqueIdentifier)` everytime a order is attached. But it's a pretty technical reason causing this issue and will sure confuse several persons out there!

I don't know how complex this could be under the hood, but would it be possible to always order by entities PK **last** if any / no orders are used? Guess it's getting pretty tricky right now but it is a terrible pitfall right now!

roji:
In split query, and also a stronger note in the pagination page.

Closes #3242

### DIFF
--- a/entity-framework/core/querying/pagination.md
+++ b/entity-framework/core/querying/pagination.md
@@ -10,7 +10,7 @@ uid: core/querying/pagination
 Pagination refers to retrieving results in pages, rather than all at once; this is typically done for large resultsets, where a user interface is shown that allows the user to navigate to the next or previous page of the results.
 
 > [!WARNING]
-> Regardless of the pagination method used, always make sure that your ordering is fully deterministic. For example, if results are ordered only by date, but there can be multiple results with the same date, then results could be skipped when paginating as they're ordered differently across two paginating queries. Ordering by both date and ID (or any other unique property) makes the ordering fully deterministic and avoids this problem. Note that relational databases do not apply any ordering by default, even on the primary key.
+> Regardless of the pagination method used, always make sure that your ordering is fully unique. For example, if results are ordered only by date, but there can be multiple results with the same date, then results could be skipped when paginating as they're ordered differently across two paginating queries. Ordering by both date and ID (or any other unique property or combination of properties) makes the ordering fully unique and avoids this problem. Note that relational databases do not apply any ordering by default, even on the primary key.
 
 ## Offset pagination
 
--- a/entity-framework/core/querying/pagination.md
+++ b/entity-framework/core/querying/pagination.md
@@ -10,7 +10,7 @@ uid: core/querying/pagination
 Pagination refers to retrieving results in pages, rather than all at once; this is typically done for large resultsets, where a user interface is shown that allows the user to navigate to the next or previous page of the results.
 
 > [!WARNING]
-> Regardless of the pagination method used, always make sure that your ordering is fully deterministic. For example, if results are ordered only by date, but there can be multiple results with the same date, then results could be skipped when paginating as they're ordered differently across two paginating queries. Ordering by both date and ID (or any other unique property) makes the ordering fully deterministic and avoids this problem. Note that relational databases do not apply any ordering by default, even on the primary key; queries without explicit ordering have non-deterministic resultsets.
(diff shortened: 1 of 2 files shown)


ajcvickers (approved):

smitpatel (changes requested):
There is no issue of deterministic ordering.

roji:
@smitpatel then please propose another warning text, I don't mind.

smitpatel:
We already discussed that in teams chat.

roji:
I push [this commit](https://github.com/dotnet/EntityFramework.Docs/pull/3685/commits/2f0d6f5caa4acc9eea3233a3c9af0610b297ea9f) as a result of that discussion, believing that's what you wanted. If it isn't, just propose whatever you think is right.

smitpatel:
Pagination queries without explicit ordering causes non-deterministic result set is right. (though non-pagination page should mention pagination, we discussed that pagination doesn't required to mentioned on the page which talks about pagination). But the main issue remains, there is no non-deterministic ordering, there is only non-deterministic result set.

If you want me to update the message, just assign the issue to me. I will get around to it at some point.

roji:
> But the main issue remains, there is no non-deterministic ordering, there is only non-deterministic result set.

I don't see it that way: a query whose orderings can produce different resultsets (because of lack of a unique column) can be said to have non-deterministic ordering.

But I'll close this and let you deal with this whichever way you want.

entity-framework/core/querying/pagination.md
```diff
@@ -9,6 +9,9 @@ uid: core/querying/pagination
 
 Pagination refers to retrieving results in pages, rather than all at once; this is typically done for large resultsets, where a user interface is shown that allows the user to navigate to the next or previous page of the results.
 
+> [!WARNING]
+> Regardless of the pagination method used, always make sure that your ordering is fully deterministic. For example, if results are ordered only by date, but there can be multiple results with the same date, then results could be skipped when paginating as they're ordered differently across two paginating queries. Ordering by both date and ID (or any other unique property) makes the ordering fully deterministic and avoids this problem. Note that relational databases do not apply any ordering by default, even on the primary key.
```
ajcvickers, line 13:
```suggestion
> Regardless of the pagination method used, always make sure that your ordering is fully unique. For example, if results are ordered only by date, but there can be multiple results with the same date, then results could be skipped when paginating as they're ordered differently across two paginating queries. Ordering by both date and ID (or any other unique property or combination of properties) makes the ordering fully unique and avoids this problem. Note that relational databases do not apply any ordering by default, even on the primary key.
```

entity-framework/core/querying/single-split-queries.md
```diff
@@ -42,6 +42,9 @@ INNER JOIN [Post] AS [p] ON [b].[BlogId] = [p].[BlogId]
 ORDER BY [b].[BlogId]
 ```
 
+> [!WARNING]
+> When using split queries with Skip/Take, pay special attention to making your query ordering fully deterministic; not doing so could cause incorrect data to be returned. For example, if results are ordered only by date, but there can be multiple results with the same date, then each one of the split queries could each get different results from the database. Ordering by both date and ID (or any other unique property) makes the ordering fully deterministic and avoids this problem. Note that relational databases do not apply any ordering by default, even on the primary key.
```
ajcvickers, line 46:
```suggestion
> When using split queries with Skip/Take, pay special attention to making your query ordering fully unique; not doing so could cause incorrect data to be returned. For example, if results are ordered only by date, but there can be multiple results with the same date, then each one of the split queries could each get different results from the database. Ordering by both date and ID (or any other unique property or combination of properties) makes the ordering fully unique and avoids this problem. Note that relational databases do not apply any ordering by default, even on the primary key.
```

ajcvickers (approved):

ajcvickers:
@smitpatel Can you sign off on this so I can merge?

smitpatel (approved):

PR #3708, opened 2022-01-31
Use file-scoped namespaces everywhere in samples

roji:
Done automatically with Rider (and manually removed changes to migration files).

Note that if we have any code snippets in the conceptual pages which don't reference a region, this would break them (since line numbers will now shift). However, we almost always use regions which wouldn't be affected (since line numbers are relative).


ErikEJ:
Also simple to do automatically with Visual Studio 😉

ajcvickers:
@roji Do we want to do this? Generally, we want code copied from samples to be usable by people who might not be running the latest of everything.

roji:
I'm thinking that at least EF Core 6.0 users have to use the latest SDK anyway, so they'd have to be manually restricting the language version to pre-10 for this to fail... But am OK if you prefer we didn't do this.

PR #3735, opened 2022-02-17
Fix small typo

Thomas-X:
was Microsoft.EntityFramework.Design -> should be Microsoft.EntityFrameworkCore.Design

Couldn't find it on nuget but I could find Microsoft.EntityFrameworkCore.Design which I assume was the intended package 😄


roji (approved):
Thanks!

PR #3740, opened 2022-02-20
A more formal way to eagerly load entities with entity framework.

alancuriel:
More [formal](https://www.entityframeworktutorial.net/efcore/querying-in-ef-core.aspx) way of eagerly loading entities. Previous method  caused an error since Select wasn't a property. (dotnet 6)


ajcvickers:
@alancuriel This is documentation for [EF6, not EF Core](https://docs.microsoft.com/en-us/ef/efcore-and-ef6/). The syntax is correct for EF6.

Issue #3385, opened 2021-08-15
Backing Fields|Basic configuration – bad link Including & Excluding Properties


Uzivatel919:
From [Basic configuration](https://docs.microsoft.com/en-us/ef/core/modeling/backing-field?tabs=data-annotations#basic-configuration)

> For more information on which properties are included in the model, see [Including & Excluding Properties](https://docs.microsoft.com/en-us/ef/core/modeling/entity-properties).

_Including & Excluding Properties_ link apparently links directly to article _Entity Properties_ that has chapter _Included and excluded properties_. 

I guess on some microinconsistence.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2e162696-b19f-d73f-8b1f-d1b4c3048b62
* Version Independent ID: a83e4fb5-e4ae-a46f-7248-487ea2679ac5
* Content: [Backing Fields - EF Core](https://docs.microsoft.com/en-us/ef/core/modeling/backing-field?tabs=data-annotations)
* Content Source: [entity-framework/core/modeling/backing-field.md](https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/modeling/backing-field.md)
* Product: **entity-framework**
* Technology: **entity-framework-core**
* GitHub Login: @ajcvickers
* Microsoft Alias: **avickers**

Vadimyan:
Since `Including & Excluding Properties` is a first section in `Entity Properties` page, so it's not big problem. But I think it'll be better to use concrete link to the section. I'm going to prepare a PR for this issue.

Vadimyan:
Hi! I've prepared a small PR that changes the link from [Entity Properties](https://docs.microsoft.com/en-us/ef/core/modeling/entity-properties) page to an exact section [Including & Excluding Properties](https://docs.microsoft.com/en-us/ef/core/modeling/entity-properties).

PR #3767, opened 2022-03-15
Fix a link to Including & Excluding Properties in Backing Fields|Basic configuration section

Vadimyan:
Since Including & Excluding Properties is a first section in Entity Properties page, so it's not big problem. But I think it'll be better to use concrete link to the section.

Fixes https://github.com/dotnet/EntityFramework.Docs/issues/3385


roji (approved):
Thanks!

PR #3770, opened 2022-03-17
Update keyless-entity-types.md

onlytropics:
Added tip on how to have the code be automatically generated from a database table


roji:
@onlytropics why do you think a note pointing to scaffolding an existing database is needed in this particular page (keyless entities)? We can't add links to scaffolding from all the pages (e.g. on tables, sequences...), is there any particular reason we should do it for this one?

onlytropics:
Maybe you are right, I'm not sure. However, I was learning about all this last weekend, not having implemented a controller or something like that before, and had one test case I wanted to implement which I already had in another language, so I could easily compare results and performance. It used a keyless table. I stumbled across this page, but was not able to recreate the steps in my own program. It was not clear to me where the example code has to go.
The next day, I read an article about how the code can be generated automatically, followed the steps, and ended up with sample code which showed how things work. It was then easy to reproduce. Having this info on this page would have saved me hours of searching around the internet, that's why I thought this would be a good place to put it.

ajcvickers:
Note from triage: discussed this with the team, and we agree with @roji.

@onlytropics Thanks for your interest, but we're not going to merge this.

PR #3780, opened 2022-03-24
Fix Grammar

erhanalankus:
Although stored computed columns seem like a good solution => Although using stored computed columns seems like a good solution

### DIFF
--- a/entity-framework/core/modeling/generated-properties.md
+++ b/entity-framework/core/modeling/generated-properties.md
@@ -91,7 +91,7 @@ Be sure to select the appropriate function, as several may exist (e.g. `GETDATE(
 
 ### Update timestamp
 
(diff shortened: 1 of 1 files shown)


roji (commented):
How is the proposed text more grammatical than the existing one?

erhanalankus:
The existing one seems incorrect to me.
If it's talking about plural things, the sentence should go like "seem like good solutions".
It's talking about using stored columns. So it should go like "seems like a good solution.

roji:
I don't think it's incorrect to refer to a plural as a singular, e.g. "apples are a nice fruit", "these apartments seem like a good investment".

ajcvickers:
Note from triage: we decided not to take this. Thanks anyway!

PR #3781, opened 2022-03-24
Fix Grammar

erhanalankus:
The value and state of these properties is maintained => The value and state of these properties are maintained


roji (approved):
Thanks!

PR #3803, opened 2022-04-06
Add What MARS abbreviation stands for

erhanalankus:
This is the one and only occurrence of the MARS abbreviation, so it should include the full form.

### DIFF
--- a/entity-framework/core/performance/efficient-querying.md
+++ b/entity-framework/core/performance/efficient-querying.md
@@ -163,7 +163,7 @@ If your queries return just a few results, then you probably don't have to worry
 In certain situations, EF will itself buffer the resultset internally, regardless of how you evaluate your query. The two cases where this happens are:
 
 * When a retrying execution strategy is in place. This is done to make sure the same results are returned if the query is retried later.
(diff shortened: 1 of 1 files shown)


entity-framework/core/performance/efficient-querying.md
```diff
@@ -163,7 +163,7 @@ If your queries return just a few results, then you probably don't have to worry
 In certain situations, EF will itself buffer the resultset internally, regardless of how you evaluate your query. The two cases where this happens are:
 
 * When a retrying execution strategy is in place. This is done to make sure the same results are returned if the query is retried later.
-* When [split query](xref:core/querying/single-split-queries) is used, the resultsets of all but the last query are buffered - unless MARS is enabled on SQL Server. This is because it is usually impossible to have multiple query resultsets active at the same time.
+* When [split query](xref:core/querying/single-split-queries) is used, the resultsets of all but the last query are buffered - unless Multiple Active Result Sets(MARS) is enabled on SQL Server. This is because it is usually impossible to have multiple query resultsets active at the same time.
```
roji, line 166:
```suggestion
* When [split query](xref:core/querying/single-split-queries) is used, the resultsets of all but the last query are buffered - unless MARS (Multiple Active Result Sets) is enabled on SQL Server. This is because it is usually impossible to have multiple query resultsets active at the same time.
```

roji (approved):
Thanks.

PR #3809, opened 2022-04-07
Update index.md

Folivi95:
ef core now requires the user to specify the project name/path hence the reason for adding the "--project" tag to the docs. Please let me know if this is okay or basically just what you think.


smitpatel:
`--project` is not required when running the command from the directory which contains the csproj file for the project and there are not other csproj file in the same directory. This change is not needed.

PR #3822, opened 2022-04-17
Update docs for DbContext pooling with state

### LINKED ISSUE #3787: Update context pooling multitenancy sample to more closely approximate a real application
Hi,
As you know the DbContext should be injected as Scoped mode. But I guess the WeatherForecastScopedFactory can be injected as singleton, because there is no need to have a scoped factory object per request (per scope). The singleton WeatherForecastScopedFactory can create a scoped DbContext per scope and the corresponding DbContext will be pooled.
Could you please describe that why the "WeatherForecastScopedFactory" service is injected as scoped? 

There is the line which scoped injection is occured: 

https://github.com/dotnet/EntityFramework.Docs/blob/a23030677b160bebaa82bac686bc1dd8de0882eb/samples/core/Performance/AspNetContextPoolingWithState/Program.cs#L21

roji:
To have a proper, scoped ITenant service.

Closes #3787

### DIFF
--- a/entity-framework/core/performance/advanced-performance-topics.md
+++ b/entity-framework/core/performance/advanced-performance-topics.md
@@ -51,15 +51,19 @@ Context pooling works by reusing the same context instance across requests; this
 
 A typical scenario involving context state would be a multi-tenant ASP.NET Core application, where the context instance has a *tenant ID* which is taken into account by queries (see [Global Query Filters](xref:core/querying/filters) for more details). Since the tenant ID needs to change with each web request, we need to have go through some extra steps to make it all work with context pooling.
 
-First, register a pooling context factory as a Singleton service, as usual:
+Let's assume that your application registers a scoped `ITenant` service, which wraps the tenant ID and any other tenant-related information:
+
+[!code-csharp[Main](../../../samples/core/Performance/AspNetContextPoolingWithState/Program.cs#TenantResolution)]
+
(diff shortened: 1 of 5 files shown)


entity-framework/core/performance/advanced-performance-topics.md
```diff
@@ -51,16 +51,20 @@ Context pooling works by reusing the same context instance across requests; this
 
 A typical scenario involving context state would be a multi-tenant ASP.NET Core application, where the context instance has a *tenant ID* which is taken into account by queries (see [Global Query Filters](xref:core/querying/filters) for more details). Since the tenant ID needs to change with each web request, we need to have go through some extra steps to make it all work with context pooling.
 
-First, register a pooling context factory as a Singleton service, as usual:
+Let's assume that your application registers a scoped `ITenant` service, which wraps the tenant ID and any other tenant-related information:
+
+[!code-csharp[Main](../../../samples/core/Performance/AspNetContextPoolingWithState/Program.cs#TenantResolution)]
+
+As written above, pay special attention to where you get the tenant ID from: this is an important aspect of your application's security.
```
AndriySvyryd, line 58:
```suggestion
As written above, pay special attention to where you get the tenant ID from - this is an important aspect of your application's security.
```

AndriySvyryd:
I wonder whether `Tenant` would be reset if it implements `IResettableService`. We should mention it one way or the other.

AndriySvyryd (approved):

PR #3823, opened 2022-04-17
Update transactions.md

eersonmez:
Under the 'Share Connection and transaction' section, the 3rd highlight is on the wrong line. Instead of highlighting the `UseTransaction`, it highlights the opening curly bracket.


roji:
Thanks - but all of the lines referenced here seem to be off-by-one, not just the one you fixed. Do you mind taking care of the others as well?

eersonmez:
@roji Sure thing. Fixed those other lines as well. The first line (`new SqlConnection`) also seemed relevant to be highlighted so kept that highlighted.

roji (approved):
Thanks!

PR #3828, opened 2022-04-20
Refactor code to minimal style

vanillajonathan:


roji:
Sorry it took me a long while to review this. I ended up merging #3865 which does the same, but also omits the unneeded namespace (the less the better in this kind of sample).

Thanks for your contribution!

PR #3841, opened 2022-04-29
Change "perf" to "performance"

w-syss:
The choice of using "perf" instead of "performance" seems a bit out of place, when "perf" is used nowhere else as an abbreviation for "performance".


roji (approved):
Thanks.

smitpatel:
We shouldn't use abbreviations unless well defined in language else it will break localization.

Issue #3729, opened 2022-02-14
Remove warning about #23408


ajcvickers:
See https://github.com/dotnet/efcore/issues/23408#issuecomment-1031877250

PR #3844, opened 2022-05-03
Update user-defined-function-mapping.md

MACEL94:
solves #3729


MACEL94:
relative comment: https://github.com/dotnet/efcore/issues/23408#issuecomment-1031877250
#23408

PR #3918, opened 2022-07-04
Fix InvalidOperationException

Cogax:
There was a small mismatch between the thrown exception message and the mentioned exception message. This leads to no results on searches by exact phrasing.


roji (approved):
Thanks

PR #3945, opened 2022-07-22
Removing obsolete extensions

ErikEJ:


ajcvickers (approved):
Looks good to me, but I'll run it by the team before merging.

ajcvickers:
Thanks, @ErikEJ.

Note: if an extension was removed here that works with newer versions of EF Core, then anyone should feel free to send a PR to bring it back with updated information.

PR #3965, opened 2022-08-03
Add note that pre-convention configuration takes precedence over data annotation

Peter-B-:
According to [28541](https://github.com/dotnet/efcore/issues/28541), pre-convention configuration takes precedence over data annotation. I think this should be stated in the documentation.


ajcvickers:
@Peter-B- We discussed this in triage. The note is good, but please leave it as a single line and remove the link and other text.

Peter-B-:
Hi @ajcvickers, thanks for your feedback. I removed the second line.

ajcvickers (approved):

ajcvickers:
Thanks!

PR #3982, opened 2022-08-16
What's new for ExecuteUpdate and ExecuteDelete

ajcvickers:


entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -19,6 +19,316 @@ EF7 is also available as [daily builds](https://github.com/dotnet/efcore/blob/ma
...
+
+However, it is sometimes useful to execute update or delete commands on the database without involving the change tracker. EF7 enables this with the new `ExecuteUpdate` and `ExecuteDelete` methods. These methods are applied to a LINQ query and will update or delete entities in the database based on the results of that query. Many entities can be updated with a single command and the entities are not loaded into memory, which means this can result in more efficient updates and deletes.
+
+However, keep in mind that:
+
+- The specific changes to make must be specified explicitly; they are not automatically detected by EF Core.
+- Any tracked entities will not be kept in sync.
+- Additional commands may need to be sent in the correct order so as not to violate database constraints. For example deleting dependents before a principal can be deleted.
+
+All of this means that the `ExecuteUpdate` and `ExecuteDelete` methods compliment, rather than replace, the existing `SaveChanges` mechanism.
```
roji, line 34:
```suggestion
All of this means that the `ExecuteUpdate` and `ExecuteDelete` methods complement, rather than replace, the existing `SaveChanges` mechanism.
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -19,6 +19,316 @@ EF7 is also available as [daily builds](https://github.com/dotnet/efcore/blob/ma
...
+
+> [!TIP]
+> The sample model can be found in [BlogsContext.cs](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Miscellaneous/NewInEFCore7/BlogsContext.cs).
+
+### Basic `ExecuteDelete` examples
+
+> [!TIP]
+> The code shown here comes from [ExecuteDeleteSample.cs](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Miscellaneous/NewInEFCore7/ExecuteDeleteSample.cs).
+
+Calling `ExecuteDelete` or `ExecuteDeleteAsync` on a `DbSet` will immediately delete all entities of that `DbSet` from the database. For example, to delete all `Tag` entities:
```
roji, line 116: Generally try to avoid the future tense

```suggestion
Calling `ExecuteDelete` or `ExecuteDeleteAsync` on a `DbSet` immediately deletes all entities of that `DbSet` from the database. For example, to delete all `Tag` entities:
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -19,6 +19,316 @@ EF7 is also available as [daily builds](https://github.com/dotnet/efcore/blob/ma
...
+FROM [Blogs] AS [b]
+```
+
+As with `ExecuteDelete`, the query can be used to filter which entities are updated. In addition, multiple calls to `SetProperty` can be used to update more than one property on the target entity. For example, to update the `Title` and `Content` of all posts published before 2022:
+
+<!--
+        await context.Posts
+            .Where(p => p.PublishedOn.Year < 2022)
+            .ExecuteUpdateAsync(
+                s => s.SetProperty(b => b.Title, b => b.Title + " (" + b.PublishedOn.Year + ")")
```
roji, line 191: Maybe indent so that the SetProperty's are aligned

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -19,6 +19,316 @@ EF7 is also available as [daily builds](https://github.com/dotnet/efcore/blob/ma
...
+```csharp
+await context.Authors.ExecuteDeleteAsync();
+```
+
+Will result in the following exception on SQL Server:
+
+> Microsoft.Data.SqlClient.SqlException (0x80131904): The DELETE statement conflicted with the REFERENCE constraint "FK_Posts_Authors_AuthorId". The conflict occurred in database "TphBlogsContext", table "dbo.Posts", column 'AuthorId'.
+The statement has been terminated.
+
+To fix this, we must first either delete the posts, or sever the relationship between each post and its author by setting `AuthorId` foreign key property to null. For example, using the delete option:
```
roji, line 288: I'd put the database cascading delete solution first, since it's really the thing to do wheter supported. Honestly, that's the universal relational way to do this, and this whole complexity is only because SQL Server doesn't support cycles... Like in PG I'd always recommend cascade deletes...

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -19,6 +19,316 @@ EF7 is also available as [daily builds](https://github.com/dotnet/efcore/blob/ma
...
+```sql
+DELETE FROM [a]
+FROM [Authors] AS [a]
+```
+
+> [!IMPORTANT]
+> Multiple `ExecuteDelete` and `ExecuteUpdate` commands will not be contained in a single transaction by default. However, the [DbContext transaction APIs](xref:core/saving/transactions) can be used in the normal way to wrap these commands in a transaction.
+
+> [!TIP]
+> Sending these commands in a single round-trip depends on [Issue #10879](https://github.com/dotnet/efcore/issues/10879) Vote for this issue if it's something you would like to see implemented.
```
roji, line 314:
```suggestion
> Sending these commands in a single round-trip depends on [Issue #10879](https://github.com/dotnet/efcore/issues/10879). Vote for this issue if it's something you would like to see implemented.
```

roji (approved):

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
ajcvickers: I'd rather leave it the way it is, because talking about the problem first shows why cascades are useful.

roji: Makes sense!

PR #3987, opened 2022-08-18
Document SQLite 3.35 requirement

bricelam:
Waiting for the outcome of https://github.com/dotnet/efcore/issues/28776 before merging, since we might decide to change this.


bricelam:
```
If you're using SQLitePCLRaw.bundle_winsqlite3, the minimum required version of Windows 10 is 1903
```

PR #3993, opened 2022-08-22
Remove warnings

smitpatel:


roji (approved):
No idea what this means but ship it

entity-framework/index.yml
```diff
@@ -41,11 +41,11 @@ highlightedContent:
     # Card
     - title: "API reference browser"
       itemType: reference
-      url: /dotnet/api/?view=efcore-6.0
+      url: /dotnet/api/?view=efcore-6.0&preserve-view=true
     # Card
     - title: "Explore other options for .NET Data"
       itemType: reference
-      url: dotnet-data
+      url: dotnet-data/index.yml
```
smitpatel, line 48: This file needs exact file URL. Earlier it was the folder. Which in web world, tries to go to index.* and it worked by accident.

entity-framework/toc.yml
```diff
@@ -411,7 +411,7 @@
         href: core/cli/services.md
 
     - name: EF Core API reference >>
-      href: /dotnet/api/?view=efcore-6.0
+      href: /dotnet/api/?view=efcore-6.0&preserve-view=true
```
smitpatel, line 414: They recommend not using the version in URL string so it can go to latest by default but since it is a shared docset there is no way for us to specify efcore without using the version (monikers are tied to version). So the later arg explicitly state that we want 6.0 version and don't generate warning.

PR #4001, opened 2022-08-27
Add EF7 function translations

### LINKED ISSUE #3868: Document new function translations in 7.0
* [x] https://github.com/dotnet/efcore/issues/25396
* [x] https://github.com/dotnet/efcore/issues/28078
* [x] https://github.com/dotnet/efcore/issues/28110
* [x] https://github.com/dotnet/efcore/issues/28115

bricelam:
Resolves #3868

### DIFF
--- a/entity-framework/core/providers/sql-server/functions.md
+++ b/entity-framework/core/providers/sql-server/functions.md
@@ -9,6 +9,23 @@ uid: core/providers/sql-server/functions
 
 This page shows which .NET members are translated into which SQL functions when using the SQL Server provider.
 
+## Aggregate functions
+
+.NET                                                                    | SQL                              | Added in
+----------------------------------------------------------------------- | -------------------------------- | --------
+EF.Functions.StandardDeviationSample(group.Select(x => x.Property))     | STDEV(Property)                  | EF Core 7.0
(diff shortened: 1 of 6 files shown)


entity-framework/core/providers/sql-server/functions.md
```diff
@@ -9,6 +9,23 @@ uid: core/providers/sql-server/functions
 
 This page shows which .NET members are translated into which SQL functions when using the SQL Server provider.
 
+## Aggregate functions
+
+.NET                                                                    | SQL                              | Added in
+----------------------------------------------------------------------- | -------------------------------- | --------
+EF.Functions.StandardDeviationSample(group.Select(x => x.Property))     | STDEV(Property)                  | EF Core 7.0
```
roji, line 16: We should probably have a general conceptual doc section showing how to use aggregate functions in various scenarios, opened #4002.

BTW technically the thing that GroupBy returns is called a "grouping" rather than a group, we may want to align (though I think group is OK too).

For reference, here are the [PG translations](https://www.npgsql.org/efcore/mapping/translations.html#aggregate-functions).

roji (approved):

entity-framework/core/providers/sql-server/functions.md
bricelam: > technically the thing that GroupBy returns is called a "grouping" rather than a group

lol, I had the same internal debate. In the end, I went for the shorter one.

PR #4007, opened 2022-08-31
Sample for stored procedure mapping

ajcvickers:
No docs yet.

### DIFF
--- a/samples/core/Miscellaneous/NewInEFCore7/DocumentsContext.cs
+++ b/samples/core/Miscellaneous/NewInEFCore7/DocumentsContext.cs
@@ -115,7 +115,7 @@ public abstract class DocumentsContext : DbContext
                         "Person_Update",
                         storedProcedureBuilder =>
                         {
-                            storedProcedureBuilder.HasParameter(person => person.Id);
+                            storedProcedureBuilder.HasOriginalValueParameter(person => person.Id);
                             storedProcedureBuilder.HasOriginalValueParameter(
                                 person => person.Name, parameterBuilder => parameterBuilder.HasName("Name_Original"));
                             storedProcedureBuilder.HasParameter(person => person.Name);
@@ -126,7 +126,7 @@ public abstract class DocumentsContext : DbContext
                         "Person_Delete",
                         storedProcedureBuilder =>
                         {
-                            storedProcedureBuilder.HasParameter(person => person.Id);
+                            storedProcedureBuilder.HasOriginalValueParameter(person => person.Id);
                             storedProcedureBuilder.HasOriginalValueParameter(person => person.Name);
                             storedProcedureBuilder.HasRowsAffectedResultColumn();
                         });
@@ -151,7 +151,7 @@ public abstract class DocumentsContext : DbContext
                                 "Contacts_Update",
                                 storedProcedureBuilder =>
                                 {
-                                    storedProcedureBuilder.HasParameter("PersonId");
+                                    storedProcedureBuilder.HasOriginalValueParameter("PersonId");
                                     storedProcedureBuilder.HasParameter(contactDetails => contactDetails.Phone);
                                     storedProcedureBuilder.HasRowsAffectedResultColumn();
                                 });
@@ -159,7 +159,7 @@ public abstract class DocumentsContext : DbContext
                             ownedNavigationBuilder.DeleteUsingStoredProcedure(
                                 "Contacts_Delete", storedProcedureBuilder =>
                                 {
-                                    storedProcedureBuilder.HasParameter("PersonId");
+                                    storedProcedureBuilder.HasOriginalValueParameter("PersonId");
                                     storedProcedureBuilder.HasRowsAffectedResultColumn();
                                 });
                         }
@@ -186,7 +186,7 @@ public abstract class DocumentsContext : DbContext
                                         "Addresses_Update",
                                         storedProcedureBuilder =>
                                         {
-                                            storedProcedureBuilder.HasParameter("ContactDetailsPersonId");
+                                            storedProcedureBuilder.HasOriginalValueParameter("ContactDetailsPersonId");
                                             storedProcedureBuilder.HasParameter(address => address.Street);
                                             storedProcedureBuilder.HasParameter(address => address.City);
                                             storedProcedureBuilder.HasParameter(address => address.Postcode);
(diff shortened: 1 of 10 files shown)


samples/core/Miscellaneous/NewInEFCore7/DocumentsContext.cs
```diff
@@ -0,0 +1,656 @@
...
+                            storedProcedureBuilder.HasParameter(document => document.NumberOfPages);
+                            storedProcedureBuilder.HasParameter(document => document.PublicationDate);
+                            storedProcedureBuilder.HasParameter(document => document.CoverArt);
+                            storedProcedureBuilder.HasResultColumn(document => document.Id);
+                            storedProcedureBuilder.HasResultColumn(document => document.FirstRecordedOn);
+                            storedProcedureBuilder.HasResultColumn(document => document.RetrievedOn);
+                            storedProcedureBuilder.HasResultColumn(document => document.RowVersion);
+                        });
+
+                    entityTypeBuilder.UpdateUsingStoredProcedure(
```
roji, line 387: This mapping is off: Document_Update accepts all columns for the entire hierarchy (so suitable for TPH), but is here being used for TPT... So the parameter list below doesn't correspond to the actual parameters accepted by the sproc.

samples/core/Miscellaneous/NewInEFCore7/DocumentsContext.cs
```diff
@@ -0,0 +1,656 @@
...
+                        {
+                            storedProcedureBuilder.HasParameter(person => person.Id);
+                            storedProcedureBuilder.HasOriginalValueParameter(
+                                person => person.Name, parameterBuilder => parameterBuilder.HasName("Name_Original"));
+                            storedProcedureBuilder.HasParameter(person => person.Name);
+                            storedProcedureBuilder.HasRowsAffectedResultColumn();
+                        });
+
+                    entityTypeBuilder.DeleteUsingStoredProcedure(
+                        "Person_Delete",
```
roji, line 126: This mapping - and many others - is missing a rows affected result column definition; the sproc definition does return a result set with it (`OUTPUT 1`). This means that result sets in the batch can get misaligned as EF thinks this sproc doesn't return anything but it does.

roji (changes requested):
@ajcvickers I found some mapping issues here which explain most of the errors in #2487 - take a look below (but this also uncovered issues in my code!). We can iterate on this together tomorrow  if you want.

samples/core/Miscellaneous/NewInEFCore7/DocumentsContext.cs
```diff
@@ -0,0 +1,656 @@
...
+                    entityTypeBuilder.UpdateUsingStoredProcedure(
+                        "Person_Update",
+                        storedProcedureBuilder =>
+                        {
+                            storedProcedureBuilder.HasParameter(person => person.Id);
+                            storedProcedureBuilder.HasOriginalValueParameter(
+                                person => person.Name, parameterBuilder => parameterBuilder.HasName("Name_Original"));
+                            storedProcedureBuilder.HasParameter(person => person.Name);
+                            storedProcedureBuilder.HasRowsAffectedResultColumn();
+                        });
```
roji, line 123: BTW any reason not to just chain calls in a single expression here and everywhere else?

```suggestion
                    entityTypeBuilder.UpdateUsingStoredProcedure(
                        "Person_Update",
                        storedProcedureBuilder => storedProcedureBuilder
                            .HasParameter(person => person.Id)
                            .HasOriginalValueParameter(person => person.Name, parameterBuilder => parameterBuilder.HasName("Name_Original"))
                            .HasParameter(person => person.Name)
                            .HasRowsAffectedResultColumn());
```

ajcvickers: No reason, other than I added them one at a time and didn't think about it. Will do.

samples/core/Miscellaneous/NewInEFCore7/DocumentsContext.cs
ajcvickers: The TPT table is this:

```sql
      CREATE TABLE [Documents] (
          [Id] int NOT NULL IDENTITY,
          [RowVersion] rowversion NULL,
          [Title] nvarchar(max) NOT NULL,
          [NumberOfPages] int NOT NULL,
          [PublicationDate] datetime2 NOT NULL,
          [CoverArt] varbinary(max) NULL,
          [FirstRecordedOn] datetime2 NOT NULL DEFAULT (getutcdate()),
          [RetrievedOn] AS getutcdate(),
          CONSTRAINT [PK_Documents] PRIMARY KEY ([Id])
```

Which parameters are now correct?

ajcvickers:
@roji Pushed some updates. Updating data is still failing for TPH. Optimistic concurrency test 1 is still failing in all cases.

samples/core/Miscellaneous/NewInEFCore7/DocumentsContext.cs
roji: Sorry, I think I missed that you have different versions of the same sproc (Document_Update) based on which inheritance strategy is being used (maybe consider calling the sproc Document_Tpt_Update). I'll look again.

roji:
Found two more bugs while investigating this:

1. When updating, if a column hasn't changed (e.g. the discriminator), we don't create a parameter for it, and everything gets mismatched. This was raised by Andriy in https://github.com/dotnet/efcore/pull/28908#discussion_r957513986, and I'm working on it there. It basically means that partial updates are currently broken.
2. We don't properly detect concurrency exceptions when the rows affected is in the result set, *and* there are other result set values to propagate. 
  a. In regular SQL, when there's nothing to propagate, we expect a single value with 1 or 0 (from OUTPUT 1 SQL).
  b. But when there's something to propagate, we interpret the existence of a row as a sign that the update succeeded.
  c. While this works for regular SQL, with sprocs we always get back a row. So we need to add logic to check for 1 or 0 even when there's propagation.

I'll be back to full work on Tuesday (will be partially doing stuff on Monday too) and will concentrate on fixing all this (not feeling great about the work I've done here...)

roji:
@ajcvickers just making sure that everything works for you and you're not blocked on me by anything.

roji (approved):

PR #4042, opened 2022-09-19
Pack of doc changes for SQL queries

### LINKED ISSUE #1787: Document ExecuteSql{Raw,Interpolated}
We have a page on [Raw SQL Queries](https://docs.microsoft.com/en-us/ef/core/querying/raw-sql), but we don't document the non-query raw SQL methods.

### LINKED ISSUE #3971: Guide users to doing raw SQL via FormattableString by default
Via the new suffix-less FromSql/ExecuteSql methods being introduced in https://github.com/dotnet/efcore/issues/28609, since they're safe from SQL injection.

### LINKED ISSUE #4041: Document SQL queries for basic, non-entity types
See https://github.com/dotnet/efcore/issues/11624

roji:
* Switch to showing the new `FromSql` by default.
* Clearly document FromSqlRaw as appropriate only for dynamic SQL scenarios.
* Possibly controversial: this renames the raw SQL "topic" to "SQL queries". The problem with "raw" is that its used in one of the APIs (FromSqlRaw) but not on the others (is FromSqlInterpolated a raw SQL query? :trollface:)
* Documented the new SqlQuery API for primitive types
* Added a "Non-querying SQL" section under the "SQL queries" page, for ExecuteSql. Yeah, it's weird - but doing a whole page just for that seemed really excessive.

Closes #3971
Closes #4041
Closes #1787


entity-framework/core/querying/sql-queries.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Change Tracking
+
+Queries that use <xref:Microsoft.EntityFrameworkCore.RelationalQueryableExtensions.FromSql%2A> or <xref:Microsoft.EntityFrameworkCore.RelationalQueryableExtensions.FromSqlRaw%2A> follow the exact same change tracking rules as any other LINQ query in EF Core. For example, if the query projects entity types, the results are tracked by default.
+
+The following example uses a SQL query that selects from a Table-Valued Function (TVF), then disables change tracking with the call to [`AsNoTracking`](xref:core/querying/tracking#no-tracking-queries):
+
+[!code-csharp[Main](../../../samples/core/Querying/SqlQueries/Program.cs#FromSqlAsNoTracking)]
+
+## Querying primitive (non-entity) types
```
roji, line 128: /cc @smitpatel

entity-framework/core/querying/raw-sql.md
```diff
@@ -1,99 +0,0 @@
----
-title: Raw SQL Queries - EF Core
-description: Using raw SQL for queries in Entity Framework Core
-author: smitpatel
-ms.date: 11/15/2021
-uid: core/querying/raw-sql
```
ajcvickers, line 6: We need to make sure links to this are not broken.

entity-framework/core/querying/sql-queries.md
```diff
@@ -0,0 +1,163 @@
...
+
+[!code-csharp[Main](../../../samples/core/Querying/SqlQueries/Program.cs#FromSqlStoredProcedureNamedSqlParameter)]
+
+If you need more control over the database parameter being sent, you can also construct a `DbParameter` and supply it as a parameter value. This allows you to set the precise database type of the parameter, or facets such as its size, precision or length:
+
+[!code-csharp[Main](../../../samples/core/Querying/SqlQueries/Program.cs#FromSqlStoredProcedureSqlParameter)]
+
+> [!NOTE]
+>
+> Pay attention to the ordering of your parameters when executing a stored procedure. Unless using named parameters, the parameter in the SQL must exactly match the stored procedure definition, otherwise you may encounter type conversion exceptions and/or unexpected behavior when the procedure is executed.
```
ajcvickers, line 58: This feels like two notes; one for ordering, one for making sure the parameters match.

entity-framework/core/querying/sql-queries.md
```diff
@@ -0,0 +1,163 @@
...
+```c#
+var propertyName = "User";
+var propertyValue = "johndoe";
+
+var blogs = context.Blogs
+    .FromSql($"SELECT * FROM [Blogs] WHERE {propertyName} = {propertyValue}")
+    .ToList();
+```
+
+This code doesn't work, since database do not allow parameterizing column names (or any other part of the schema).
```
ajcvickers, line 73:
```suggestion
This code doesn't work, since databases do not allow parameterizing column names (or any other part of the schema).
```

entity-framework/core/querying/sql-queries.md
```diff
@@ -0,0 +1,163 @@
...
+
+First, carefully consider whether you really want to fully parameterize the column name. For example, users may choose a column that isn't indexed, making the query run extremely slowly and overload your database. Except for truly dynamic scenarios, it's usually better to have two SQL queries for two column names, rather than using parameterization to collapse them into a single SQL query.
```
ajcvickers, line 75: Not sure this really belongs here; the same is true when using the more common LINQ method builder to choose a property dynamically. I.e. it's not a consequence of using raw SQL. But I don't feel strongly.

entity-framework/core/querying/sql-queries.md
```diff
@@ -0,0 +1,163 @@
...
+
+If you've decided you do want to use parameterization, you'll have to use <xref:Microsoft.EntityFrameworkCore.RelationalQueryableExtensions.FromSqlRaw%2A>, which allows interpolating variable data directly into the SQL string, instead of using a database parameter:
```
ajcvickers, line 77: I don't think this is true. Using `Where(e => EF.Property<object>(thePropertyName) == x)` is a better way of doing this.

ajcvickers (approved):
Good stuff. Just make sure not to break links to existing docs.

.openpublishing.redirection.json
```diff
@@ -464,6 +464,11 @@
             "source_path": "entity-framework/core/managing-schemas/scaffolding.md",
             "redirect_url": "/ef/core/managing-schemas/scaffolding/index",
             "redirect_document_id": false
+        },
+        {
+            "source_path": "entity-framework/core/querying/raw-sql.md",
```
smitpatel, line 469: Also verify if uid redirects too.

smitpatel: It works. links should be good.

entity-framework/core/querying/sql-queries.md
roji: Oh sure, the idea was "if you've decided to use parameterization and have to do SQL as opposed to LINQ" (somewhat implied in the page context). I'll amend to make sure this is clearer.

entity-framework/core/querying/sql-queries.md
roji: You mean one note that ordering must match, and another that e.g. facets must match? If so, I'll make that clearer.

entity-framework/core/querying/sql-queries.md
```diff
@@ -0,0 +1,163 @@
...
+> [!NOTE]
+>
+> Prior to EF Core 7.0, it was sometimes necessary to use the `ExecuteSql` APIs to perform a "bulk update" on the database, as above; this is considerably more efficient than querying for all matching rows and then using `SaveChanges` to modify them. EF Core 7.0 introduced [ExecuteUpdate and ExecuteDelete](xref:core/what-is-new/ef-core-7#executeupdate-and-executedelete-bulk-updates), which made it possible to express efficient bulk update operations via LINQ. It's recommended to use those APIs whenever possible, instead of `ExecuteSql`.
+
+## Limitations
+
+There are a few limitations to be aware of when returning entity types from SQL queries:
+
+- The SQL query must return data for all properties of the entity type.
+- The column names in the result set must match the column names that properties are mapped to. Note this behavior is different from EF6. EF6 ignored property to column mapping for SQL queries and result set column names had to match the property names.
```
smitpatel, line 162: those properties?

entity-framework/core/querying/sql-queries.md
roji: It's true that you have the same consideration with EF.Property - assuming you pass it arbitrary input, and not a fixed string (e.g. because of a shadow property - which I think is what it's usually used for).

The idea here was to make the user fully understand what it is they're doing when they interpolate a fully dynamic column name into their SQL (and possibly dissuade them from it, unless it's really required)... I do agree that a similar note could be a good idea for EF.Property as well.

I still think it's not a bad fit here to help "overall understanding/context", but can remove if people are against it.

entity-framework/core/querying/sql-queries.md
ajcvickers: But people might think that the _only way to do this_ is with raw SQL.

entity-framework/core/querying/sql-queries.md
roji: Revised this section to make it clear that it's a general note on dynamic queries rather than specific to SQL queries.

entity-framework/core/querying/sql-queries.md
roji: Tweaked the wording to make it clear that we're discussing dynamic construction only in the SQL context.

entity-framework/core/querying/sql-queries.md
roji: Sure

PR #4052, opened 2022-09-23
What's New: Interception

ajcvickers:


entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2064,3 +2064,558 @@ Configuration of stored procedures many-to-many join entities can be performed a
...
+                hasLogger.Logger = _logger;
+            }
+
+            return instance;
+        }
+    }
+-->
+[!code-csharp[LoggerInjectionInterceptor](../../../../samples/core/Miscellaneous/NewInEFCore7/InjectLoggerSample.cs?name=LoggerInjectionInterceptor)]
+
+This time a new instance of the interceptor is used for each `DbContext` instance, since the `ILogger` obtained can change per `DbContext` instance, and the `Ilogger` is cached on the interceptor:
```
roji, line 2253:
```suggestion
This time a new instance of the interceptor is used for each `DbContext` instance, since the `ILogger` obtained can change per `DbContext` instance, and the `ILogger` is cached on the interceptor:
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2064,3 +2064,558 @@ Configuration of stored procedures many-to-many join entities can be performed a
...
+
+<!--
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+            => optionsBuilder.AddInterceptors(new LoggerInjectionInterceptor());
+-->
+[!code-csharp[OnConfiguring](../../../../samples/core/Miscellaneous/NewInEFCore7/InjectLoggerSample.cs?name=OnConfiguring)]
+
+Now, whenever the `Customer.PhoneNumber` is changed, this changed will be logged to the application's log. For example:
```
roji, line 2261:
```suggestion
Now, whenever the `Customer.PhoneNumber` is changed, this change will be logged to the application's log. For example:
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2064,3 +2064,558 @@ Configuration of stored procedures many-to-many join entities can be performed a
...
+
+> [!TIP]
+> The code shown here comes from [OptimisticConcurrencyInterceptionSample.cs](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Miscellaneous/NewInEFCore7/OptimisticConcurrencyInterceptionSample.cs).
+
+EF Core supports the [optimistic concurrency pattern](xref:core/saving/concurrency) by checking that the number of rows actually affected by an update or delete is the same as the number of rows expected to be affected. This is often coupled with a concurrency token; that is, a column value that will only match its expected value if the row has not been updated since the expected value was read.
+
+EF signals a violation of optimistic concurrency by throwing a <xref:Microsoft.EntityFrameworkCore.DbUpdateConcurrencyException>. In EF7 [`ISaveChangesInterceptor`](https://github.com/dotnet/efcore/blob/main/src/EFCore/Diagnostics/ISaveChangesInterceptor.cs) has new methods `ThrowingConcurrencyException` and `ThrowingConcurrencyExceptionAsync` that are called before the `DbUpdateConcurrencyException` is thrown. These interception points allow the exception to be suppressed, possibly coupled with async database changes to resolve the violation.
```
roji, line 2421:
```suggestion
EF signals a violation of optimistic concurrency by throwing a <xref:Microsoft.EntityFrameworkCore.DbUpdateConcurrencyException>. In EF7, [`ISaveChangesInterceptor`](https://github.com/dotnet/efcore/blob/main/src/EFCore/Diagnostics/ISaveChangesInterceptor.cs) has new methods `ThrowingConcurrencyException` and `ThrowingConcurrencyExceptionAsync` that are called before the `DbUpdateConcurrencyException` is thrown. These interception points allow the exception to be suppressed, possibly coupled with async database changes to resolve the violation.
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2064,3 +2064,558 @@ Configuration of stored procedures many-to-many join entities can be performed a
...
+            InterceptionResult result,
+            CancellationToken cancellationToken = default)
+            => new(ThrowingConcurrencyException(eventData, result));
+    }
+-->
+[!code-csharp[SuppressDeleteConcurrencyInterceptor](../../../../samples/core/Miscellaneous/NewInEFCore7/OptimisticConcurrencyInterceptionSample.cs?name=SuppressDeleteConcurrencyInterceptor)]
+
+There are several things worth noting about this interceptor:
+
+- Both the synchronous and asynchronous interception methods are implemented. This is important if the application may call either `SaveChanges` or `SaveChangesAsync`. However, if all application code is async, then only `ThrowingConcurrencyExceptionAsync` needs to be implemented. Likewise, if the application never uses synchronous database methods, then only `ThrowingConcurrencyException` needs to be implemented. This is generally true for all interceptors with sync and async methods. (It might be worthwhile implementing the method your application does not use to throw, just in case some sync/async code creeps in.)
```
roji, line 2454:
```suggestion
- Both the synchronous and asynchronous interception methods are implemented. This is important if the application may call either `SaveChanges` or `SaveChangesAsync`. However, if all application code is async, then only `ThrowingConcurrencyExceptionAsync` needs to be implemented. Likewise, if the application never uses asynchronous database methods, then only `ThrowingConcurrencyException` needs to be implemented. This is generally true for all interceptors with sync and async methods. (It might be worthwhile implementing the method your application does not use to throw, just in case some sync/async code creeps in.)
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2064,3 +2064,558 @@ Configuration of stored procedures many-to-many join entities can be performed a
...
+- The interceptor has access to <xref:Microsoft.EntityFrameworkCore.ChangeTracking.EntityEntry> objects for the entities being saved. In this case, this is used to check whether or not the concurrency violation is happening for a delete operation.
+- If the application is using a relational database provider, then the [`ConcurrencyExceptionEventData`](https://github.com/dotnet/efcore/blob/main/src/EFCore/Diagnostics/ConcurrencyExceptionEventData.cs) object can be cast to a [`RelationalConcurrencyExceptionEventData`](https://github.com/dotnet/efcore/blob/main/src/EFCore.Relational/Diagnostics/RelationalConcurrencyExceptionEventData.cs) object. This provides additional, relational-specific information about the database operation being performed. In this case, the relational command text is printed to the console.
+- Returning `InterceptionResult.Suppress()` tells EF Core to suppress the action it was about to take--in this case, throwing the `DbUpdateConcurrencyException`. This ability to _change the behavior if EF Core_, rather than just observing what EF Core is doing, is one of the most powerful features of interceptors.
```
roji, line 2457:
```suggestion
- Returning `InterceptionResult.Suppress()` tells EF Core to suppress the action it was about to take--in this case, throwing the `DbUpdateConcurrencyException`. This ability to _change the behavior of EF Core_, rather than just observing what EF Core is doing, is one of the most powerful features of interceptors.
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2064,3 +2064,558 @@ Configuration of stored procedures many-to-many join entities can be performed a
...
+Connection strings are often static assets read from a configuration file. These can easily be passed to `UseSqlServer` or similar when configuring a `DbContext`. However, sometimes the connection string can change for each context instance. For example, each tenant in a multi-tenant system may have different connection string.
+
+EF7 makes it easier to handle dynamic connections and connection strings through improvements to the <xref:Microsoft.EntityFrameworkCore.Diagnostics.IDbConnectionInterceptor>. This starts with the ability to configure the `DbContext` without any connection string. For example:
+
+```csharp
+services.AddDbContext<CustomerContext>(
+    b => b.UseSqlServer();
+```
+
+One of the `IDbConnectionInterceptor` methods can then be implemented to configure the connection before it is used. `ConnectionOpeningAsync` is a good choice, since it can perform an async operation to obtain the connection string, find an access token, and so on. For example, imagine a service scoped to the current request that understands the current tenant:
```
roji, line 2473: Doing an async operation every time a connection is opening to obtain a connection string sounds pretty bad... Fetching an access token is also likely to be an anti-pattern in most cases, unless the user makes sure that caching is implemented underneath (i.e. that actual I/O only happens once every hour etc.).

(I think we've discussed this in the past - my personal opinion is that it's better to have some service managing the periodic refreshing of the token via a token, and for the interceptor to only synchronously fetch the latest token from the service.. Documenting (and even providing) the async option feels like a pit of failure to me)

In any case, if we're going to discuss these usages, it's probably a good idea to at least mention that doing an async operation on *every* opening event is a really bad idea.

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2064,3 +2064,558 @@ Configuration of stored procedures many-to-many join entities can be performed a
...
+            InterceptionResult result)
+            => throw new NotSupportedException("Synchronous connections not supported.");
+
+        public override async ValueTask<InterceptionResult> ConnectionOpeningAsync(
+            DbConnection connection, ConnectionEventData eventData, InterceptionResult result,
+            CancellationToken cancellationToken = new())
+        {
+            if (string.IsNullOrEmpty(connection.ConnectionString))
+            {
+                connection.ConnectionString = (await _connectionStringFactory.GetConnectionStringAsync(cancellationToken));
```
roji, line 2530: It's really hard for me to imagine a situation where someone should do this...

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2064,3 +2064,558 @@ Configuration of stored procedures many-to-many join entities can be performed a
...
+
+> [!IMPORTANT]
+> The `ConnectionCreating` and `ConnectionCreated` methods are only called when EF Core creates a `DbConnection`. They will not be called if the application creates the `DbConnection` and passes to EF Core.
+
+Running some code that uses these interceptors show SQL Server query statistics in the log:
+
+```output
+info: Microsoft.EntityFrameworkCore.Database.Command[20101]
+      Executed DbCommand (26ms) [Parameters=[@p0='?' (Size = 4000), @p1='?' (Size = 4000), @p2='?' (Size = 4000), @p3='?' (Size = 4000)], CommandType='Text', CommandTimeout='30']
+      SET STATISTICS IO ON;SET IMPLICIT_TRANSACTIONS OFF;
```
roji, line 2603: Maybe put a newline? :)

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2064,3 +2064,558 @@ Configuration of stored procedures many-to-many join entities can be performed a
...
+
+> [!TIP]
+> This interceptor overrides the non-async `ConnectionOpening` method to throw since the service to get the connection string must be called from an async code path.
+
+### Logging SQL Server query statistics
+
+> [!TIP]
+> The code shown here comes from [QueryStatisticsLoggerSample.cs](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Miscellaneous/NewInEFCore7/QueryStatisticsLoggerSample.cs).
+
+Finally, let's create two interceptors that work together to send SQL Server query statistics to the application log. To generate the statistics, we need an <xref:Microsoft.EntityFrameworkCore.Diagnostics.IDbCommandInterceptor> to do two things.
```
roji, line 2547: Cute idea...!

roji (approved):
@ajcvickers should we be adding such long/comprehensive sections to What's New at this point, as opposed to just adding the real docs for them? We have to do the real docs anyway now, it seems like a lot of duplication...

Apart from that mainly see the comments on async interception for ConnectionOpening and retrieving the connection string/token asynchronously, we should be careful not to show something that would lead people to be very inefficient...

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
ajcvickers: > Doing an async operation every time a connection is opening 

It only happens the first time the connection is opened. So, no more times that currently happens when it is done in `UseSqlServer`.

>  doing an async operation on every opening event is a really bad idea.

Will do this, but just because it's an async call doesn't mean it has to be an async operation. It might, for example, only actually do something async once per hour, or whatever, when it needs to refreshed something cached. Or it might only be async if a cache of connections strings doesn't have an entry for some given client. Regardless, it is common that people need to, at least sometimes, do something async when getting a connection string, and I think it makes sense to show this. Otherwise people will keep doing things like sync-over-async in the constructor, which is worse.

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
ajcvickers: See previous comment.

ajcvickers:
> should we be adding such long/comprehensive sections to What's New at this point

I think so. Different audiences. This is for, "I know EF Core 6.0; show me everything new in EF7." Main docs are for, "I want to use an interceptor for something, show me how to do it. I don't care if it's new or not." All this content is useful in both places.

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
roji: OK. I specifically have never seen a need for remotely accessing connection strings (as opposed to loading them on application start), unless maybe there's a facility that integrates an auth token into them (it always goes back to the auth token).

Yeah, I agree that *if* your async auth token code already includes a cache, it's OK to call this from the interceptor's async method. However, IIRC the last time I looked into this, it was very unclear/undocumented if e.g. the Azure API did, and I suspect lots of users end up doing actual async on every opening.

The optimistic side of me hopes that not providing the async hook would nudge people towards the right solution; the pessimistic/realistic agrees that it may just nudge them towards sync-over-async.

Anyway, maybe worth just adding a note ("verify that you have caching and are not doing I/O every time, otherwise perf will suffer greatly").

roji:
Well, you know my opinion... Assuming we have comprehensive docs (which we need to in any case), the depth of the what's new sections seems way beyond "just show me the new 7.0 stuff". I personally think the huge page size is pretty scary and may be deterring users from actually reading it. And writing long comprehensive docs twice for the same feature, well... We all have our hobbies :)

But anyway, we've discussed this in the past - am OK with merging etc.

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
ajcvickers: How's this?

> [!WARNING]
> Performing a asynchronous lookup for a connection string, access token, or similar every time is is needed can be very slow. Consider caching these things and only refreshing the cached string or token periodically. For example, access tokens can often be used for a significant period of time before needing to be refreshed.

roji: Yep, looks good (just deduplicate the `is is`)

PR #4056, opened 2022-09-24
Update samples to EF7 RC2 and check they all build and run

ajcvickers:
(We may want to hold off merging until RC2 is on NuGet, instead of just in daily builds.)


roji (approved):
Agreed that it's probably better to wait until rc2 is out

roji (approved):

PR #4058, opened 2022-09-26
What's new: queries

ajcvickers:


entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -2627,3 +2627,449 @@ info: Microsoft.EntityFrameworkCore.Database.Command[20101]
...
+
+<!--
+            var query = context.Posts.OfType<Post>();
+-->
+[!code-csharp[OfType](../../../../samples/core/Miscellaneous/NewInEFCore7/MiscellaneousTranslationsSample.cs?name=OfType)]
+
+Which translates to different SQL:
+
+```sql
+      SELECT [p].[Id], [p].[AuthorId], [p].[BlogId], [p].[Content], [p].[Discriminator], [p].[PublishedOn], [p].[Title], [p].[PromoText]
```
roji, line 3004: Huh, I wonder if it makes sense for us to filter on the discriminator with an `IN ('Post', 'FeaturedPost')` post, rather than fetching everything and filtering on the client...

ajcvickers: We normally filter by the discriminator, but this is a query for the base type (there are only two types in the model) so it's like the `OfType` isn't there. Perhaps not the best example to show.

roji: Ohh I see, makes sense. Yeah, using an intermediate hierarchy type may make this clearer.

smitpatel: Yes, we don't filter anything on client.

PR #4067, opened 2022-10-01
What's new: SaveChanges

ajcvickers:


entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3076,3 +3076,243 @@ LEFT JOIN (
...
 ```
+
+## Faster SaveChanges
+
+In EF7, the performance of <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChanges%2A> and <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChangesAsync%2A> has been significantly improved. In some scenarios, saving changes is now four times faster than with EF Core 6.0!
+
+Most of these improvements come from:
+
+- Generating faster SQL
+- Performing fewer roundtrips to the database
```
roji, line 3087: I'd switch the order here :) The fewer roundtrips are more important and universal: the faster SQL is mainly for SQL Server bulk insert, and sometimes for unlocking the fewer roundtrips (through the removal of the transaction).

```suggestion
- Performing fewer roundtrips to the database
- Generating faster SQL
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3076,3 +3076,243 @@ LEFT JOIN (
...
+      INSERT INTO [Blogs] ([Name])
+      OUTPUT INSERTED.[Id]
+      VALUES (@p0);
+```
+
+This removes two database roundtrips, which can make a huge difference to overall performance, especially when the latency of calls to the database is high. In typical production systems, the database is not co-located on the same machine as the application. This means latency is often relatively high, making this optimization particularly effective in real-world production systems.
+
+### Improved SQL for simple Identity insert
+
+The case above inserts a single row with an `IDENTITY` key column and no other database-generated values. EF7 simplifies the SQL in this case by using `OUTPUT INSERTED`. While this simplification is not valid for many other cases, it is still significant because this kind of single-row insert is very common in many applications.
```
roji, line 3142: Yeah, just note that I haven't measured this in isolation, and I suspect that the difference between a single INSERT with `OUTPUT INSERTED` or an INSERT followed by a batched `SELECT SCOPE_IDENTITY` would be *very* similar. The improvement here really is more about unlocking the removal of the transaction (1 statement instead of 2).

(technically, you don't need an insert around INSERT followed by `SELECT SCOPE_IDENTITY`, since the latter select just reads from the connection state and doesn't modify or even read the database. But once you have a non-identity database-generated column things become more complicated...)

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3076,3 +3076,243 @@ LEFT JOIN (
...
+
+### Inserting multiple rows
+
+In EF Core 6.0, the default approach for inserting multiple rows was driven by limitations in SQL Server support for tables with triggers. We wanted to make sure that the default experience worked even for the minority of users with triggers in their tables. This meant that we could not use a simple `OUTPUT` clause, because, on SQL Server, this [doesn't work with triggers](/sql/t-sql/queries/output-clause-transact-sql#triggers). Instead, when inserting multiple entities, EF Core 6.0 generated some fairly convoluted SQL. For example, this call to `SaveChanges`:
```
roji, line 3146: FWIW I still think the EF 7 SQL is pretty convoluted ;) But it's true the removal of the temporary table helps.

roji (approved):

PR #4076, opened 2022-10-10
Finish What's New in EF7

ajcvickers:


entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+
+> [!TIP]
+> The code for samples in this section comes from [ModelBuildingSample.cs](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Miscellaneous/NewInEFCore7/ModelBuildingSample.cs).
+
+### Indexes can be ascending or descending
+
+By default, EF Core creates ascending indexes. EF7 also supports creation of descending indexes. For example:
+
+```csharp
+            modelBuilder
```
roji, line 3565: Unindent? (here and below)

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+                .Entity<Post>()
+                .HasIndex(post => post.Title)
+                .IsDescending();
```
roji, line 3568: The main reason to use descending indexes IIRC is for composite indexes, since a single-column index can be traversed either way anyway; maybe it's good to mention that and have the basic example be composite?

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+    [PrimaryKey(nameof(PostId), nameof(CommentId))]
+    public class Comment
+    {
+        public int PostId { get; set; }
+        public int CommentId { get; set; }
+    }
+-->
+[!code-csharp[CompositePrimaryKey](../../../../samples/core/Miscellaneous/NewInEFCore7/ModelBuildingSample.cs?name=CompositePrimaryKey)]
+
+Defining the index on the class also means it can be used to specify private properties or fields as keys, even though these would usually be ignored when building the EF model. For example:
```
roji, line 3681: Huh, didn't know we didn't support [Key] on private properties/fields... Any specific reason for that?

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+[PrimaryKey(nameof(_id))]
+public class Tag
+{
+    private readonly int _id;
+}
+```
+
+### `DeleteBehavior` mapping attribute
+
+EF7 introduces a mapping attribute (aka "data annotation") to specify the <xref:Microsoft.EntityFrameworkCore.DeleteBehavior> for a relationship. For example, [required relationships](xref:core/modeling/relationships) are created with <xref:Microsoft.EntityFrameworkCore.DeleteBehavior.Cascade?displayProperty=nameWithType> by default. This can be changed to <xref:Microsoft.EntityFrameworkCore.DeleteBehavior.NoAction?displayProperty=nameWithType> by default using [DeleteBehaviorAttribute](https://github.com/dotnet/efcore/blob/main/src/EFCore.Abstractions/DeleteBehaviorAttribute.cs):
```
roji, line 3693:
```suggestion
EF7 introduces a mapping attribute (aka "data annotation") to specify the <xref:Microsoft.EntityFrameworkCore.DeleteBehavior> for a relationship. For example, [required relationships](xref:core/modeling/relationships) are created with <xref:Microsoft.EntityFrameworkCore.DeleteBehavior.Cascade?displayProperty=nameWithType> by default. This can be changed to <xref:Microsoft.EntityFrameworkCore.DeleteBehavior.NoAction?displayProperty=nameWithType> by default using [DeleteBehavior](https://github.com/dotnet/efcore/blob/main/src/EFCore.Abstractions/DeleteBehaviorAttribute.cs):
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+            public CategoryIdConverter()
+                : base(v => v.Value, v => new(v))
+            {
+            }
+        }
+-->
+[!code-csharp[KeyConverters](../../../../samples/core/Miscellaneous/NewInEFCore7/ValueGenerationSample.cs?name=KeyConverters)]
+
+> [!NOTE]
+> The code here uses `struct` types. This means they have appropriate value-type semantics for use as keys. If `class` types are used instead, then they will need to either override equality semantics or also specify a [value comparer](xref:core/modeling/value-comparers).
```
roji, line 4143:
```suggestion
> The code here uses `struct` types. This means they have appropriate value-type semantics for use as keys. If `class` types are used instead, then they need to either override equality semantics or also specify a [value comparer](xref:core/modeling/value-comparers).
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+### UseSqlServer etc. accept null
+
+It is very common to read a connection string from a configuration file and then pass that connection string to `UseSqlServer`, `UseSqlite`, or the equivalent method for another provider. For example:
+
+```csharp
+    services.AddDbContext<BloggingContext>(options =>
+        options.UseSqlServer(Configuration.GetConnectionString("BloggingDatabase")));
+```
+
+It is also common to pass a connection string [applying migrations](xref:core/managing-schemas/migrations/applying). For example:
```
roji, line 4242:
```suggestion
It is also common to pass a connection string when [applying migrations](xref:core/managing-schemas/migrations/applying). For example:
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+dotnet ef database update --connection "Server=(localdb)\mssqllocaldb;Database=MyAppDb"
+```
+
+Or when using a [Migrations bundle](xref:core/managing-schemas/migrations/applying#bundles).
+
+```dotnetcli
+./bundle.exe --connection "Server=(localdb)\mssqllocaldb;Database=MyAppDb"
+```
+
+In this case, even though the connection string read from configuration will not be used, the application startup code will still attempt to read it from configuration and pass it to `UseSqlServer`. If the configuration is not available, then this results in passing null to `UseSqlServer`. In EF7, this is allowed, as long as the connection string is ultimately set later, such as by passing `--connection` to the command-line tool.
```
roji, line 4254:
```suggestion
In this case, even though the connection string read from configuration isn't used, the application startup code still attempts to read it from configuration and to pass it to `UseSqlServer`. If the configuration is not available, then this results in passing null to `UseSqlServer`. In EF7, this is allowed, as long as the connection string is ultimately set later, such as by passing `--connection` to the command-line tool.
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+}
+```
+
+EF Core sets the `IsDesignTime` to `true` when application code is running on behalf of tools.
+
+## Performance enhancements for proxies
+
+EF Core supports dynamically generated proxies for [lazy-loading](xref:core/querying/related-data/lazy) and [change-tracking](xref:core/change-tracking/change-detection#change-tracking-proxies). EF7 contains two performance improvements when using these proxies:
+
+- The proxy types are now created lazily. This means that the initial model building time when using proxies can be massively faster with EF7 than it was with EF Core 6.0.
```
roji, line 4276: Massively :muscle::muscle::muscle:

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+|-------------------------------------------------------------|------------------|--------:|---------:|---------:|
+| EF Core 6.0 without proxies                                 | TimeToFirstQuery | 1.085 s | 0.0083 s | 0.0167 s |
+| EF Core 6.0 with change-tracking proxies                    | TimeToFirstQuery | 13.01 s | 0.2040 s | 0.4110 s |
+| EF Core 7.0 without proxies                                 | TimeToFirstQuery | 1.442 s | 0.0134 s | 0.0272 s |
+| EF Core 7.0 with change-tracking proxies                    | TimeToFirstQuery | 1.446 s | 0.0160 s | 0.0323 s |
+| EF Core 7.0 with change-tracking proxies and compiled model | TimeToFirstQuery | 0.162 s | 0.0062 s | 0.0125 s |
+
+So, in this case, a model with change-tracking proxies can be ready to execute the first query 80 times faster in EF7 than was possible with EF Core 6.0.
+
+## First-class Windows Forms Data Binding
```
roji, line 4291:
```suggestion
## First-class Windows Forms data binding
```

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
```diff
@@ -3527,3 +3549,783 @@ Notice:
...
+
+The Windows Forms team have been making some [great improvements to the Visual Studio Designer experience](https://devblogs.microsoft.com/dotnet/state-of-the-windows-forms-designer-for-net-applications/). This includes [new experiences for data binding](https://devblogs.microsoft.com/dotnet/databinding-with-the-oop-windows-forms-designer/) that integrates well with EF Core.
+
+In brief, the new experience provides Visual Studio U.I. for creating an <xref:System.Web.UI.WebControls.ObjectDataSource>:
+
+![Choose Category data source type](../../get-started/_static/winforms-choose-category-type.png)
+
+This can then be bound to an EF Core `DBset` with some simple code:
```
roji, line 4299:
```suggestion
This can then be bound to an EF Core `DbSet` with some simple code:
```

roji (approved):

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
ajcvickers: I did not realize this! Thanks!

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
ajcvickers: We do...but only if the member is already mapped. It won't cause it to be mapped.

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
roji: (Should probably confirm this across database at some point! But it makes sense...)

entity-framework/core/what-is-new/ef-core-7.0/whatsnew.md
roji: Should we consider enabling this (i.e. mapping the member because it has [Key] on it)? Or would that potentially map something unwanted (because [Key] is being used for some other non-EF framework)?

ajcvickers: Yes, and yes. :-D I made a note to file an issue; it's something I have been thinking about for a while, and something that Diego had ideas on.

PR #4090, opened 2022-10-18
Save future generations time

migueldeicaza:


entity-framework/core/modeling/concurrency.md
```diff
@@ -7,6 +7,8 @@ uid: core/modeling/concurrency
 ---
 # Concurrency Tokens
 
+These are not supported on SQLite or Postgress databases (https://github.com/dotnet/efcore/issues/22193, https://github.com/dotnet/efcore/issues/2195)
```
roji, line 10: @migueldeicaza that's not quite true...

First, there's stored-generated concurrency tokens (like SQL Server's rowversion, detailed below) and user-managed "manual" concurrency tokens (like the last name in the first example). User-managed tokens don't depend on any database feature, and are always supported (it's just about making EF add a check for that column in the WHERE clause).

It's true that store-generated tokens aren't supported by SQLite (hence there's the note "The precise details depend on the database provider being used"). But they are supported in PostgreSQL ([see the docs](https://www.npgsql.org/efcore/modeling/concurrency.html)).

Though I agree we could improve this page to make the stored-generated vs. user-managed distinction more clear, and also make it more clear that store-generated may not be supported depending on the database in use...

roji: One more thought - a store-generated concurrency token can probably be done in SQLite with a trigger that e.g. generates a new random GUID (or increases an int) every time a row is updated.

roji:
Am closing this in favor of #4093, which redoes a lot of documentation around concurrency handling and makes it clearer that stored-generated rowversion-like concurrency tokens are database-specific.

PR #4151, opened 2022-11-18
Highlight split query drawback with reference navigations

stevendarby:
Draws attention to drawback raised in https://github.com/dotnet/efcore/issues/29182


roji:
@stevendarby apologies, I stumbled upon this and noticed we haven't reviewed it...

Do you think this is still something we should merge, especially after the recent changes in #4201? I generally prefer to keep docs concise and not overburden users in the name of completeness - the added JOINs for the reference navigations may be in danger of doing that a bit.

stevendarby:
@roji I think I had two motivations for updating the SQL:

1. Primarily, the drawback I added at the bottom of the article is a bit easier to show than to explain, so the updated example query helps to demonstrate it. 

2. I think that if a unit/integration test or sample code *somewhere* in EF code/docs had demonstrated this behaviour of split query, it would have been fixed early on- as a bug. So I think there is value in having slightly more complete examples at the expense of conciseness.

But I do take your point and would be happy to take it out if you think best. Do you think adding the drawback point on its own is ok though?

roji:
Apologies once again for letting this sit so long. I've looked at this again, and yeah, I still believe we shouldn't make the samples heavier for what is essentially an under-optimized case that we'd like to see fixed. I'm keeping the "drawback" note as you suggested though.

roji (approved):

PR #4153, opened 2022-11-21
Fix page content being part of tabbed content

jvandertil:
Value generation and other page content are being rendered inside the tabbed code sample.

See below:

image


jvandertil:
I am not sure if this is the correct syntax to fix the issue

roji (approved):
Good catch, and thanks! A significant part of that page is hidden because of this bug...

PR #4162, opened 2022-11-26
avoid exception handling when it adds no value

SimonCropp:


ajcvickers:
@SimonCropp As far as I can tell, there is a reason for the try/catch blocks in each of these cases. Can you provider more details why you think they should be removed?

PR #4177, opened 2022-12-07
Add new extension

ErikEJ:


ErikEJ:
>> but doesn't provide any query translations for methods/members on DateOnly/TimeOnly

Hmm...  Never thought of that, I have created https://github.com/ErikEJ/EFCore.SqlServer.DateOnlyTimeOnly/issues/1 and will investigate

roji:
You can check out https://github.com/dotnet/efcore/pull/24989 for the translations (and tests!) done for SQLite, I recommend looking at the PG provider too, it should have more extensive query translations.

ErikEJ:
Thanks  I will do that

PR #4184, opened 2022-12-14
Fix EF8 links

ajcvickers:


entity-framework/toc.yml
```diff
@@ -64,7 +64,15 @@
         href: core/what-is-new/index.md
       - name: Release planning process
         href: core/what-is-new/release-planning.md
-      - name: EF Core 7.0
+      - name: EF Core 8 (EF8)
+        items:
+          - name: High-level plan
+            href: core/what-is-new/ef-core-8.0/plan.md
+          - name: "What's new?"
```
roji, line 71: We could also not add What's new and Breaking changes to the TOC until we actually have content, but not important.

roji (approved):

Issue #4157, opened 2022-11-23
Consider merging doc PRs directly to live


roji:
We currently merge doc PRs to main, and periodically merge main into live to actually make changes appear on-line. I'm not sure what the extra process adds, given that the PR was already approved before merging to main, and it delays doc changes and adds extra process.

PR #4185, opened 2022-12-14
Merge to live (again)

ajcvickers:


roji (approved):
We should really just do #4157, AFAIK there wasn't any technical reason for this.

ajcvickers:
Have we discussed it yet?

roji:
Not really, we've been deferring :)

PR #4208, opened 2023-01-13
Rename identifier in ThenInclude method (Cartesian explosion section)

YuraRov:
Hi team,

This is my first contribution to such a big project. I hope that these changes will clarify and supplement the article.

1. Propose to rename variables from one letter style to the whole word. 'blog' and 'post' are not as long as they could be. By the way, another sections contain the whole word naming.
2. The main change is in the ThenInclude method. For some people, the letter "b" here can be confusing, so this modification should clarify that Comments is a collection navigation of Post type, not of Blog type.

Thank you in advance 😃

### DIFF
--- a/entity-framework/core/querying/single-split-queries.md
+++ b/entity-framework/core/querying/single-split-queries.md
@@ -17,8 +17,8 @@ Let's examine the following LINQ query and its translated SQL equivalent:
 
 ```c#
 var blogs = ctx.Blogs
-    .Include(blog => blog.Posts)
-    .Include(blog => blog.Contributors)
+    .Include(b => b.Posts)
+    .Include(b => b.Contributors)
     .ToList();
 ```
 
@@ -36,8 +36,8 @@ Note that cartesian explosion does not occur when the two JOINs aren't at the sa
 
 ```c#
 var blogs = ctx.Blogs
-    .Include(blog => blog.Posts)
-    .ThenInclude(post => post.Comments)
+    .Include(b => b.Posts)
+    .ThenInclude(p => p.Comments)
     .ToList();
 ```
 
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/EntityFramework.Docs?pullRequest=4208) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/EntityFramework.Docs?pullRequest=4208) before we can accept your contribution.<br/><br/>:x: YuraRov [sign now](https://cla.dotnetfoundation.org/dotnet/EntityFramework.Docs?pullRequest=4208)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/EntityFramework.Docs?pullRequest=4208) it.</sub>

roji:
I think it's pretty standard to use single-letter identifiers for lambda parameters; the scope of that parameter is very restricted, so I don't think there's much potential for confusion.

Unless I'm mistaken, most of our docs do use single-letter lambda parameter names; our [getting started sample](https://learn.microsoft.com/en-us/ef/core/#querying) is an example. It's certainly possible we're not 100% consistent and use a full name somewhere, but I'd rather we made it consistent to single-letter rather than the other direction.

YuraRov:
@roji, thank you for your feedback. Yes, I also agree with you about consistency and single-letter identifiers for lambda parameters. Then, just a little change left in the ThenInclude method. Actually, it was the main goal of this request☺️. The single-letter style is OK but for someone the letter 'b' can be associated with the Blog type. That's why I propose to change it to letter 'p' for clarifying that Comments is a collection navigation of Post type😄

roji (approved):
Makes sense, thanks!

PR #4216, opened 2023-01-18
Clarify that both filtered include examples are valid

ascott18:
Previous wording made it sound like only the second example was the correct way to do things.


ajcvickers (approved):

ajcvickers:
Thanks!

Issue #4032, opened 2022-09-16
Document ExecuteDelete/ExecuteUpdate


smitpatel:
* Update the "efficient updating" page, there's a bulk update section there.

PR #4224, opened 2023-01-25
Introduce bulk update and bulk delete in EF Core 7

yanxiaodi:
EF Core 7 enables bulk update/delete already. So it should be added to the documents.

### DIFF
--- a/entity-framework/core/performance/efficient-updating.md
+++ b/entity-framework/core/performance/efficient-updating.md
@@ -41,6 +41,22 @@ UPDATE [Employees] SET [Salary] = [Salary] + 1000;
 
 This performs the entire operation in a single roundtrip, without loading or sending any actual data to the database, and without making use of EF's change tracking machinery, which imposes an additional overhead.
 
-Unfortunately, EF doesn't currently provide APIs for performing bulk updates. Until these are introduced, you can use raw SQL to perform the operation where performance is sensitive:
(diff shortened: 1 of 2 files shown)


roji:
This is #4032. Note that I have partially written docs for this (which I haven't completed yet); we need to update more than just the perf page - also the "save data" section of the docs.

yanxiaodi:
Thanks @roji I didn't notice that issue. Sure we can close this PR and wait for your update. Just think that users may not get the new features because of the outdated doc. Thanks!

roji:
@yanxiaodi we do have pretty extensive documentation in the [7.0 What's New page](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/whatsnew#executeupdate-and-executedelete-bulk-updates), though it's true we also need to integrate this into our regular docs.

yanxiaodi:
Agree @roji. When will complete the doc? Should we just add a link to [7.0 What's New page](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/whatsnew#executeupdate-and-executedelete-bulk-updates) to avoid any confusion? I can make changes on this PR. Thanks.

ajcvickers:
Note from triage: we will wait for PR from @roji. Thanks anyway, @yanxiaodi!

PR #4245, opened 2023-02-07
Add DateOnly/TimeOnly to What's New in EF8

ajcvickers:


entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -493,6 +493,181 @@ CREATE TABLE [Documents] (
 > [!TIP]
 > Fibonacci numbers are used to limit the number of times a migration is generated to change the column length as new types are added to the hierarchy.
 
+### DateOnly/TimeOnly supported on SQL Server
+
+The <xref:System.DateOnly> and <xref:System.TimeOnly> types were introduced in .NET 6 and have been supported for several database providers (e.g. SQLite, MySQL, and PostgreSQL) since their introduction. For SQL Server, the recent release of a [Microsoft.Data.SqlClient](https://www.nuget.org/packages/Microsoft.Data.SqlClient/) package targeting .NET 6 has allowed [ErikEJ to add support for these types at the ADO.NET level](https://github.com/dotnet/SqlClient/pull/1813). This in turn paved the way for support in EF8 for `DateOnly` and `TimeOnly` as properties in entity types.
+
```
ErikEJ, line 499: Mention support is also available for EF Core 6/7 ? :smile:

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -493,6 +493,181 @@ CREATE TABLE [Documents] (
...
+    public TimeOnly? ClosesAt { get; set; }
+}
+-->
+[!code-csharp[BritishSchools](../../../../samples/core/Miscellaneous/NewInEFCore8/DateOnlyTimeOnlySample.cs?name=BritishSchools)]
+
+> [!TIP]
+> The code shown here comes from [DateOnlyTimeOnlySample.cs](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Miscellaneous/NewInEFCore8/DateOnlyTimeOnlySample.cs).
+
+> [!NOTE]
+> This model represents only British schools and stores times as local (GMT) times. Handling different timezones would complicate this code significantly. Note that using `DateTimeOffset` would not help here, since opening and closing times have different offsets depending whether daylight saving time is active or not.
```
roji, line 542: FWIW unless I'm mistaken, this could be made international simply by introducing a timezone column in the table, so that we know what timezone each school is in. Then, the e.g. opening hour is always interpreted with respect to that timezone. But I agree it's not worth going into that here.

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
roji: Yeah, a note seems to be a good idea.

roji (approved):

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
ajcvickers: What's the recommended way of writing queries against that? Presumably, I would use a UTC time as the parameter, but then convert the column value to UTC in the query? How would I do that conversion--is there a .NET BCL way of doing, or do I need to use a database-specific function? Also, the conversion will be dependent on the day for which the query is running, since it will give a different UTC depending on daylight saving or not. And finally, I would need to be sure that overflow to the next/previous day didn't return incorrect results.

I gave up! :-)

roji: Ahahaha yeah time zone stuff is always awesome :)

So that's why we added [EF.Functions.AtTimeZone](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/whatsnew#support-for-at-time-zone) for SQL Server. The PG provider does translate various BCL APIs because they're a good fit ([see these docs](https://www.npgsql.org/efcore/mapping/translations.html#date-and-time-functions), scroll down a bit).

But you're right that you can't do time zone conversion on just a time - you need a full timestamp (8:00AM converts to different times in other time zones e.g. based on daylight savings). So if the goal is e.g. to get a UTC time for the school opening (for whatever reason), the query would need to construct a timestamp from the database time - providing some date as appropriate - convert that to UTC (with AtTimeZone), and then extract the time component out of that. I think we should have all the building blocks for that, but I'm not 100% sure.

(either way, I definitely think we\ don't need to go into all this complexity in an intro to DateOnly/TimeOnly!).

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
ajcvickers: Done

PR #4260, opened 2023-02-17
Update spatial.md

wstaelens:
it was not clear


ErikEJ (commented):
The Nettopologysuite reference is not required. Also, make it a note.

roji:
Just above is the "Installing" section, which seems very clear:

> In order to use spatial data with EF Core, you need to install the appropriate supporting NuGet package. Which package you need to install depends on the provider you're using.

IMHO the repetition of the nuget package doesn't seem justified here (and as @ErikEJ wrote, depending on NetTopologySuite directly isn't needed).

PR #4311, opened 2023-04-03
Update one-to-one.md

Kumima:
Fix a little mistake.


Kumima:
@dotnet-policy-service agree

PR #4335, opened 2023-04-29
Update modeling-for-performance.md

Sergi0Martin:
Add link to materialized view docs.


Sergi0Martin:
> 

@dotnet-policy-service agree

roji:
We don't typically link to specific documentation pages on general database concepts like materialized views, indexes, etc. - users can easily find resources on that on the web. You're proposing linking specifically to Azure documentation, although EF isn't specific to that in any way; in other words, it's not clear why we'd link to that rather than e.g. the PostgreSQL documentation on materialized views.

Sergi0Martin:
> We don't typically link to specific documentation pages on general database concepts like materialized views, indexes, etc. - users can easily find resources on that on the web. You're proposing linking specifically to Azure documentation, although EF isn't specific to that in any way; in other words, it's not clear why we'd link to that rather than e.g. the PostgreSQL documentation on materialized views.

Nice to know. So I prceed to close the PR.

PR #4346, opened 2023-05-14
What's new in EF8 Preview 4

ajcvickers:
Primitive collections

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
+++ b/entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
@@ -1146,7 +1146,7 @@ Starting with Preview 4, EF8 now includes built-in support for the second option
 
 #### Primitive collection properties
 
-EF Core can map any `IEnumerable<T>` property, where `T` is a primitive type, to a JSON column in the database. This is done by convention for public properties which have both a getter and a setter. For example, all properties  in the following entity type are mapped to JSON columns by convention:
+EF Core can map any `IEnumerable<T>` property, where `T` is a primitive type, to a JSON column in the database. This is done by convention for public properties which have both a getter and a setter. For example, all properties in the following entity type are mapped to JSON columns by convention:
 
 ```csharp
 public class PrimitiveCollections
--- a/entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
+++ b/entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
@@ -2,7 +2,7 @@
 title: What's New in EF Core 8
 description: Overview of new features in EF Core 8
 author: ajcvickers
-ms.date: 03/11/2023
+ms.date: 05/14/2023
 uid: core/what-is-new/ef-core-8
 ---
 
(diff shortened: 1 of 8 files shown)


entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -1127,3 +1127,468 @@ In addition to the enhancements described above, EF8 Preview 2 also includes som
...
+        River,
+        Hills,
+        Village,
+        Park,
+        Beach,
+    }
+-->
+[!code-csharp[DogWalk](../../../../samples/core/Miscellaneous/NewInEFCore8/PrimitiveCollectionsSample.cs?name=DogWalk)]
+
+Like `Pub`, `DogWalk` also contains a collection of the dates visited, and a link to the closest pub since, you know, sometimes the dog needs a saucer of beer after a long walk.
```
ErikEJ, line 1255: LOL

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -1127,3 +1127,468 @@ In addition to the enhancements described above, EF8 Preview 2 also includes som
...
+The Prince of Wales Feathers was visited 5 times in 8 "Ailsworth to Nene" walks.
+The Prince of Wales Feathers was visited 6 times in 9 "Caster Hanglands" walks.
+The Royal Oak was visited 6 times in 8 "Ferry Meadows" walks.
+The White Swan was visited 7 times in 9 "Woodnewton" walks.
+The Eltisley was visited 6 times in 8 "Eltisley" walks.
+Farr Bay Inn was visited 7 times in 11 "Farr Beach" walks.
+Farr Bay Inn was visited 7 times in 9 "Newlands" walks.
+```
+
+Looks like beer and dog walking are a winning combination!
```
ErikEJ, line 1469: Chuckle

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -1127,3 +1127,468 @@ In addition to the enhancements described above, EF8 Preview 2 also includes som
 
 - [Configuration to opt out of occasionally problematic SaveChanges optimizations](https://github.com/dotnet/efcore/issues/29916)
 - [Add convention types for triggers](https://github.com/dotnet/efcore/issues/28687)
+
+## New in EF8 Preview 4
```
roji, line 1131: Should we just have a flat list, do we think it matters which features were introduced in which preview?

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -1127,3 +1127,468 @@ In addition to the enhancements described above, EF8 Preview 2 also includes som
...
+
+### Collections of primitive types
+
+A persistent question when using relational databases is what to do with collections of primitive types? That is, lists or arrays of integers, date/times, strings, and so on? If you're using PostgreSQL, then its easy to store these things using PostgreSQL's [built-in array type](https://www.postgresql.org/docs/current/arrays.html). For other databases, there are two common approaches:
```
roji, line 1135:
```suggestion
A persistent question when using relational databases is what to do with collections of primitive types; that is, lists or arrays of integers, date/times, strings, and so on. If you're using PostgreSQL, then its easy to store these things using PostgreSQL's [built-in array type](https://www.postgresql.org/docs/current/arrays.html). For other databases, there are two common approaches:
```

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -1127,3 +1127,468 @@ In addition to the enhancements described above, EF8 Preview 2 also includes som
...
+
+Starting with Preview 4, EF8 now includes built-in support for the second option, using JSON as the serialization format. JSON works well for this since modern relational databases include built-in mechanisms for querying and manipulating JSON, such that the JSON column can, effectively, be treated as a table when needed, without the overhead of actually creating that table. These same mechanisms allow JSON to be passed in parameters and then used in similar way to table-valued parameters in queries--more about this later.
+
+> [!TIP]
+> The code shown here comes from [PrimitiveCollectionsSample.cs](https://github.com/dotnet/EntityFramework.Docs/tree/main/samples/core/Miscellaneous/NewInEFCore8/PrimitiveCollectionsSample.cs).
+
+#### Primitive collection properties
+
+EF Core can map any `IEnumerable<T>` property, where `T` is a primitive type, to a JSON column in the database. This is done by convention for public properties which have both a getter and a setter. For example, all properties  in the following entity type are mapped to JSON columns by convention:
```
roji, line 1149: The public part with getter and setter doesn't seem specific to primitive collections, so maybe can be removed?

```suggestion
EF Core can map any `IEnumerable<T>` property, where `T` is a primitive type, to a JSON column in the database. This is done by convention for public properties which have both a getter and a setter. For example, all properties in the following entity type are mapped to JSON columns by convention:
```

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -1127,3 +1127,468 @@ In addition to the enhancements described above, EF8 Preview 2 also includes som
...
+> What do we mean by "primitive type" in this context? Essentially, something that the database provider knows how to map, using some kind of value conversion if necessary. For example, in the entity type above, the types `int`, `string`, `DateTime`, `DateOnly` and `bool` are all handled without conversion by the database provider. SQL Server does not have native support for unsigned ints or URIs, but `uint` and `Uri` are still treated as primitive types because there are [built-in value converters](xref:core/modeling/value-conversions#built-in-converters) for these types.
+
+By default, EF Core uses an unconstrained Unicode string column type to hold the JSON, since this protects against data loss with large collections. However, on some database systems, such as SQL Server, specifying a maximum length for the string can improve performance. This, along with other column configuration, can be done [in the normal way](xref:core/modeling/entity-properties). For example:
+
+```csharp
+modelBuilder
+    .Entity<PrimitiveCollections>()
+    .Property(e => e.Booleans)
+    .HasMaxLength(1024)
+    .IsUnicode(false);
```
roji, line 1174: Maybe consider not showing IsUnicode(false).. I get that we want to show facet config in general, but some people may copy paste the whole thing or think that it's somehow needed for max length

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -1127,3 +1127,468 @@ In addition to the enhancements described above, EF8 Preview 2 also includes som
...
+    configurationBuilder
+        .Properties<List<DateOnly>>()
+        .AreUnicode(false)
+        .HaveMaxLength(4000);
+}
+```
+
+#### Queries with primitive collections
+
+Let's look at some of the queries that make use of collections of primitive types. For this, we'll need a simple model with two entity types. The first represents a [British public house](https://en.wikipedia.org/wiki/Pub), or "pub":
```
roji, line 1199: Your model is indeed cooler than mine :)

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -1127,3 +1127,468 @@ In addition to the enhancements described above, EF8 Preview 2 also includes som
...
+```sql
+SELECT [p].[Name]
+FROM [Pubs] AS [p]
+WHERE EXISTS (
+    SELECT 1
+    FROM OpenJson([p].[DaysVisited]) AS [d]
+    WHERE DATEPART(year, CAST([d].[value] AS date)) = @__thisYear_0)
+```
+
+Notice that the query makes use of the date-specific function `DATEPART` here because EF _knows that the primitive collection contains dates_. It might not seem it, but this is actually really important. Because EF knows what's in the collection, it can generate appropriate SQL to use the typed values with parameters, functions, other columns etc.
```
roji, line 1393:
```suggestion
Notice that the query makes use of the date-specific function `DATEPART` here because EF _knows that the primitive collection contains dates_. It might not seem like it, but this is actually really important. Because EF knows what's in the collection, it can generate appropriate SQL to use the typed values with parameters, functions, other columns etc.
```

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
```diff
@@ -1127,3 +1127,468 @@ In addition to the enhancements described above, EF8 Preview 2 also includes som
...
+The Eltisley was visited 6 times in 8 "Eltisley" walks.
+Farr Bay Inn was visited 7 times in 11 "Farr Beach" walks.
+Farr Bay Inn was visited 7 times in 9 "Newlands" walks.
+```
+
+Looks like beer and dog walking are a winning combination!
+
+#### Primitive collections in JSON documents
+
+In all the examples above, column for primitive collection contains JSON. However, this is not the same as mapping [an owned entity type to a column containing a JSON document](xref:core/what-is-new/ef-core-7#json-columns), which was introduced in EF7. But what if that JSON document itself contains a primitive collection? Well, all the queries above still work in the same way! For example, imagine we move the _days visited_ data into an owned type `Visits` mapped to a JSON document:
```
roji, line 1473: Yeah, the below works but we don't really support this properly 😬 Since the primitive array is encoded as a string within the owned JSON document, some stuff doesn't work correctly. It may be worth at least mentioning this problem (and maybe even leave this section out until we fix this)

roji (approved):
Looks good! See some nits, no strong feelings on any of them.

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
ajcvickers: Not sure. We should discuss.

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
ajcvickers: Doesn't it make sense to not use Unicode for a JSON list of bool values?

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
ajcvickers: Will merge for now, and we can discuss. I was impressed that all the queries I tried actually worked. Hopefully nobody looks at the actual JSON. ;-)

entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
roji: For bools (non-strings in general) yeah, you're right - it's just that before you discuss only the max length part. Ok with it as-is as well.

PR #4358, opened 2023-05-23
Update `samples/core/Saving/Transactions/SharingTransaction.cs`

n0099:
https://learn.microsoft.com/en-us/ef/core/saving/transactions#share-connection-and-transaction

User might forget to invoke `DbContext.SaveChanges()` on the derived contexts, like https://stackoverflow.com/questions/41020221/transaction-between-contexts/41020690#41020690


n0099:
@dotnet-policy-service agree

ajcvickers (approved):

ajcvickers:
Thanks!

PR #4362, opened 2023-05-26
Update the message below Deleting Data

szentgyorgyimate:
The message below Deleting Data incorrectly referenced DbSet<TEntity>.Add instead of the DbSet<TEntity>.Remove method.


szentgyorgyimate:
@dotnet-policy-service agree

roji (approved):
Thanks!

PR #4382, opened 2023-06-15
Update navigations.md

ddal:
Dashes corrected.

### DIFF
--- a/entity-framework/core/modeling/relationships/navigations.md
+++ b/entity-framework/core/modeling/relationships/navigations.md
@@ -17,7 +17,7 @@ EF Core relationships are defined by [foreign keys](xref:core/modeling/relations
 
 ## Reference navigations
 
-Navigations come in two forms--reference and collection. Reference navigations are simple object references to another entity. They represent the "one" side(s) of [one-to-many](xref:core/modeling/relationships/one-to-many) and [one-to-one](xref:core/modeling/relationships/one-to-one) relationships. For example:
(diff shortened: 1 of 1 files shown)


bricelam:
IMHO, this is should happen automatically in the docs pipeline. They should honor [SmartyPants notation](https://daringfireball.net/projects/smartypants/) in markdown files.

entity-framework/core/modeling/relationships/navigations.md
```diff
@@ -17,7 +17,7 @@ EF Core relationships are defined by [foreign keys](xref:core/modeling/relations
 
 ## Reference navigations
 
-Navigations come in two forms--reference and collection. Reference navigations are simple object references to another entity. They represent the "one" side(s) of [one-to-many](xref:core/modeling/relationships/one-to-many) and [one-to-one](xref:core/modeling/relationships/one-to-one) relationships. For example:
+Navigations come in two forms — reference and collection. Reference navigations are simple object references to another entity. They represent the "one" side(s) of [one-to-many](xref:core/modeling/relationships/one-to-many) and [one-to-one](xref:core/modeling/relationships/one-to-one) relationships. For example:
```
bricelam, line 20: If we do this, we should probably use the HTML entity instead (assuming it's supported)
```suggestion
Navigations come in two forms&mdash;reference and collection. Reference navigations are simple object references to another entity. They represent the "one" side(s) of [one-to-many](xref:core/modeling/relationships/one-to-many) and [one-to-one](xref:core/modeling/relationships/one-to-one) relationships. For example:
```

roji: Yuck :( As this is markdown shouldn't we be able to just user a plain regular dash and have it converted (or whatever is needed) when that gets rendered as HTML?

ajcvickers:
@ddal After much debate we decided to go with ` - ` everywhere. That is space, hyphen, space.

PR #4386, opened 2023-06-16
Clarify interation between MARS and Savepoints

### LINKED ISSUE #4377: Note about Save Points and MARS is unclear
In the text that states the following:
> Savepoints are incompatible with SQL Server's Multiple Active Result Sets, and are not used. If an error occurs during SaveChanges, the transaction may be left in an unknown state.

The statement "are not used" is unqualified and therefore seems to imply that Savepoints are _never_ used under any circumstance. It would be nice if this could be clarified. For example, "... and are not used when the connection to SQL Server has Multiple Active Result Sets enabled", or some other appropriate description of what the actual functionality is here.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 1073c238-9389-d826-a696-f1108a4f2ce0
* Version Independent ID: cf56babd-def0-d76e-8eee-e33341b6a009
* Content: [Transactions - EF Core](https://learn.microsoft.com/en-us/ef/core/saving/transactions)
* Content Source: [entity-framework/core/saving/transactions.md](https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/saving/transactions.md)
* GitHub Login: @roji

ascott18:
Resolves #4377


roji (approved):
Thanks!

PR #4410, opened 2023-07-05
Create version-sweep.yml GitHub Action workflow.

IEvangelist:
I'm proposing that the EF docs rely on this GitHub Action workflow as so many other docs repos do. It will run on the first of every month, and conditionally create an issue to help ensure that the sample source code projects remain up-to-date.

https://github.com/dotnet/docs/issues/36108

### DIFF
--- a/.github/workflows/version-sweep.yml
+++ b/.github/workflows/version-sweep.yml
@@ -1,0 +1,47 @@
+# This is a basic workflow to help you get started with Actions
+
+name: "target supported version"
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the default branch
+  schedule:
+    - cron: "0 0 1 * *"
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: "The reason for running the workflow"
+        required: true
+        default: "Manual run"
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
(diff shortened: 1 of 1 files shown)


roji:
@IEvangelist is this basically like dependabot, for updating nuget dependencies in the sample projects?

IEvangelist:
> @IEvangelist is this basically like dependabot, for updating nuget dependencies in the sample projects?

Not like dependabot, no. Dependabot does `PackageReference` updates, where as this focuses on TFMs. This examines all the .NET projects, solutions, or Dockerfiles and determines if that target framework monikers (TFMs) are targeting a soon to be out of support version. If so, an issue is created to track the upgrading of said TFM.

roji:
OK. We do a single pass over all samples and update TFMs once a year to the latest, so I'm not sure this would be that valuable.

IEvangelist:
> OK. We do a single pass over all samples and update TFMs once a year to the latest, so I'm not sure this would be that valuable.

Ok. No worries. This would just help to automate that, and at the very least serve as a means for tracking that effort and acting as a reminder. Eventually, this tool will instead generate PRs to automate the update of TFMs, which could save you time.

ajcvickers:
@IEvangelist Thanks for the PR, but as Shay says, it probably isn't quite what we need.

PR #4420, opened 2023-07-20
Remove ellipses from commands making them valid

aateeque:
The ellipses contained within commands on this doc make them invalid so you can't copy paste and follow along this introduction doc. I think the expectation with these ellipses being there is to say _add other parameters here that you may wish_. However, this isn't described anywhere and for anyone following along can be quite confusing to suddenly not have commands not work as expected

### DIFF
--- a/entity-framework/core/managing-schemas/scaffolding/index.md
+++ b/entity-framework/core/managing-schemas/scaffolding/index.md
@@ -145,13 +145,13 @@ For example, to scaffold only the `Artists` and `Albums` tables:
 #### [.NET CLI](#tab/dotnet-core-cli)
 
 ```dotnetcli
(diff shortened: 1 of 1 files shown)


roji (commented):
dbcontext scaffold requires a connection string and a provider name to use, so these command-lines wouldn't work if copy-pasted; in that sense it's better to leave an ellipsis to make it clear that they're not copy-pastable.

We could add something like `<CONNECTION> <PROVIDER>` but I'm not sure that's very helpful; at the end of the day the CLI command will helpfully tell you what's missing etc.

ajcvickers:
Note from triage: we don't believe this makes the docs clearer, so closing.

PR #4425, opened 2023-07-24
Clarify SQL Servers Indexed Views

Suchiman:
SQL Server provides something similiar but with a distinct feature set from Postgres Materialized Views. Pointing out the difference, especially between automatic and manual refreshs, which can be a deal breaker.


roji:
@Suchiman so it took us so long to review this and thanks for pointing this out. I rewrote the materialized views section a bit and gave some place to SQL Server index views based on your suggestions.

roji:
@ajcvickers can you please review my changes?

ajcvickers (approved):

PR #4458, opened 2023-08-23
Reposition Training module on overview hub

CamSoper:
Per email thread with @kjo-msft

[Internal preview](https://review.learn.microsoft.com/en-us/ef/?branch=pr-en-us-4458)

### DIFF
--- a/entity-framework/index.yml
+++ b/entity-framework/index.yml
@@ -22,10 +22,6 @@ highlightedContent:
       itemType: get-started
       url: core/get-started/overview/first-app.md
     # Card
-    - title: "Tutorial: Query and manage data using C# (Entity Framework Core)"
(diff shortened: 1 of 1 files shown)


jcjiang (approved):
I like this change. Right now, the top two elements have very similar titles and the path we want most developers to take - at least to start - is the one under "Getting Started." 

You might consider removing the following section from additionalContent, since it is already on the highlightedContent - then we can avoid having the same link in two places on the page (with different titles) and the clutter/confusion that it can create. 

" - url: core/get-started/overview/first-app.md
            text: ".NET Core console app accessing SQLite with EF Core""

kjo-msft:
@CamSoper - it looks like I don't have permissions to sign off on the PR. :-(

PR #4460, opened 2023-08-23
Fix broken links

bricelam:


entity-framework/core/modeling/value-conversions.md
```diff
@@ -792,7 +792,7 @@ Value converters can be used to encrypt property values before sending them to t
 There are a few known current limitations of the value conversion system:
 
 * As noted above, `null` cannot be converted. Please vote (👍) for [GitHub issue #13850](https://github.com/dotnet/efcore/issues/13850) if this is something you need.
-* It isn't possible to query into value-converted properties, e.g. reference members on the value-converted .NET type in your LINQ queries. Please vote (👍) for [GitHub issue #10434](https://github.com/dotnet/efcore/issues/10434) if this is something you need - but considering using a [JSON column](xref:core/what-is-new/ef-core-7.0/whatsnew#json-columns) instead.
+* It isn't possible to query into value-converted properties, e.g. reference members on the value-converted .NET type in your LINQ queries. Please vote (👍) for [GitHub issue #10434](https://github.com/dotnet/efcore/issues/10434) if this is something you need - but considering using a [JSON column](xref:core/what-is-new/ef-core-7#json-columns) instead.
```
roji, line 795: Huh, it seems like we have an inconsistency here between the `uid` (core/what-is-new/ef-core-7) and the URL (which really does have "7.0", https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/whatsnew).

Should we change the UID instead to align with the URL, to keep things consistenct?

roji (approved):

entity-framework/core/modeling/value-conversions.md
bricelam: Yeah, let's do that instead.

PR #4461, opened 2023-08-24
Fixed typo on -No-OnConfiguring parameter

0darkthirty:
Missing hyphen

image


bricelam:
These docs are for the PowerShell command `Scaffold-DbContext` where there is no hyphen. Please refer to the .NET CLI docs when using `dotnet ef` commands.

Issue #1920, opened 2019-11-14
Document EF Core architecture and internal technical details


ajcvickers:
To make it easier for contributors and provider authors to understand the code.

ajcvickers:
@aspnet/efteam Please add anything with internals that it might be worth documenting. I would rather have too many things on this list and then trim it, so treat this as brainstorming list--no bad ideas!

bricelam:
- [ ] Plugins (type mapping, translators, etc.)
- [ ] Migrations
  - [ ] Diffing (especially how provider model attributes become migration operation attributes)
  - [ ] DDL generation
- [ ] Scaffolding/Reverse Engineering
  - [ ] How design-time services are discovered
  - [ ] Reading database metadata
  - [ ] Provider code hooks (UseMyProvider, literals, etc.)
- [ ] How the tools work (dotnet-ef and the PMC cmdlets)

HamedFathi:
Please consider these:

1. How to get current Table/Column/... information from DbContext. Like [this](https://stackoverflow.com/questions/42455279/how-to-get-column-name-and-corresponding-database-type-from-dbcontext-in-entity)

2. How to call the scaffold/reverse tool by code (Not CLI). Like [this](https://romiller.com/2017/02/13/ef-core-1-1-run-reverse-engineer-from-code/)

ssg:
I'm not sure if this counts as "internals" but after EF Core 3.0, satisfying the query compiler to run the code on the server-side has been really time consuming. I don't argue the benefits of the change but it's not a straightforward process. I wish there was a documentation comparing client-side running queries with equivalent server-side running ones.

tudort-iquest:
It would be great to see how the major patterns were implemented (or not) in EF - identity map, data mapper, repository, unit-of-work, lazy loading, association table mapping, metadata mapping, query object etc..

kant2002:
I in general agree with @ssg , but I want to see how do you generate SQL from different kind of queries. All releases of EF Core starting from betas has changes which require rewriting of the queries to satisfy new version. Expecially many issues I see withing queries which involve GROUP BY or other aggregate functions. If you explain how do you generate SQL using your internal model, it would be easier to write LINQ which will conform to your expected set of queries.

kant2002:
Also what is TPT, TPH, TPC and PMC  ?  It is not very clear.

tudort-iquest:
> Also what is TPT, TPH, TPC and PMC ? It is not very clear.
[...]

PR #4473, opened 2023-08-31
Document plug-in APIs

bricelam:
Part of #1920


AndriySvyryd (approved):

roji (commented):
Really nice to see this! Does it makes sense to have it under the "Tools & Extension" section rather than under Misc?

bricelam:
@roji I considered it, but decided that the number of people wanting to write an extension is considerably less than the number of people wanting to write one.

PR #4475, opened 2023-09-02
Update mapping-attributes.md

khebrati:
1 grammatical error : mapping attributes are used to "modify" ( not "modified") configurations.


roji (approved):
Thanks!

PR #4490, opened 2023-09-19
Document 7.0 changes around RETURNING and non-WAL

### LINKED ISSUE #4411: SQLite SaveChanges: Don't use RETURNING unless WAL enabled
As discovered in dotnet/efcore#30851, we can't automatically retry on busy/locked errors from statements with a RETURNING clause. Because of this, we should only use the RETURNING clause when the journaling mode of the current database is WAL. This mode avoids the situation where we can't retry, making it safe to use them.

bricelam:
Resolves #4411

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-7.0/breaking-changes.md
+++ b/entity-framework/core/what-is-new/ef-core-7.0/breaking-changes.md
@@ -2,7 +2,7 @@
 title: Breaking changes in EF Core 7.0 (EF7) - EF Core
 description: Complete list of breaking changes introduced in Entity Framework Core 7.0 (EF7)
 author: ajcvickers
-ms.date: 09/20/2022
+ms.date: 09/20/2023
 uid: core/what-is-new/ef-core-7.0/breaking-changes
 ---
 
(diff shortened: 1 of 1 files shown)


entity-framework/core/what-is-new/ef-core-7.0/breaking-changes.md
```diff
@@ -286,6 +287,39 @@ modelBuilder
...
+
+By default, EF Core now saves changes via a more efficient technique, using the RETURNING clause. Unfortunately, this technique is not able to automatically retry when busy/locked. In a multi-threaded application (like a web app) not using write-ahead logging, it is common to encounter these errors.
+
+#### Why
+
+The simplifications and performance improvements linked to the new method are significant enough that it's important to bring them to users by default. Databases created by EF Core also enable write-ahead logging by default. The SQLite team also recommends enabling write-ahead logging by default.
+
+#### Mitigations
+
+If possible, you should enable write-ahead logging on your database by executing the following command.
```
roji, line 308: I'd add a note reiterating that if you use EF to create the database (presumably the very common case), nothing needs to be done and you're fine.

roji (approved):

PR #4498, opened 2023-09-26
Fixed typo in change-detection.md

Erik-McKelvey:
The word `using` appeared twice, once normally and once in italic. I left the italic one


roji (approved):
Thanks!

PR #4584, opened 2023-12-05
Update version number

mguinness:
In addition to FileBaseContext provider, there are others that need updating after the release of EF Core 8


roji:
@mguinness thanks for the reminder, submitted #4585 to document some other providers.

PR #4586, opened 2023-12-07
Update generated-properties.md

khebrati:
changed "when the column was first inserted" to "when the row was first inserted"


roji (approved):
Thanks!

PR #4599, opened 2023-12-20
Update EF Core version + Fixed link

JonathanMagnan:
In this pull:

- I updated 3 libraries I own to specify they now support EF Core 8
- Fixed the NuGet link for Entity Framework Extensions (was the wrong link)

Question:

- Would you want me on a different pull to go over all libraries and update the version if they support EF Core 8? (By using the convention "EF Core: [minVersion]-[maxVersion]"?

### DIFF
--- a/entity-framework/core/extensions/index.md
+++ b/entity-framework/core/extensions/index.md
@@ -128,15 +128,15 @@ Triggers for EF Core. Respond to changes in your DbContext before and after they
 
 ### Entity Framework Plus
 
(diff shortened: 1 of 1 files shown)


ErikEJ:
LGTM

roji (approved):
Thanks!

roji:
> Would you want me on a different pull to go over all libraries and update the version if they support EF Core 8? (By using the convention "EF Core: [minVersion]-[maxVersion]"?

Sure, if you're up for that! The maintainers of each package sometimes submit PRs to update these, or we do that in batches from time to time.

PR #4601, opened 2023-12-24
no more HashSet<T> in ctor

DickBaker:
the collection is now initialised in the nav member as List<T> and not as HashSet<T> in the ctor

IMHO the ICollection<T> abstraction is also pointless as you lose out on specifics like Count property [having to invoke the expensive Count() method instead]. Yes I agree there is T4 sophistication for experts but novices should fall into the pit of success.


ajcvickers (approved):

ajcvickers:
Thanks!

Issue #4604, opened 2023-12-31
SQL Trigger example is wrong


Peled-Zohar:
The SQL Trigger shown in the [Update timestamp](https://learn.microsoft.com/en-us/ef/core/modeling/generated-properties?tabs=data-annotations#update-timestamp) paragraph as a major flaw in it as it assumes the update can only work on a single row, where as in the real world, SQL Server triggers are statemet-based, not row-based, meaning that the inserted (or deleted) auto-generated table can contain multiple rows. Therefor, a statement such as `SELECT @Id = INSERTED.BlogId
    FROM INSERTED` will cause the trigger to fail in the update was made over multiple rows.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 97d622b2-1e4b-4d76-b3df-fefaa5ba99cb
* Version Independent ID: ec833af2-275e-981f-689f-8c65edfc00cd
* Content: [Generated Values - EF Core](https://learn.microsoft.com/en-us/ef/core/modeling/generated-properties?tabs=data-annotations#datetime-value-generation)
* Content Source: [entity-framework/core/modeling/generated-properties.md](https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/modeling/generated-properties.md)
* Product: **entity-framework**
* Technology: **entity-framework-core**
* GitHub Login: @AndriySvyryd
* Microsoft Alias: **avickers**

roji:
@Peled-Zohar would you like to submit a PR fixing this?

Peled-Zohar:
@roji Yeah, sure, why not? I'll start working on that within the next 30 minutes

Peled-Zohar:
@roji Done. [PR #4606](https://github.com/dotnet/EntityFramework.Docs/pull/4606)

PR #4606, opened 2023-12-31
Update generated-properties.md

Peled-Zohar:
The previous trigger example had a major flaw in it as it assumes the update can only work on a single row, where as in the real world, SQL Server triggers are statement-based, not row-based, meaning that the inserted (or deleted) auto-generated table can contain multiple rows.

I've changed the trigger example to a trigger that will not break in multiple-rows update statements.

Related Issue: [SQL Trigger example is wrong #4604](https://github.com/dotnet/EntityFramework.Docs/issues/4604)

### DIFF
--- a/entity-framework/core/modeling/generated-properties.md
+++ b/entity-framework/core/modeling/generated-properties.md
@@ -97,14 +97,11 @@ BEGIN
 
     IF ((SELECT TRIGGER_NESTLEVEL()) > 1) RETURN;
 
-    DECLARE @Id INT
-
-    SELECT @Id = INSERTED.BlogId
-    FROM INSERTED
-
-    UPDATE dbo.Blogs
+    UPDATE B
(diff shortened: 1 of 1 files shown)


roji (approved):
Thanks!

Peled-Zohar:
@roji This is probably something I don't understand since this is my first time contributing to this project, but I thought the change would so up in documentation once the PR is merged.... Hope it's not rude of me to ask but can you explain why I don't see the change yet? (I have tried deleting temp files of course)

ErikEJ:
@Peled-Zohar I think the docs are only published on a schedule, not after each merge to main

ajcvickers:
It's less of a "schedule" and more "when I do it". :-) I just did it.

Peled-Zohar:
I see, thanks everyone!

PR #4609, opened 2024-01-04
fix l/r in many-to-many.md for UsingEntity method calls

DickBaker:
typical signature overloads have params (joinEntityName, configureRight, configureLeft, configureJoinEntityType) in that order, but all the docs invert the left / right delegate ordering.

This can give rise to join table with conflicting field names and FK navs, so is dangerous for learning "skip navigation".

As Arthur wrote earlier "**the code gets complicated quickly and its easy to make a mistake**", yup!
The Issue #4269  ' Major overhaul of the "relationships" documentation' needs correcting, so I offer this PR.

HTH.


ajcvickers:
Thanks!

PR #4616, opened 2024-01-06
Replaced new() with []

LassazVegaz:
Replaced `new()` with .NET 8 new collection initialization syntax `[]`
Only in [ManyToMany.cs](https://github.com/dotnet/EntityFramework.Docs/blob/main/samples/core/Modeling/Relationships/ManyToMany.cs) which affects [Many-to-Many relationships docs](https://learn.microsoft.com/en-us/ef/core/modeling/relationships/many-to-many).


LassazVegaz:
@dotnet-policy-service agree

roji (approved):
Goes well with https://github.com/dotnet/efcore/pull/32736, which I just now submitted.

PR #4634, opened 2024-01-25
Explanation correction of dependent navigation unavailability

darul9061:
correction on the statement for adding shadow foreign when the dependent navigation is not available


ajcvickers (approved):

ajcvickers:
Thanks for the contribution!

PR #4689, opened 2024-03-20
Update '.NET implementations supported by EF Core' page to include EF Core 8.0

davepcallan:
Small update to include EF Core 8.0 and mention it targets .NET 8.0


roji (approved):
Thanks!

PR #4690, opened 2024-03-21
add virtual keyword to properties

rominator1983:
virtual seems to be missing as pointed out in the comments of this answer: https://stackoverflow.com/a/22708800 I also needed that for the migration to work at all.


roji:
That comment only seems to say that a DbSet property is needed, not that it must necessarily be virtual. You definitely shouldn't need to make DbSet properties virtual; if you're running into issues, can you please open an issue on the main EF Core repo with a full, runnable code sample?

PR #4702, opened 2024-04-07
Update QueryTrackingBehavior benchmark as it fails due to attempts to insert nulls into not nullable db columns

davepcallan:
Without modification this benchmark fails as blog url and post title and content are not marked as nullable ? and we don't explicitly insert into these fields in the SeedData method so SQL Server rejects the insert.

Have amended SeedData to insert dummy data into blog.url, and post.title and post.content. 

Alternative is to remove these fields or mark them as nullable but in the context of tracking v no-tracking, better to keep a more realistic scenario and just populate the fields IMHO.


roji:
@maumar want to take this?

maumar (approved):

maumar:
Thank you for the contribution @davepcallan !

PR #4716, opened 2024-04-19
fix "Unrecognized escape sequence" in snippet

timdeschryver:
Today, a student tried to copy paste the existing snippet, which resulted in the "Unrecognized escape sequence" error.
This PR updates the example to indicate that the connection string is to be interpreted verbatim.

This occurs on the page: https://learn.microsoft.com/en-us/ef/core/providers/?tabs=dotnet-core-cli#adding-a-database-provider-to-your-application


roji (approved):
Thanks!

PR #4719, opened 2024-05-02
Update advanced-performance-topics.md

Tornhoof:
Wrong Link to Benchmark source, it pointed to ContextPooling, not to CompiledQueries.


roji (approved):
Thanks!

PR #4727, opened 2024-05-15
Fix rowversion code sample in 8.0 what's new docs

### LINKED ISSUE #4725: Numeric rowversions doesn't need a HasConversion call
### Type of issue

Missing information

### Description

The section [Numeric rowversions for SQL Azure/SQL Server](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-8.0/whatsnew#numeric-rowversions-for-sql-azuresql-server) is written in a way that implies that the call to `HasConversion<byte[]>()` is required to get this new feature to work.

But from my tests it looks like this works automatically when you have an `ulong` property with the `Timestamp` attribute.
I haven't tested it but I assume the same is true for fluent mapping via `IsRowVersion()`.

So why does the documentation explicitly call `HasConversion<byte[]>()`?
The difference isn't that important with fluent mapping, but in the case of attribute mapping it's very important since there is no way to specify `HasConversion<byte[]>()` as an attribute.

### Page URL

https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-8.0/whatsnew#numeric-rowversions-for-sql-azuresql-server

### Content source URL

https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md

97203c5d-fe52-b71e-dcb3-71c9492b4e04

### Article author

@ajcvickers

roji:
Fixes #4725

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
+++ b/entity-framework/core/what-is-new/ef-core-8.0/whatsnew.md
@@ -2588,7 +2588,7 @@ SQL Server automatic [optimistic concurrency](xref:core/saving/concurrency) is h
 ```csharp
 modelBuilder.Entity<Blog>()
     .Property(e => e.RowVersion)
-    .HasConversion<byte[]>()
+    .HasConversion<ulong>()
     .IsRowVersion();
 ```
 


cremor:
This change doesn't actually improve any of my confusion described in issue #4725

Why do we need a `HasConversion` call? What does it do that isn't already done automatically when no `HasConversion` call is there at all?

roji:
@ajcvickers can you take a quick look at all of this when you're back?

ajcvickers:
@roji If the property is already a `ulong`, then specifying a conversion to `ulong` is pointless. If the conversion to `byte[]` is not now needed, that's great. I'm pretty sure last time I tested this, it was still needed in the fluent API.

PR #4733, opened 2024-05-22
Update advanced-performance-topics.md to match SQL and C# Queries

Raamyy:
At the "Query caching and parameterization" section in the document the C# snippets evolves around "Post" entity while the SQL snippets evolves around "Blog" entity. 

This commit unifies these entities to "Post" entity.

image


roji (approved):
Thanks!

PR #4738, opened 2024-05-30
Add documentation for breaking change.

cincuranet:
For https://github.com/dotnet/efcore/pull/33815.

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-9.0/breaking-changes.md
+++ b/entity-framework/core/what-is-new/ef-core-9.0/breaking-changes.md
@@ -20,10 +20,11 @@ EF Core 9 targets .NET 8. This means that existing applications that target .NET
 
 ## Summary
 
-| **Breaking change**                                                                | **Impact** |
-|:-----------------------------------------------------------------------------------|------------|
-| [Sync I/O via the Azure Cosmos DB provider is no longer supported](#cosmos-nosync) | Medium     |
-| [EF.Functions.Unhex() now returns `byte[]?`](#unhex)                               | Low        |
+| **Breaking change**                                                                                  | **Impact** |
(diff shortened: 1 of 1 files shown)


roji (approved):

entity-framework/core/what-is-new/ef-core-9.0/breaking-changes.md
```diff
@@ -56,3 +57,27 @@ protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
...
+
+Starting with EF Core 9.0, EF now throws if the number of arguments and nullability propagation arguments do not match.
+
+#### Why
+
+Not having matching number of arguments and nullability propagation arguments can lead to unexpected behavior.
+
+#### Mitigations
+
+Make sure the `argumentsPropagateNullability` has same number of elements as the `arguments`. When in doubt use `false` for nullability argument.
```
roji, line 83: Why the "use false when in doubt"? Can't that lead to incorrect results if the function does propagate nulls (which almost all functions do)? (BTW we don't really have to provide a tip on this at all in this breaking note)

roji: I'm guessing maybe I missed a design discussion on this..

cincuranet: If you put `true` you might get wrong results in case you're incorrect. If you put `false` you might get less efficient query, but the result will be correct even if you're incorrect.

@maumar has opinion about including this sentence. ;)

cincuranet: Discussion resolved offline.

PR #4760, opened 2024-07-04
Fixed typo and added some space

Ramo-Y:
Hi there!
I just fixed a typo and added a space where hyphens were right before a word.


roji (approved):
Thanks!

PR #4776, opened 2024-08-05
Update index.md

stamminator:
Clarify use of "hand-code" so it doesn't read as if a model is being handed to code


roji (approved):
Thanks!

Issue #4793, opened 2024-08-31
Bug in Client Evaluation sample code


mamdos:
### Issue or suggestion

I reached to a bug in this address https://github.com/dotnet/EntityFramework.Docs/blob/main/samples/core/Querying/ClientEvaluation/Program.cs#ClientMethod

mamdos:
when the entry URL starts with https:// it also adds another http:// to it and return it, and this produces bugs

roji:
This sample is just a demonstration of a method that executes client-side; this particular code just adds an http:// prefix if there isn't one; it's not meant to be a fully usable method for handling URLs, just a minimal demo. Feel free to change it in your own code to detect https:// or anything else.

PR #4794, opened 2024-08-31
fix issue introduced in issue NO. #4793

mamdos:


roji:
As in https://github.com/dotnet/EntityFramework.Docs/issues/4793#issuecomment-2323205394, this method isn't supposed to be a robust URL detection mechanism - it's just some arbitrary sample code. The fact that it doesn't check for https:// isn't a bug, and I'll point out that adding https:// detection still leaves it incomplete in many ways (for example, URLs can start with yet other protocols, e.g. ftp://).

PR #4828, opened 2024-10-07
whats new in ef9 - migrations

maumar:


entity-framework/core/managing-schemas/migrations/applying.md
```diff
@@ -343,3 +343,23 @@ Note that `Migrate()` builds on top of the `IMigrator` service, which can be use
 >
 > * Carefully consider before using this approach in production. Experience has shown that the simplicity of this deployment strategy is outweighed by the issues it creates. Consider generating SQL scripts from migrations instead.
 > * Don't call `EnsureCreated()` before `Migrate()`. `EnsureCreated()` bypasses Migrations to create the schema, which causes `Migrate()` to fail.
+
+## Migration locking
```
maumar, line 347: is this the right place to put the extra info? @AndriySvyryd

roji: @maumar do we think users need to know about the migration lock, and the implementation difference between SQLite and SQL Server? I do think it makes sense to have a "what's new" section on it (because it's a new feature), but I'm not sure why a user using migrations should need to know that the migration lock exists or how it works...

In contrast, further up on this page there are warnings against using Migrate() because of concurrency on startup - that's exactly what the migration lock fixes, so we should look at amending the page to not scare people away from Migrate() (when using 9.0 and above).

maumar: sounds good, I removed the details section and added a short note instead

entity-framework/core/what-is-new/ef-core-9.0/whatsnew.md
```diff
@@ -1182,6 +1182,68 @@ protected override void Up(MigrationBuilder migrationBuilder)
 }
 ```
 
+<a name="concurrent-migrations"></a>
+
+### Protection against concurrent migrations
+
+EF9 introduces a locking mechanism when executing migrations to protect against multiple migrations happening simultaneously, which could leave the database in a corrupted state. This should be a rare occurrence when migrations are deployed to the production environment using [recommended methods](/ef/core/managing-schemas/migrations/applying#sql-scripts), however it could happen more often if migrations are applied at runtime using the `IMigrator.Migrate()` method.
```
AndriySvyryd, line 1189:
```suggestion
EF9 introduces a locking mechanism to protect against multiple migrations executions happening simultaneously, as that could leave the database in a corrupted state. This should be a rare occurrence when migrations are deployed to the production environment using [recommended methods](/ef/core/managing-schemas/migrations/applying#sql-scripts), however it is more likely to happen if migrations are applied at runtime using the [`DbContext.Database.Migrate()`](/dotnet/api/microsoft.entityframeworkcore.relationaldatabasefacadeextensions.migrate) method.
```

entity-framework/core/what-is-new/ef-core-9.0/whatsnew.md
```diff
@@ -1182,6 +1182,68 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
+
+> [!NOTE]
+> We recommend not to apply migrations at runtime outside of the development cycle, but in some scenarios it is unavoidable, e.g. [working with using .NET Aspire projects](/dotnet/aspire/database/ef-core-migrations).
+
+<a name="warn-when-no-transaction"></a>
+
+### Warn when multiple migration operations can't be run inside a transaction
+
+Majority of operations performed during migrations are protected by a transaction. This ensures that if for some reason migration fails, the database does not end up in a corrupted state. However, some operations are not wrapped in a transaction (e.g. [operations on memory-optimized tables](/sql/relational-databases/in-memory-oltp/unsupported-sql-server-features-for-in-memory-oltp#scenarios-not-supported), or database altering operations like modifying the database collation). To avoid corrupting the database in case of migration failure, it is recommended that these operations are performed in isolation using a separate migration. EF9 now detects a scenario when a migration contains multiple operations, one of which can't be wrapped in a transaction, and issues a warning.
```
AndriySvyryd, line 1198:
```suggestion
The majority of operations performed during migrations are protected by a transaction. This ensures that if for some reason migration fails, the database does not end up in a corrupted state. However, some operations are not wrapped in a transaction (e.g. [operations on memory-optimized tables](/sql/relational-databases/in-memory-oltp/unsupported-sql-server-features-for-in-memory-oltp#scenarios-not-supported), or database-altering operations like modifying the database collation). To avoid corrupting the database in case of a migration failure, it is recommended that these operations are performed in isolation using a separate migration. EF9 now detects a scenario when a migration contains multiple operations, one of which can't be wrapped in a transaction, and issues a warning.
```

entity-framework/core/what-is-new/ef-core-9.0/whatsnew.md
```diff
@@ -1182,6 +1182,68 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
+
+Tring to migrate from the following model:
```
AndriySvyryd, line 1200:
```suggestion
Trying to migrate from the following model:
```

entity-framework/core/managing-schemas/migrations/applying.md
```diff
@@ -320,6 +320,9 @@ It's possible for the application itself to apply migrations programmatically, t
 * It's important to be able to roll back an applied migration in case of an issue. The other strategies provide this easily and out of the box.
 * The SQL commands are applied directly by the program, without giving the developer a chance to inspect or modify them. This can be dangerous in a production environment.
 
+> [!NOTE]
+> EF9 introduced migration locking, which protects against multiple migrations happening simultaneously. However, other problems listed above are relevant and as such applying migration at runtime is still strongly discouraged in production.
```
roji, line 324: I'd maybe make the wording a bit softer (e.g. remove `strongly`).. I think that once locking is implemented, doing this is maybe not that bad (we can discuss what our policy/recommendation is).

In any case, I'd add "On version of EF prior to 9, " at the start of the first bullet point above.

entity-framework/core/what-is-new/ef-core-9.0/whatsnew.md
roji: Re "This should be a rare occurrence when migrations are deployed to the production environment using recommended methods" - shouldn't it never happen at all if you're using the recommended methods? Like if you're using migration bundles or SQL scripts, you're in control and are presumably never running in them in parallel?

roji (changes requested):

entity-framework/core/managing-schemas/migrations/applying.md
maumar: ended up removing most of the superfluous info, but left some around. Added a documentation note on what to do when lock doesn't get lifted on sqlite (due to failed transaction). User needs to manually drop the migration lock table, so little info adds some context why user needs to do this

entity-framework/core/managing-schemas/migrations/applying.md
```diff
@@ -314,7 +314,7 @@ PS C:\local\AllTogetherNow\SixOh>
 
 It's possible for the application itself to apply migrations programmatically, typically during startup. While productive for local development and testing of migrations, this approach is inappropriate for managing production databases, for the following reasons:
 
-* If multiple instances of your application are running, both applications could attempt to apply the migration concurrently and fail (or worse, cause data corruption).
+* On version of EF prior to 9, if multiple instances of your application are running, both applications could attempt to apply the migration concurrently and fail (or worse, cause data corruption).
```
AndriySvyryd, line 317:
```suggestion
* For versions of EF prior to 9, if multiple instances of your application are running, both applications could attempt to apply the migration concurrently and fail (or worse, cause data corruption).
```

entity-framework/core/managing-schemas/migrations/applying.md
```diff
@@ -343,3 +343,10 @@ Note that `Migrate()` builds on top of the `IMigrator` service, which can be use
...
 > * Don't call `EnsureCreated()` before `Migrate()`. `EnsureCreated()` bypasses Migrations to create the schema, which causes `Migrate()` to fail.
+
+## Migration locking
+
+EF9 introduced a locking mechanism when executing migrations. It aims to protect against multiple migration executions happening simultaneously, as that could leave the database in a corrupted state. This is one of the potential problems resulting from applying migrations at runtime using [`DbContext.Database.Migrate()`](/dotnet/api/microsoft.entityframeworkcore.relationaldatabasefacadeextensions.migrate) method. To mitigate this, EF will create an exclusive lock on the database before any migration operations are applied. The lock is released when the migration completes. The locking mechanism used depends on the underlying database provider. On SQL Server, EF takes advantage of the database built-in locking mechanism using stored procedures `sp_getapplock` and `sp_releaseapplock`. Sqlite lacks such mechanism, so instead EF creates a separate table (`__EFMigrationsLock`) and uses it for locking.
```
AndriySvyryd, line 349:
```suggestion
EF9 introduced a locking mechanism when executing migrations. It aims to protect against multiple migration executions happening simultaneously, as that could leave the database in a corrupted state. This is one of the potential problems resulting from applying migrations at runtime using [`DbContext.Database.Migrate()`](/dotnet/api/microsoft.entityframeworkcore.relationaldatabasefacadeextensions.migrate) method. To mitigate this, EF will create an exclusive lock on the database before any migration operations are applied. The lock is released when the migration completes and the seeding code finishes execution. The locking mechanism used depends on the underlying database provider. On SQL Server, EF takes advantage of the database built-in locking mechanism using stored procedures `sp_getapplock` and `sp_releaseapplock`. Sqlite lacks such mechanism, so instead EF creates a separate table (`__EFMigrationsLock`) and uses it for locking.
```

AndriySvyryd: Probably best to move the Sqlite info to the paragraph below

AndriySvyryd:
By the way, do you plan on covering seeding as well?

AndriySvyryd (approved):

entity-framework/core/providers/sqlite/limitations.md
```diff
@@ -90,6 +90,10 @@ Otherwise, we recommend using `dotnet ef database update` to apply migrations. Y
 dotnet ef database update --connection "Data Source=My.db"
 ```
 
+## Concurrent migrations protection
+
+EF9 introduced a locking mechanism when executing migrations. It aims to protect against multiple migration executions happening simultaneously, as that could leave the database in a corrupted state. This is one of the potential problems resulting from applying migrations at runtime using [`DbContext.Database.Migrate()`](/dotnet/api/microsoft.entityframeworkcore.relationaldatabasefacadeextensions.migrate) method. To mitigate this, EF will create an exclusive lock on the database before any migration operations are applied. Sqlite does not have built-in locking mechanism, so EF creates a separate table (`__EFMigrationsLock`) and uses it for locking. The lock is released when the migration completes and the seeding code finishes execution. However, if for some reason migration fails in a non-recoverable way, the lock may not be released correctly. If this happens, consecutive migrations will be blocked from executing SQL and therefore never complete. You can manually unblock them by deleting the `__EFMigrationsLock` table in the database.
```
roji, line 95:
```suggestion
EF9 introduced a locking mechanism when executing migrations. It aims to protect against multiple migration executions happening simultaneously, as that could leave the database in a corrupted state. This is one of the potential problems resulting from applying migrations at runtime using the [`DbContext.Database.Migrate()`](/dotnet/api/microsoft.entityframeworkcore.relationaldatabasefacadeextensions.migrate) method (see [Applying migrations](xref:core/managing-schemas/migrations/applying) for more information). To mitigate this, EF creates an exclusive lock on the database before any migration operations are applied.

Unfortunately, SQLite does not have built-in locking mechanism, so EF creates a separate table (`__EFMigrationsLock`) and uses it for locking. The lock is released when the migration completes and the seeding code finishes execution. However, if for some reason migration fails in a non-recoverable way, the lock may not be released correctly. If this happens, consecutive migrations will be blocked from executing SQL and therefore never complete. You can manually unblock them by deleting the `__EFMigrationsLock` table in the database.
```

entity-framework/core/what-is-new/ef-core-9.0/whatsnew.md
```diff
@@ -1182,6 +1182,68 @@ protected override void Up(MigrationBuilder migrationBuilder)
 }
 ```
 
+<a name="concurrent-migrations"></a>
+
+### Protection against concurrent migrations
+
+EF9 introduces a locking mechanism to protect against multiple migration executions happening simultaneously, as that could leave the database in a corrupted state. This shouldn't happen when migrations are deployed to the production environment using [recommended methods](/ef/core/managing-schemas/migrations/applying#sql-scripts), however it is more likely to occur if migrations are applied at runtime using the [`DbContext.Database.Migrate()`](/dotnet/api/microsoft.entityframeworkcore.relationaldatabasefacadeextensions.migrate) method. We recommend not to apply migrations at runtime outside of the development cycle, but in some scenarios it is unavoidable, e.g. [working with using .NET Aspire projects](/dotnet/aspire/database/ef-core-migrations).
```
roji, line 1189: Applying migrations at runtime isn't unavoidable with Aspire - it's possible to have a separate project managing migrations, and to make that container start up first, applying the migrations in a single place before the app instances are started. However, this is heavier and more complicated to set up.

We really should work with one of the Aspire people to have EF+Aspire docs either here in our docs, or in the Aspire docs, or both (or at least cross-link). /cc @SamMonoRT 

```suggestion
EF9 introduces a locking mechanism to protect against multiple migration executions happening simultaneously, as that could leave the database in a corrupted state. This doesn't happen when migrations are deployed to the production environment using [recommended methods](/ef/core/managing-schemas/migrations/applying#sql-scripts), but can happen if migrations are applied at runtime using the [`DbContext.Database.Migrate()`](/dotnet/api/microsoft.entityframeworkcore.relationaldatabasefacadeextensions.migrate) method. We recommend applying migrations at deployment, rather than as part of application startup, but that can result in more complicated application architectures (e.g. [when using .NET Aspire projects](/dotnet/aspire/database/ef-core-migrations).
```

entity-framework/core/what-is-new/ef-core-9.0/whatsnew.md
```diff
@@ -1182,6 +1182,68 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
+
+> [!NOTE]
+> If you are using Sqlite database, see [potential issues associated with this feature](/ef/core/providers/sqlite/limitations#concurrent-migrations-protection).
+
+<a name="warn-when-no-transaction"></a>
+
+### Warn when multiple migration operations can't be run inside a transaction
+
+The majority of operations performed during migrations are protected by a transaction. This ensures that if for some reason migration fails, the database does not end up in a corrupted state. However, some operations are not wrapped in a transaction (e.g. [operations on memory-optimized tables](/sql/relational-databases/in-memory-oltp/unsupported-sql-server-features-for-in-memory-oltp#scenarios-not-supported), or database altering operations like modifying the database collation). To avoid corrupting the database in case of migration failure, it is recommended that these operations are performed in isolation using a separate migration. EF9 now detects a scenario when a migration contains multiple operations, one of which can't be wrapped in a transaction, and issues a warning.
```
roji, line 1198:
```suggestion
The majority of operations performed during migrations are protected by a transaction. This ensures that if for some reason migration fails, the database does not end up in a corrupted state. However, some operations are not wrapped in a transaction (e.g. [operations on SQL Server memory-optimized tables](/sql/relational-databases/in-memory-oltp/unsupported-sql-server-features-for-in-memory-oltp#scenarios-not-supported), or database altering operations like modifying the database collation). To avoid corrupting the database in case of migration failure, it is recommended that these operations are performed in isolation using a separate migration. EF9 now detects a scenario when a migration contains multiple operations, one of which can't be wrapped in a transaction, and issues a warning.
```

entity-framework/core/what-is-new/ef-core-9.0/whatsnew.md
```diff
@@ -1182,6 +1182,68 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
+
+Trying to migrate from the following model:
```
roji, line 1200: Am not sure a code sample is really needed here... Code samples are great to show people how to do something; but this is a new warning - users don't need to be taught how to trigger it, they'll just encounter it when they create the problematic migration etc. (so the paragraph above seems sufficient to me).

roji (approved):
Thanks, see additional comments, LGTM once those are fixed.

PR #4836, opened 2024-10-16
whats new - seeding

### LINKED ISSUE #2879: Add HasData seeding example for many-to-many relationship
When using many-to-many with generated join entity, the exception we throw is not great, e.g.:

```
The seed entity for entity type 'Tag' cannot be added because it has the navigation 'Posts' set. To seed relationships,  add the entity seed to 
'PostTag (Dictionary<string, object>)' and specify the foreign key values {'TagsTagId'}.
```

all these are generated by EF so it might be nice to show people how to do this (or maybe throw better exception for m2m with generated join entity?)

### LINKED ISSUE #4806: Document seeding
See https://github.com/dotnet/efcore/pull/34338

Also change the references to the `HasData` feature to "Managed data"

maumar:
- rename old seeding (i.e. HasData) to managed data (managing data)
- modifying HasData sample so that it adds static data (which is how we want people to use HasData)
- add example of HasData for Many-to-Many

Fixes #4806
Fixes #2879


roji (commented):
As an alternative to the term "managed data", we should consider calling this "migration-based seeding" or "model-based seeding". I think the term "seeding" does need to remain in the name, because that's the name we've always used (and therefore our users recognize, e.g. there are blog posts out there that use it); "managed data" is also extremely general and could refer also to e.g. Add/SaveChanges.

entity-framework/core/modeling/data-seeding.md
```diff
@@ -11,32 +11,33 @@ Data seeding is the process of populating a database with an initial set of data
 
 There are several ways this can be accomplished in EF Core:
 
-* Model seed data
+* Model managed data
 * Manual migration customization
 * Custom initialization logic
+* Configuration options seeding methods
```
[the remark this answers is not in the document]
AndriySvyryd: Agree, and concerning the renaming of the feature this is the place where we should explain that it was referred to as "seeding" before, but this name set up incorrect expectations, that's why the new name does not contain "seeding".

As to the order of preference. It should be:

1. UseSeeding
2. Custom initialization logic (only if the above is not enough, e.g. you need to seed using two contexts in one transaction)
3. HasData
4. Manual migration customization (if the above is not enough)

entity-framework/core/modeling/data-seeding.md
```diff
@@ -86,3 +87,14 @@ Depending on the constraints of your deployment the initialization code can be e
...
+<a name="use-seeding-method"></a>
+
+### Configuration options `UseSeeding` and `UseAsyncSeeding` methods
+
+EF 9 introduced `UseSeeding` and `UseAsyncSeeding` methods, which provide convenient way of configuring custom initialization logic. These methods can be set up in the [options configuration step](/ef/core/dbcontext-configuration/#dbcontextoptions). Here is an example analogous to the custom initialization showed above:
+
+[!code-csharp[LanguageSeed](../../../samples/core/Modeling/DataSeeding/DataSeedingContext.cs?name=ContextOptionSeeding)]
+
+> !NOTE
+> `UseSeeding` is called from the `EnsureCreated` method, and `UseAsyncSeeding` is called from the `EnsureCreatedAsync` method.  When using this feature, it is recommended to implement both `UseSeeding` and `UseAsyncSeeding` methods using similar logic. This way the seeding code runs for synchronous and asynchrouns scenarios.
```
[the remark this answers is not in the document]
AndriySvyryd: This is a safeguard, as the tooling will only call the sync version. At least until https://github.com/dotnet/efcore/issues/34923 is implemented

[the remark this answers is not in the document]
roji: Ah OK, I forgot about the tooling problem. Should we maybe be more explicit, saying that (currently) sync `UseSeeding` **must** be implemented in order for the tooling to work? The current wording makes it a bit unclear why/when...

entity-framework/core/modeling/data-seeding.md
```diff
@@ -45,18 +46,18 @@ Once the data has been added to the model, [migrations](xref:core/managing-schem
...
+This type of data is managed by migrations and the script to update the data that's already in the database needs to be generated without connecting to the database. This imposes some restrictions:
 
 * The primary key value needs to be specified even if it's usually generated by the database. It will be used to detect data changes between migrations.
-* Previously seeded data will be removed if the primary key is changed in any way.
+* Previously managed data will be removed if the primary key is changed in any way.
 
 Therefore this feature is most useful for static data that's not expected to change outside of migrations and does not depend on anything else in the database, for example ZIP codes.
 
-If your scenario includes any of the following it is recommended to use custom initialization logic described in the last section:
+If your scenario includes any of the following it is recommended to use custom initialization logic (or `UseSeeding` and `UseAsyncSeeding` methods) described in the last section:
```
AndriySvyryd, line 60:
```suggestion
If your scenario includes any of the following it is recommended to use custom initialization logic (or `UseSeeding` and `UseAsyncSeeding` methods) described in the first section:
```

entity-framework/core/what-is-new/ef-core-9.0/whatsnew.md
```diff
@@ -1377,6 +1377,15 @@ FROM "Walks" AS "w"
 INNER JOIN "Pubs" AS "p" ON "w"."ClosestPubId" = "p"."Id"
 ```
 
+### Data seeding
+
+EF9 introduced a convenient way to perform data seeding, that is populating the database with initial data. `DbContextOptionsBuilder` now contains `UseSeeding` and `UseAsyncSeeding` methods which get executed when the DbContext is initialized (as part of `EnsureCreated`).
+
+> !NOTE
+> If the application had ran previously, the database may already contain the sample data (which would have been added on the first initialization of the context). As such, `UseSeeding` `UseAsyncSeeding` should check if data exists before attempting to populate the database. This can be achieved by issueing a simple EF query.
```
AndriySvyryd, line 1385:
```suggestion
> If the application had ran previously, the database may already contain the sample data (which would have been added on the first initialization of the context). As such, `UseSeeding` `UseAsyncSeeding` should check if data exists before attempting to populate the database. This can be achieved by issuing a simple EF query.
```

entity-framework/core/modeling/data-seeding.md
```diff
@@ -11,32 +11,33 @@ Data seeding is the process of populating a database with an initial set of data
...
-[!code-csharp[BlogSeed](../../../samples/core/Modeling/DataSeeding/DataSeedingContext.cs?name=BlogSeed)]
+[!code-csharp[CountrySeed](../../../samples/core/Modeling/DataSeeding/ManagingDataContext.cs?name=CountrySeed)]
 
 To add entities that have a relationship the foreign key values need to be specified:
 
-[!code-csharp[PostSeed](../../../samples/core/Modeling/DataSeeding/DataSeedingContext.cs?name=PostSeed)]
+[!code-csharp[CitySeed](../../../samples/core/Modeling/DataSeeding/ManagingDataContext.cs?name=CitySeed)]
 
-If the entity type has any properties in shadow state an anonymous class can be used to provide the values:
+When managing data for Many-to-Many navigations, the join entity needs to be defined explicitly. If the entity type has any properties in shadow state (e.g. the `LanguageCountry` join entity below), an anonymous class can be used to provide the values:
```
AndriySvyryd, line 34:
```suggestion
When managing data for many-to-many navigations, the join entity needs to be configured explicitly. If the entity type has any properties in shadow state (e.g. the `LanguageCountry` join entity below), an anonymous class can be used to provide the values:
```

entity-framework/core/modeling/data-seeding.md
```diff
@@ -86,3 +87,14 @@ Depending on the constraints of your deployment the initialization code can be e
 * Deploying the initialization app with the main app, invoking the initialization routine and disabling or removing the initialization app.
 
 This can usually be automated by using [publish profiles](/aspnet/core/host-and-deploy/visual-studio-publish-profiles).
+
+<a name="use-seeding-method"></a>
+
+### Configuration options `UseSeeding` and `UseAsyncSeeding` methods
+
+EF 9 introduced `UseSeeding` and `UseAsyncSeeding` methods, which provide convenient way of configuring custom initialization logic. These methods can be set up in the [options configuration step](/ef/core/dbcontext-configuration/#dbcontextoptions). Here is an example analogous to the custom initialization showed above:
```
AndriySvyryd, line 95: Also mention that one of these methods is called when `Migrate` is called or when `dotnet ef database update` is run, even if no migrations were applied

entity-framework/core/modeling/data-seeding.md
```diff
@@ -5,38 +5,74 @@ author: AndriySvyryd
...
 Data seeding is the process of populating a database with an initial set of data.
 
 There are several ways this can be accomplished in EF Core:
 
-* Model seed data
-* Manual migration customization
+* Configuration options data seeding (`UseSeeding`)
 * Custom initialization logic
+* Model managed data (`HasData`)
+* Manual migration customization
```
AndriySvyryd, line 18: Add links pointing to the anchors below

entity-framework/core/modeling/data-seeding.md
```diff
@@ -5,38 +5,74 @@ author: AndriySvyryd
...
+
+<a name="use-seeding-method"></a>
+
+## Configuration options `UseSeeding` and `UseAsyncSeeding` methods
+
+EF 9 introduced `UseSeeding` and `UseAsyncSeeding` methods, which provide a convenient way of seeding the database with initial data. These methods aim to improve the experience of using custom initialization logic (explained below). They provide one clear location where all the data seeding code can be placed. Moreover, the code inside `UseSeeding` and `UseAsyncSeeding` methods is protected by the [migration locking mechanism](/ef/core/what-is-new/ef-core-9.0/whatsnew#concurrent-migrations) to prevent concurrency issues. The new seeding methods are called as part of `EnsureCreated` operation, `DbContext.Database.Migrate()` and `dotnet ef database update` command, even if there are no model changes and no migrations were applied.
```
AndriySvyryd, line 24:
```suggestion
EF 9 introduced `UseSeeding` and `UseAsyncSeeding` methods, which provide a convenient way of seeding the database with initial data. These methods aim to improve the experience of using custom initialization logic (explained below). They provide one clear location where all the data seeding code can be placed. Moreover, the code inside `UseSeeding` and `UseAsyncSeeding` methods is protected by the [migration locking mechanism](/ef/core/what-is-new/ef-core-9.0/whatsnew#concurrent-migrations) to prevent concurrency issues and are now the recommended way to add initial data in your app.

The new seeding methods are called as part of `<xref:Microsoft.EntityFrameworkCore.Storage.IDatabaseCreator.EnsureCreated>` operation, `<xref:Microsoft.EntityFrameworkCore.RelationalDatabaseFacadeExtensions.Migrate>` and `dotnet ef database update` command, even if there are no model changes and no migrations were applied.
```

entity-framework/core/modeling/data-seeding.md
```diff
@@ -45,18 +81,21 @@ Once the data has been added to the model, [migrations](xref:core/managing-schem
...
+> Populating the database using the `HasData` method used to be referred to as "data seeding". This naming sets incorrect expectations, as the feature has a number of limitations and is only appropriate for specific types of data. That is why we decided to rename it to "model managed data". `UseSeeding` and `UseAsyncSeeding` methods should be used for general purpose data seeding.
 
-This type of seed data is managed by migrations and the script to update the data that's already in the database needs to be generated without connecting to the database. This imposes some restrictions:
+### Limitations of model managed data
+
+This type of data is managed by migrations and the script to update the data that's already in the database needs to be generated without connecting to the database. This imposes some restrictions:
 
 * The primary key value needs to be specified even if it's usually generated by the database. It will be used to detect data changes between migrations.
-* Previously seeded data will be removed if the primary key is changed in any way.
+* Previously managed data will be removed if the primary key is changed in any way.
```
AndriySvyryd, line 94:
```suggestion
* Previously inserted managed data will be removed if the primary key is changed in any way.
```

entity-framework/core/modeling/data-seeding.md
```diff
@@ -5,38 +5,74 @@ author: AndriySvyryd
 ms.date: 11/02/2018
 uid: core/modeling/data-seeding
 ---
+
```
AndriySvyryd, line 8: Update date

entity-framework/core/modeling/data-seeding.md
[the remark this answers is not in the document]
AndriySvyryd: I still don't see a sentence saying that the feature name was changed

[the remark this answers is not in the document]
maumar: > [!NOTE]
> Populating the database using the `HasData` method used to be referred to as "data seeding". This naming sets incorrect expectations, as the feature has a number of limitations and is only appropriate for specific types of data. That is why we decided to rename it to "model managed data". `UseSeeding` and `UseAsyncSeeding` methods should be used for general purpose data seeding.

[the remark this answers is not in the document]
maumar: added it in the section for HasData rather than on top

PR #4850, opened 2024-10-24
SFI: EFDocs: Correct all image issues all severity levels

wadepickett:
Fixes issue dotnet/aspnetcore.docs#33756

Fixes all SFI issues for all severity levels.   There were 7 image files total requiring a change.

PR #4853, opened 2024-10-28
Revert "SFI: EFDocs: Correct all image issues all severity levels"

Rick-Anderson:
Reverts dotnet/EntityFramework.Docs#4850

Why were these deleted?
cc @roji @AndriySvyryd


tdykstra (approved):
Approving the revert PR so we can take another look at what needs to be changed in these images.

PR #4859, opened 2024-11-01
DO NOT MERGE, TEST PR

Rick-Anderson:

PR #4860, opened 2024-11-01
Don't build includes

Rick-Anderson:
After this merges I'll test and  revert #4856
Do the same test as #4859


tdykstra (approved):

tdykstra:
How is a build setting a better solution for markdownlint warnings than markdownlint settings?

Rick-Anderson:
AFAIK, the build setting applies to all docs. We want to retain that warning on non include docs

PR #4862, opened 2024-11-02
Update whatsnew.md

okyrylchuk:
I believe there is a mistake in the "What's New in EF Core 9" docs.

In the "[Force or prevent query parameterization](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-9.0/whatsnew#force-or-prevent-query-parameterization)" there is an information that EF Core 8.0.2 introduces an `EF.Constant` method. 

I guess it should be EF Core 9.0.2. 

If it's not a mistake, please abandon the PR.


okyrylchuk:
Ok, I can see that's not a mistake, sorry about conduction. I am closing the PR.

PR #4866, opened 2024-11-06
update examples to be more clear/correct

### LINKED ISSUE #4846: Code example calls a method with lambdas as parameters, where the lambda parameter names are the opposite of what the method labels them.
### Type of issue

Other (describe below)

### Description

In the docs about how to configure many-to-many relationships with a class for a join entity:

https://learn.microsoft.com/en-us/ef/core/modeling/relationships/many-to-many#many-to-many-with-class-for-join-entity

The code sample looks like this:

```csharp
protected override void OnModelCreating(ModelBuilder modelBuilder)
{
    modelBuilder.Entity<Post>()
        .HasMany(e => e.Tags)
        .WithMany(e => e.Posts)
        .UsingEntity<PostTag>(
            l => l.HasOne<Tag>().WithMany().HasForeignKey(e => e.TagId),
            r => r.HasOne<Post>().WithMany().HasForeignKey(e => e.PostId));
}
```

The call to `UsingEntity<PostTag>()` has those 2 lambdas, the first one with a parameter of `l` and the 2nd with a parameter named `r`, suggesting "left" and "right".

The problem is that the `UsingEntity` method signature has those reversed, where "right" is first, then "left". Like this:  `UsingEntity(Func<...> configureRight, Func<...> configureLeft)`  The sample suggests the parameter order is (left, right) when in actuality it is (right, left).

Yes, type checking should prevent the user from doing something invalid, but it may confuse developer as to what is considered "left" and what is considered "right".

### Page URL

https://learn.microsoft.com/en-us/ef/core/modeling/relationships/many-to-many#many-to-many-with-class-for-join-entity

### Content source URL

https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/modeling/relationships/many-to-many.md

9234f104-802f-8ee0-8a54-ff3b8f49857e

### Article author

@ajcvickers

timdeschryver:
Update code snippets to match lambda expressions.
I went ahead with this change because the inline snippets are already updated, e.g. https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/modeling/relationships/many-to-many.md?plain=1#L152-L162

Closes #4846

### DIFF
--- a/samples/core/Modeling/Relationships/ManyToMany.cs
+++ b/samples/core/Modeling/Relationships/ManyToMany.cs
@@ -76,8 +76,8 @@ public class ManyToMany
                     .WithMany(e => e.Posts)
                     .UsingEntity(
                         "PostTag",
-                        l => l.HasOne(typeof(Tag)).WithMany().HasForeignKey("TagsId").HasPrincipalKey(nameof(Tag.Id)),
-                        r => r.HasOne(typeof(Post)).WithMany().HasForeignKey("PostsId").HasPrincipalKey(nameof(Post.Id)),
+                        r => r.HasOne(typeof(Tag)).WithMany().HasForeignKey("TagsId").HasPrincipalKey(nameof(Tag.Id)),
+                        l => l.HasOne(typeof(Post)).WithMany().HasForeignKey("PostsId").HasPrincipalKey(nameof(Post.Id)),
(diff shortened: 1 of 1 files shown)


AndriySvyryd (approved):

AndriySvyryd:
Thanks for your contribution!

PR #4869, opened 2024-11-11
correct TimeOnly member names to not be plural

IanKemp:
also reorder DateOnly static members to appear before instance members - in line with DateTime and DateTimeOffset


roji (approved):
Thanks!

PR #4873, opened 2024-11-12
update squash docs to include community option

pdevito3:
As [requested in the EF squashing issue](https://github.com/dotnet/efcore/issues/2174#issuecomment-2469707171), I enriched the docs a bit to include a community option I made for squashing migrations


ErikEJ:
I actually meant that you could add this to the list of tools page

pdevito3:
Not sure if I know what tool reference you're talking about. Can you link to it? 

Also, would people looking at this squashing section know to look there (I wouldn't)

roji:
@pdevito3 we generally don't embed references to 3rd-party tools directly in docs, that would raise various questions on why one tool is preferred over another, what sort of responsibility Microsoft has when pointing to it, etc. etc. So yeah, it's better to simply list it in [the extensions page](https://learn.microsoft.com/en-us/ef/core/extensions/).

PR #4880, opened 2024-11-15
Remove reference to .NET 9 'Preview' in the EF9 targeting sentence as…

davepcallan:
… .NET9 is now GA

Remove reference to .NET 9 'Preview' in the EF9 targeting sentence as .NET9 is now GA


roji (approved):
Thanks. We need to do a systematic pass over our docs following the 9 GA.

PR #4882, opened 2024-11-16
Bump some supported versions

bricelam:
The current Bricelam.EntityFrameworkCore.Pluralizer and EntityFrameworkCore.ConfigurationManager binaries continue to work in EF9


roji (approved):
Thanks @bricelam!

Issue #4837, opened 2024-10-16
Update references to the new API in EF 9 with links to the API documentation


AndriySvyryd:
Depends on the API documentation being published first

roji:
Add to our release checklist (post-release steps)? Also, we need to update the link to the API docs from the conceptual TOC as well.

SamMonoRT:
This can be closed right? @AndriySvyryd

AndriySvyryd:
No. Someone still needs to go through and add links to the API docs from:

- https://learn.microsoft.com/ef/core/what-is-new/ef-core-9.0/whatsnew
- https://learn.microsoft.com/ef/core/what-is-new/ef-core-9.0/breaking-changes
- https://learn.microsoft.com/ef/core/providers/cosmos/querying
- https://learn.microsoft.com/ef/core/modeling/data-seeding

PR #4885, opened 2024-11-21
Point to latest API reference

bricelam:


roji (approved):
Thanks @bricelam!

@AndriySvyryd @ajcvickers @sampatel we should add this (and similar actions) to our release checklist...

AndriySvyryd:
@roji It's already [there](https://github.com/dotnet/efcore/issues/32915) and also tracked by https://github.com/dotnet/EntityFramework.Docs/issues/4837

PR #4886, opened 2024-11-22
Fix links, versions and some texts for SQLite docs.

cincuranet:


entity-framework/core/providers/sqlite/index.md
```diff
@@ -29,7 +29,7 @@ Install-Package Microsoft.EntityFrameworkCore.Sqlite
 
 ## Supported Database Engines
 
-* SQLite (3.7 onwards)
+* SQLite (3.46.1 onwards)
```
roji, line 32: Just curious - how do we decide/test which SQLite version is supported by our EF providers?

entity-framework/core/providers/sqlite/limitations.md
```diff
@@ -7,11 +7,11 @@ uid: core/providers/sqlite/limitations
 ---
 # SQLite EF Core Database Provider Limitations
 
-The SQLite provider has a number of migrations limitations. Most of these limitations are a result of limitations in the underlying SQLite database engine and are not specific to EF.
+The SQLite provider has a number of migrations limitations. Most of these limitations are a result of limitations in the underlying SQLite database engine and are not specific to EF Core.
```
roji, line 10: FYI we do routinely refer to EF Core as "EF" in our docs, for brevity. IIRC we sometimes say "EF Core" or even "Entity Framework Core" the first time in a page, and then just use "EF throughout".

entity-framework/core/providers/sqlite/limitations.md
```diff
@@ -26,9 +26,9 @@ SQLite doesn't natively support the following data types. EF Core can read and w
 * TimeSpan
 * UInt64
 
-Instead of `DateTimeOffset`, we recommend using DateTime values. When handling multiple time zones, we recommend converting the values to UTC before saving and then converting back to the appropriate time zone.
+Instead of `DateTimeOffset`, we recommend using `DateTime` values. When handling multiple time zones, we recommend converting the values to UTC before saving and then converting back to the appropriate time zone.
 
-The `Decimal` type provides a high level of precision. If you don't need that level of precision, however, we recommend using double instead. You can use a [value converter](xref:core/modeling/value-conversions) to continue using decimal in your classes.
+The `Decimal` type provides a high level of precision. If you don't need that level of precision, however, we recommend using `Double` instead. You can use a [value converter](xref:core/modeling/value-conversions) to continue using `Decimal` in your classes.
```
roji, line 31: nit: I think the built-in lower-case names are a bit better, that's how they're usually referred to in .NET (but obviously OK either way).

```suggestion
The `decimal` type provides a high level of precision. If you don't need that level of precision, however, we recommend using `double` instead. You can use a [value converter](xref:core/modeling/value-conversions) to continue using `decimal` in your classes.
```

entity-framework/core/providers/sqlite/limitations.md
```diff
@@ -70,7 +70,7 @@ Delete               | ✔
 
 ### Migrations limitations workaround
 
-You can workaround some of these limitations by manually writing code in your migrations to perform a rebuild. Table rebuilds involve creating a new table, copying data to the new table, dropping the old table, renaming the new table. You will need to use the `Sql(string)` method to perform some of these steps.
+You can workaround some of these limitations by manually writing code in your migrations to perform a rebuild. Table rebuilds involve creating a new table, copying data to the new table, dropping the old table, renaming the new table. You will need to use the [`MigrationBuilder.Sql`](/dotnet/api/microsoft.entityframeworkcore.migrations.migrationbuilder.sql) method to perform some of these steps.
```
roji, line 73: Use `<xref>` to refer to our API documentation (and to other pages in the conceptual docs).

entity-framework/core/providers/sqlite/limitations.md
```diff
@@ -92,9 +92,9 @@ dotnet ef database update --connection "Data Source=My.db"
...
+EF Core 9 introduced a locking mechanism when executing migrations. It aims to protect against multiple migration executions happening simultaneously, as that could leave the database in a corrupted state. This is one of the potential problems resulting from applying migrations at runtime using the [`DbContext.Database.Migrate()`](/dotnet/api/microsoft.entityframeworkcore.relationaldatabasefacadeextensions.migrate) method (see [Applying migrations](xref:core/managing-schemas/migrations/applying) for more information). To mitigate this, EF Core creates an exclusive lock on the database before any migration operations are applied.
```
roji, line 95: Again, just writing EF9 is standard across docs, and I'd personally prefer that... There's no actual ambiguity here (these are EF Core docs, there's no non-core EF 9), and I don't remember any user ever complaining or getting confused...

entity-framework/toc.yml
```diff
@@ -423,7 +423,7 @@
         - name: Spatial data
           displayName: GIS
           href: core/providers/sqlite/spatial.md
-        - name: Microsoft.Data.Sqlite >>
+        - name: Microsoft.Data.Sqlite
```
roji, line 426: I think the `>>` is an indication that this directs users to a page outside the EF conceptual documentation, ilke other links below - do you think there's a reason to remove that? If so, I think we'd do it everywhere, not just for this one.

roji (changes requested):

entity-framework/core/providers/sqlite/index.md
cincuranet: I don't think we "decide". Before .NET 9 release we took what was latest and referenced the underlying packages. It will probably work with older versions (minus the fixed bugs, etc. like i.e. #34590), but we don't explicitly test this.

roji (approved):

PR #4898, opened 2024-12-06
Update advanced-performance-topics.md

azarboon:
Connection pooling and relationship between EF and ADO isn't explained well in the documentation (AFAIK). I've added it to clarify it. Given the authority of the doc, it should clearly explain it.


entity-framework/core/performance/advanced-performance-topics.md
```diff
@@ -77,6 +77,12 @@ The full source code for this sample is available [here](https://github.com/dotn
 > [!NOTE]
 > Although EF Core takes care of resetting internal state for `DbContext` and its related services, it generally does not reset state in the underlying database driver, which is outside of EF. For example, if you manually open and use a `DbConnection` or otherwise manipulate ADO.NET state, it's up to you to restore that state before returning the context instance to the pool, e.g. by closing the connection. Failure to do so may cause state to get leaked across unrelated requests.
 
+### Connection Pooling Considerations
+
+Connection pooling in Entity Framework is managed at the client level through ADO.NET and works consistently across all SQL Server-based environments, including Azure SQL Database, Azure SQL Managed Instance, and SQL Server (on-premises or in virtual machines). While pooling mechanisms are consistent, service-specific factors may impact pooling efficiency. Entity Framework relies on ADO.NET for database interactions, including connection pooling, which reuses existing database connections to minimize the overhead of repeatedly opening and closing connections.
```
roji, line 82: I agree that it makes sense to discuss connection pooling and its relationship (or rather non-relationship) to EF. Though I'd avoid saying "connection pooling in EF" (since there is no connection pooling at EF, as you do explain). I think it's also inappropriate to focus on SQL Server here specifically, as the information is generally valid to all databases.

I'll take a look a bit later and probably tweak the wording a bit, but otherwise the gist is good - thanks for submitting this!

azarboon:
@roji thanks for your feedback. I've tweaked it accordingly. Please let me know your advice.

roji:
Thanks @azarboon. I've pushed a commit proposing some tweaks; we need to also keep in mind that EF supports non-relational providers as well, where ADO.NET doesn't play any role (and in fact, no pooling is done or needed, since there's no persistent connection).

Let me know if this looks OK to you.

azarboon:
> Thanks @azarboon. I've pushed a commit proposing some tweaks; we need to also keep in mind that EF supports non-relational providers as well, where ADO.NET doesn't play any role (and in fact, no pooling is done or needed, since there's no persistent connection).
> 
> Let me know if this looks OK to you.

Excellent. Looks great. Please sign it off

roji (approved):

PR #4904, opened 2024-12-10
Reorg providers by paid vs. free/OSS

mguinness:
Separate open source and commercial providers to reduce clutter


mguinness:
@dotnet-policy-service agree

roji:
I'm not sure the current list is "cluttered", it just doesn't separate free/OSS from commercial providers... Any particular reason you think this should be done?

roji:
I'll discuss this with the team and we'll see what people think.

cincuranet:
If we decide to merge, I think the "Open Source" and "Commercial" headings need change. These two are not mutually exclusive.

roji:
@cincuranet yeah, I remember the same thing - not splitting the table (we should be more disciplined about documenting our decisions).

PR #4912, opened 2024-12-30
Update index.md

on3oleg:
Support of EF Core 9 by dotConnect's and added new products dotConnect for Zoho Book, dotConnect for Zoho Desk.


roji (approved):
Thanks.

PR #4916, opened 2024-12-31
Update single-split-queries.md

NargesMousavi:
demonstrate single-split-queries by sibling joins to better understand Cartesian explosion issues


NargesMousavi:
@dotnet-policy-service agree [company=""]

NargesMousavi:
@dotnet-policy-service agree

NargesMousavi:
@dotnet-policy-service agree company="Microsoft"

roji (commented):
I understand the idea of showing a code sample with sibling joins, which is typically a particularly good case for split query, but the idea in these docs is to keep things as simple as possible, to make sure the user understands what split query actually do; understanding the principle doesn't require having two joins, which makes the sample more complicated IMHO.

roji:
I'll go ahead and close this for now as no response was provided and I'm not convinced this is a good idea (see above). But am open to discussing.

PR #4920, opened 2025-01-03
Fix sample code in SQL querying page

### LINKED ISSUE #4919: SqlParameter doesn't work for FromSqlRaw
### Type of issue

Code doesn't work

### Description

The sample code for `FromSqlRaw` doesn't work.

```C#
var columnName = "Url";
var columnValue = new SqlParameter("columnValue", "http://SomeURL");

var blogs = context.Blogs
    .FromSqlRaw($"SELECT * FROM [Blogs] WHERE {columnName} = @columnValue", columnValue)
    .ToList();
```

It produces a runtime error

```
System.InvalidCastException: Unable to cast object of type 'Microsoft.Data.SqlClient.SqlParameter' to type 'Microsoft.Data.Sqlite.SqliteParameter'.
```

Here the type `SqlParameter` is incorrect. It should be a provider specific type, say `SqliteParameter` for SQLite, `NpgsqlParameter` for Npgsql.

### Page URL

https://learn.microsoft.com/en-us/ef/core/querying/sql-queries?tabs=sqlserver

### Content source URL

https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/querying/sql-queries.md

98b01ef5-24a5-5f59-30c9-af264f739815

### Article author

@smitpatel

roji:
Our conceptual docs generally show SQL Server code, but the samples are designed to work against Sqlite (so they can run out of the box). Inlined sample code into the SQL querying page, which seems to be our general direction anyway.

Fixes #4919

### DIFF
--- a/samples/core/Querying/SqlQueries/Program.cs
+++ b/samples/core/Querying/SqlQueries/Program.cs
@@ -1,5 +1,6 @@
 ﻿using System.Linq;
 using Microsoft.Data.SqlClient;
+using Microsoft.Data.Sqlite;
 using Microsoft.EntityFrameworkCore;
 
 namespace EFQuerying.RawSQL;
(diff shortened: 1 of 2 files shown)


AndriySvyryd:
> Inlined sample code into the SQL querying page, which seems to be our general direction anyway.

The general direction is to keep them in projects, so that they can be run to ensure that they are still working.

samples/core/Querying/SqlQueries/Program.cs
```diff
@@ -49,128 +50,104 @@ from [Post] as [p]
 
         using (var context = new BloggingContext())
         {
-            #region FromSql
```
AndriySvyryd, line 52: Remove this file if it's no longer being referenced.

AndriySvyryd (approved):

samples/core/Querying/SqlQueries/Program.cs
roji: #4931

PR #4926, opened 2025-01-08
Fix wording

ErikEJ:
@brentozar FYI

https://www.brentozar.com/archive/2025/01/i-feel-sorry-for-untrained-developers-using-entity-framework/


ErikEJ:
@roji SqlClient maybe, but not "SQL Server" - dont think it does any "mapping"

roji:
Yep.

PR #4948, opened 2025-02-10
Use consistent migration name in example

stephannielsen:
Aligns the file names in the examples to be consistent (include `Blog` in the file names).


roji (approved):
Thanks!

PR #4972, opened 2025-03-08
Update concurrency.md

IvanStoychev:
Add information about circumstances under which it is possible for a `DbUpdateConcurrencyException` to be thrown when adding entities.

In the edited document it is originally stated that a `DbUpdateConcurrencyException` cannot be thrown when adding new entities. However in an SQLite table with constraint `UNIQUE("Column Name") ON CONFLICT IGNORE` attempting to add a conflicting entity will cause SQLite to simply ignore it and, thus, return fewer rows affected than what EF Core would expect, causing it to raise the exception.

The same is probably true for other RDBMS, for example: Microsoft SQL Server with `IGNORE_DUP_KEY = ON` but I have not tested it to confirm.


roji:
While what you say is technically correct, the situations in which this happens are very, very edge-casey and users have to very intentionally and specifically configure their database in a special way to make this happen. So I'd rather simply add the word `generally` just before `never` and be done with it, without going into reasons - docs need to strike a careful balance providing correct information and not overloading readers with information that's irrelevant to 99% of our users.

Does that make sense?

IvanStoychev:
I understand what you mean and it does make sense. I think we're both right - I can see your reasoning but still believe giving this information to users would be beneficial.

In the ends these are your docs, not mine, so if you want me to just add `generally` - I'll do so.

roji:
OK, thanks. At the end of the day, this is the first time anyone has raised this AFAIK, and I don't believe the the added sentences would actually help someone in advance in some concrete case. So yeah, I believe just adding `generally` is sufficient - if we see people struggling with this in the future we can always add more info though.

roji (approved):
Thanks.

PR #4979, opened 2025-03-13
EF Scaffolding many to many join Entity

lgonzalezr:
There isn't any parameter in the scaffolding command to include this type of Entity it is disabled by Default. This is required in cases where fine control of the generated SQL, keeping LINQ queries more TSQL like instead of C# like and others.

### DIFF
--- a/entity-framework/core/managing-schemas/scaffolding/templates.md
+++ b/entity-framework/core/managing-schemas/scaffolding/templates.md
@@ -180,7 +180,7 @@ To scaffold these classes, you can use a third template called `EntityTypeConfig
 
 ### Generate Join Table in Many to Many Relationships
 
-By default, the scaffolding process does not generate an entity for join tables in simple many-to-many relationships. However, there are cases where explicitly generating the join table as an entity might be necessary (e.g., when finer control over the generated TSQL query is required).
+By default, the scaffolding process does not generate an entity for join tables in simple many-to-many relationships. However, there are cases where explicitly generating the join table as an entity might be necessary (e.g., when finer control over the generated SQL query is required).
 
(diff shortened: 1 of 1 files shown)


entity-framework/core/managing-schemas/scaffolding/templates.md
```diff
@@ -144,6 +144,24 @@ If you did everything correctly, the collection navigation properties should now
 public virtual ICollection<Album> Albums { get; } = new ObservableCollection<Album>();
 ```
 
+## Generate Joint Table in Many to Many Relationships
```
AndriySvyryd, line 147:
```suggestion
### Generate Join Table in Many to Many Relationships
```

AndriySvyryd: Also move this section below `### Entity configuration classes`

lgonzalezr: Done

entity-framework/core/managing-schemas/scaffolding/templates.md
```diff
@@ -144,6 +144,24 @@ If you did everything correctly, the collection navigation properties should now
 public virtual ICollection<Album> Albums { get; } = new ObservableCollection<Album>();
 ```
 
+### Generate Join Table in Many to Many Relationships
+
+By default, the scaffolding process does not generate an entity for join tables in simple many-to-many relationships. However, there are cases where explicitly generating the join table as an entity might be necessary (e.g., when finer control over the generated TSQL query is required).
```
cincuranet, line 149: TSQL -> SQL
T-SQL is specific to SQL Server.

cincuranet (changes requested):

entity-framework/core/managing-schemas/scaffolding/templates.md
lgonzalezr: Done

cincuranet:
Thanks.

PR #4983, opened 2025-03-18
fixing dotnet core release notes link

maumar:


maumar:
@jamesmontemagno can you confirm this is the correct format for release notes link that we should use in docs

AndriySvyryd (approved):

PR #4986, opened 2025-03-20
Update standups

cincuranet:


roji (approved):
@cincuranet I personally think it should be OK for you to merge such PRs without approval (like how we e.g. merge typo PRs without approval).

PR #4998, opened 2025-04-04
Explain "payload".

DejanPopovic1:
The word payload is introduced without a corresponding definition, leaving the reader a bit confused. This provides a succinct definition when first encountering the term  in the article.


cincuranet (approved):
Thanks.

PR #5009, opened 2025-04-18
Cosmos full-text search "what's new" docs

maumar:
What's new docs for https://github.com/dotnet/efcore/issues/35476

### DIFF
--- a/entity-framework/core/providers/cosmos/full-text-search.md
+++ b/entity-framework/core/providers/cosmos/full-text-search.md
@@ -1,0 +1,131 @@
+---
+title: Full Text Search - Azure Cosmos DB Provider - EF Core
+description: Full text search with the Azure Cosmos DB EF Core Provider
+author: maumar
+ms.date: 04/19/2025
+uid: core/providers/cosmos/full-text-search
+---
+# Full text search
+
+Azure Cosmos DB now offers support for [full-text search](/azure/cosmos-db/gen-ai/full-text-search). It enables efficient and effective text searches using advanced techniques like stemming, as well as evaluating the relevance of documents to a given search query. It can be used in combination with vector search (i.e. hybrid search) to improve the accuracy of responses in some AI scenarios.
+EF Core allows for modeling the database with full-text search enabled properties and using full-text search functions inside queries targeting Azure Cosmos DB.
+
+## Model configuration
+
+A property can be configured inside `OnModelCreating` to use full-text search by enabling it for the property and defining a full-text index:
+
+```c#
+public class Blog
+{
+    ...
+
+    public string Contents { get; set; }
+}
+
+public class BloggingContext
+{
+    ...
+
+    protected override void OnModelCreating(ModelBuilder modelBuilder)
+    {
+        modelBuilder.Entity<Blog>(b =>
+        {
+            b.Property(x => x.Contents).EnableFullTextSearch();
+            b.HasIndex(x => x.Contents).IsFullTextIndex();
+        });
+    }
+}
+```
+
+> [!NOTE]
+> Configuring the index is not mandatory, but it is recommended as it greatly improves performance of full-text search queries.
+
+Full-text search operations are language specific, using American English (`en-US`) by default. You can customize the language for individual properties as part of `EnableFullTextSearch` call:
+
+```c#
+    protected override void OnModelCreating(ModelBuilder modelBuilder)
+    {
+        modelBuilder.Entity<Blog>(b =>
+        {
+            b.Property(x => x.Contents).EnableFullTextSearch();
+            b.HasIndex(x => x.Contents).IsFullTextIndex();
+            b.Property(x => x.ContentsGerman).EnableFullTextSearch("de-DE");
+            b.HasIndex(x => x.ContentsGerman).IsFullTextIndex();
+        });
+    }
+```
+
+You can also set a default language for the container - unless overridden in the `EnableFullTextSearch` method, all full-text properties inside the container will use that language.
+
+```c#
+    protected override void OnModelCreating(ModelBuilder modelBuilder)
+    {
+        modelBuilder.Entity<Blog>(b =>
+        {
+            b.HasDefaultFullTextLanguage("de-DE");
+            b.Property(x => x.ContentsEnglish).EnableFullTextSearch("en-US");
+            b.HasIndex(x => x.ContentsEnglish).IsFullTextIndex();
+            b.Property(x => x.ContentsGerman).EnableFullTextSearch();
+            b.HasIndex(x => x.ContentsGerman).IsFullTextIndex();
+            b.Property(x => x.TagsGerman).EnableFullTextSearch();
+            b.HasIndex(x => x.TagsGerman).IsFullTextIndex();
+        });
+    }
+```
+
+## Querying
+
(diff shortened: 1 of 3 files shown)


maumar:
@SamMonoRT https://github.com/dotnet/EntityFramework.Docs/pull/5010

entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
```diff
@@ -21,6 +21,158 @@ EF10 requires the .NET 10 SDK to build and requires the .NET 10 runtime to run.
 
 ## Azure Cosmos DB for NoSQL
 
+<a name="full-text-search-support"></a>
+
+### Full-text search support
+
+Azure Cosmos DB now offers support for [full-text search](/azure/cosmos-db/gen-ai/full-text-search). It enables efficient and effective text searches, as well as evaluating the relevance of documents to a given search query. It can be used in combination with vector search to improve the accuracy of responses in some AI scenarios.
```
roji, line 28: I think the main point of FTS is about handling natural language well (e.g. stemming), may be worth mentioning that. Re the combination with vector search, I'd explicitly mention that as "hybrid" search, which is the standard term.

entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
```diff
@@ -21,6 +21,158 @@ EF10 requires the .NET 10 SDK to build and requires the .NET 10 runtime to run.
...
+            b.HasIndex(x => x.TagsGerman).IsFullTextIndex();
+        });
+    }
+```
+
+<a name="full-text-search-querying"></a>
+
+#### Querying
+
+As part of the full-text search feature, Azure Cosmos DB introduced several built-in functions which allow for efficient querying of content inside the full-text search enabled properties. These functions are: `FullTextContains`, `FullTextContainsAll`, `FullTextContainsAny`, which look for specific keyword or keywords and `FullTextScore`, which returns [BM25 score](https://en.wikipedia.org/wiki/Okapi_BM25) based on provided keywords.
```
roji, line 100: Link to the Cosmos doc pages for these functions?

entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
```diff
@@ -21,6 +21,158 @@ EF10 requires the .NET 10 SDK to build and requires the .NET 10 runtime to run.
...
+> [!TIP]
+> You can combine more than two scoring functions inside `Rrf` call, as well as using only `FullTextScore`, or only `VectorDistance`.
+
+<a name="full-text-search-limitations"></a>
+
+#### Limitations
+
+Here are some current limitations of the full-text search using EF Core:
+
+- `FullTextScore` and `RRF` functions can only be used inside `OrderBy` and are mutually exclusive with other forms of ordering.
```
roji, line 163: Are these "current limitations" in EF? Isn't this just how Cosmos works (and not a limitation per se)?

entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
```diff
@@ -21,6 +21,158 @@ EF10 requires the .NET 10 SDK to build and requires the .NET 10 runtime to run.
...
+```
+
+> [!TIP]
+> You can combine more than two scoring functions inside `Rrf` call, as well as using only `FullTextScore`, or only `VectorDistance`.
+
+<a name="full-text-search-limitations"></a>
+
+#### Limitations
+
+Here are some current limitations of the full-text search using EF Core:
```
roji, line 161: None of the below seem like EF Core limitations, but rather Cosmos ones. BTW I wouldn't actually call these limitations, since that implies the feature is incomplete and these limitations will be removed in a future versions. I don't think this is true for these - that's just the (sensible) design of the feature.

On an unrelated note, I'd not put this here in "what's new" - that goes too deeply into detail and generally makes the "what's new" huge (we should generally strive to make it minimal, showing the feature in a relatively minimal but awesome way, and then linking to the full docs). So I think these belong in the docs section under the Cosmos provider documentation.

roji (approved):
Thanks, looks pretty good. Note that we also need to document this in the Cosmos docs, not just in "what's new", I'm a believer in doing both at the same time - this way the what's new can be relatively minimal, and link to the full docs.

entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
maumar: RRF in OrderBy is Cosmos, not supporting FTS on collections is EF - you can have FTS property anywhere in the document without issue, but wildcards are not supported (when defining paths for index), so people need to define a path for each element in the collection `/Collection/0/MyFts`, `/Collection/1/MyFts` and so on. This is unrealistic for us to do on EF, so I disabled the feature (so that people don't fall into the pit of using FTS without index)

entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
maumar: I mentioned hybrid search in the section below, but makes sense to do it here also

entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
maumar: Yeah, limitations doesn't sound great. Will convert these to NOTEs or something

entity-framework/core/providers/cosmos/full-text-search.md
```diff
@@ -0,0 +1,134 @@
...
+            b.Property(x => x.ContentsGerman).EnableFullTextSearch();
+            b.HasIndex(x => x.ContentsGerman).IsFullTextIndex();
+            b.Property(x => x.TagsGerman).EnableFullTextSearch();
+            b.HasIndex(x => x.TagsGerman).IsFullTextIndex();
+        });
+    }
+```
+
+> [!NOTE]
+> Configuring the index is not mandatory, but it is recommended as it greatly improves performance of full-text search queries.
```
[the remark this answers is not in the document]
maumar: oops, AI experiments :)

maumar:
updated @roji

entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
```diff
@@ -21,6 +21,68 @@ EF10 requires the .NET 10 SDK to build and requires the .NET 10 runtime to run.
...
+```
+
+Once the model is configured, we can use full-text search operations in queries using methods provided in `EF.Functions`:
+
+```c#
+var cosmosBlogs = await context.Blogs.Where(x => EF.Functions.FullTextContains(x.Contents, "cosmos")).ToListAsync();
+```
+
+The following full-text operations are currently supported: [`FullTextContains`](/azure/cosmos-db/nosql/query/fulltextcontains), [`FullTextContainsAll`](/azure/cosmos-db/nosql/query/fulltextcontainsall), [`FullTextContainsAny`](/azure/cosmos-db/nosql/query/fulltextcontainsany), [`FullTextScore`](/azure/cosmos-db/nosql/query/fulltextscore).
+
```
roji, line 63: Add "For more information on Cosmos full-text search, [see the docs](...)" (IMHO this is what every section in the what's new docs should look like).

roji (approved):
Thanks @maumar, looks great.

entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
maumar: It was already there, after Hybrid Search - hybrid was a subsection of FTS. I've elevated it to it's own section (since it's such an important concept for AI) and added the docs note for it and for FTS in general.

SamMonoRT:
@jamesmontemagno @roji - is this good to go given P4 was released?

richlander:
Yes. Can we merge it?

richlander:
Merging to keep the process going and to help out other teams who want to reference this content. We can adjust after the fact

PR #5019, opened 2025-04-29
add Couchbase to the list

vslee:
- Couchbase just announced GA of their provider: https://www.couchbase.com/blog/ef-core-provider-for-couchbase-general-availability/ so adding it here for them. 

- and updated Google Spanner to 8 (and no longer in preview)


roji (approved):
Thanks!

PR #5032, opened 2025-05-20
Update mapping-attributes.md

oasaleh:
`Blog` is the principal.


oasaleh:
Maybe even this could make it clearer?

`[ForeignKey]` can be applied to the foreign-key property, with its argument set to the name of the navigation property defined on the dependent entity that references the principal entity.

PR #5035, opened 2025-05-23
Add Named Default Constraints to What's New

AndriySvyryd:


roji (commented):
Update [the docs](https://learn.microsoft.com/en-us/ef/core/modeling/generated-properties?tabs=data-annotations#default-values) as well?

roji (approved):

PR #5056, opened 2025-06-23
Document MDS breaking changes from efcore#36229.

cincuranet:


entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -83,3 +86,95 @@ await context.Blogs.ExecuteUpdateAsync(s =>
...
+
+<a name="DateTimeOffset-read"></a>
+
+#### Reading DateTimeOffset without an offset now assumes UTC
+
+[Tracking Issue #36195](https://github.com/dotnet/efcore/issues/36195)
+
+##### Old behavior
+
+Previously, when reading a `DateTimeOffset` value that did not have an offset (e.g., `2014-04-15 10:47:16`), Microsoft.Data.Sqlite would assume the value was in the local time zone.
```
roji: > a `DateTimeOffset` value that did not have an offset

That seems like a bit of a contradiciton in terms... SQLite doesn't actually store a `DateTimeOffset` - if stores the textual representation of a timestamp.

Also, it's not quite clear to me what this "assumption" entails - consider documenting the user-visible effect, i.e. how the values read are actually different. Best to also give an example of the same value in the database, and what you get when you read it before and after.

entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -83,3 +86,95 @@ await context.Blogs.ExecuteUpdateAsync(s =>
...
+
+##### New behavior
+
+Starting with Microsoft.Data.Sqlite 10.0, when reading a `DateTimeOffset` value that does not have an offset, Microsoft.Data.Sqlite will assume the value is in UTC.
+
+##### Why
+
+Is is to align with SQLite's behavior where timestamps without an offset are treated as UTC.
```
roji, line 120: Also to remove any dependency on the local configured timezone on the machine where the .NET app happens to be running, no?

entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -83,3 +86,95 @@ await context.Blogs.ExecuteUpdateAsync(s =>
...
+
+##### Mitigations
+
+You can revert to previous behavior by setting `Microsoft.Data.Sqlite.Pre10TimeZoneHandling` AppContext switch to `true`, see [AppContext for library consumers](/dotnet/api/system.appcontext#ForConsumers) for more details.
```
roji: May seem obvious, but it's worth telling users that the first (and recommended) mitigation is to adjust their code accordingly; the context switch really should only be a temporary/last resort thing.

entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -17,6 +17,9 @@ This page documents API and behavior changes that have the potential to break ex
 
 ## Summary
 
+> [!NOTE]
+> If you are using Microsoft.Data.Sqlient, please see the [separate section below on Microsoft.Data.Sqlite breaking changes](#MDS-breaking-changes).
```
roji:
```suggestion
> If you are using Microsoft.Data.Sqlite, please see the [separate section below on Microsoft.Data.Sqlite breaking changes](#MDS-breaking-changes).
```

entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -83,3 +86,95 @@ await context.Blogs.ExecuteUpdateAsync(s =>
...
+
+Previously, when writing a `DateTimeOffset` value into a REAL column, Microsoft.Data.Sqlite would write the value without taking the offset into account.
+
+##### New behavior
+
+Starting with Microsoft.Data.Sqlite 10.0, when writing a `DateTimeOffset` value into a REAL column, Microsoft.Data.Sqlite will convert the value to UTC before doing the conversions and writing it.
+
+##### Why
+
+This change is to align with SQLite's behavior where timestamps without an offset are treated as UTC.
```
roji: More importantly, the previous behavior basically disregarded the offset entirely, and so disregarded the absolute moment in time which the DateTimeOffset represents.

entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -83,3 +86,95 @@ await context.Blogs.ExecuteUpdateAsync(s =>
...
+
+##### Mitigations
+
+You can revert to previous behavior by setting `Microsoft.Data.Sqlite.Pre10TimeZoneHandling` AppContext switch to `true`, see [AppContext for library consumers](/dotnet/api/system.appcontext#ForConsumers) for more details.
```
roji: As above, users should preferably adjust their code.

entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -83,3 +86,95 @@ await context.Blogs.ExecuteUpdateAsync(s =>
...
+
+<a name="DateTime-read"></a>
+
+#### Reading DateTime with an offset now returns value in UTC
+
+[Tracking Issue #36195](https://github.com/dotnet/efcore/issues/36195)
+
+##### Old behavior
+
+Previously, when reading a `DateTime` value that had an offset (e.g., `2014-04-15 10:47:16+02:00`), Microsoft.Data.Sqlite would return the value with `DateTimeKind.Local` (even if the offset was not local).
```
roji:
```suggestion
Previously, when reading a textual timestamp value that had an offset (e.g., `2014-04-15 10:47:16+02:00`) as a .NET `DateTime`, Microsoft.Data.Sqlite would convert the value to a local timestamp and return a `DateTime with  `DateTimeKind.Local`.
```

entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -83,3 +86,95 @@ await context.Blogs.ExecuteUpdateAsync(s =>
...
+
+##### New behavior
+
+Starting with Microsoft.Data.Sqlite 10.0, when reading a `DateTime` value that has an offset, Microsoft.Data.Sqlite will convert the value to UTC and return it with `DateTimeKind.Utc`.
+
+##### Why
+
+Even though the time was parsed correctly taking offset into account, the `DateTimeKind` was not set correctly.
```
roji: I think that here too, it's more a matter of removing the dependency on the local machine time zone (not sure the previous behavior can be said to be incorrect).

roji (changes requested):

entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -143,37 +145,41 @@ Starting with Microsoft.Data.Sqlite 10.0, when writing a `DateTimeOffset` value
...
 
 ##### New behavior
 
-Starting with Microsoft.Data.Sqlite 10.0, when reading a `DateTime` value that has an offset, Microsoft.Data.Sqlite will convert the value to UTC and return it with `DateTimeKind.Utc`.
+Starting with Microsoft.Data.Sqlite 10.0, when using `GetDateTime` on a textual timestamp that has an offset, Microsoft.Data.Sqlite will convert the value to UTC and return it with `DateTimeKind.Utc`.
 
 ##### Why
 
-Even though the time was parsed correctly taking offset into account, the `DateTimeKind` was not set correctly.
+Even though the time was parsed correctly it was dependent on a on local time zone, which could lead to unexpected results.
```
roji:
```suggestion
Even though the time was parsed correctly it was dependent on the machine-configured local time zone, which could lead to unexpected results.
```

PR #5093, opened 2025-09-03
Additions to SQL Server JSON breaking changes

roji:


entity-framework/core/what-is-new/ef-core-10.0/breaking-changes.md
```diff
@@ -87,16 +89,18 @@ The new JSON data type introduced by SQL Server is a superior, 1st-class way to
 
 #### Mitigations
 
-If you do not wish to transition to the new JSON data type right away, you can configure EF with a compatibility level lower than 170:
+If you are targeting Azure SQL Databasse and do not wish to transition to the new JSON data type right away, you can configure EF with a compatibility level lower than 170:
```
cincuranet: Databas**s**e -> Database

roji:
```suggestion
If you are targeting Azure SQL Database and do not wish to transition to the new JSON data type right away, you can configure EF with a compatibility level lower than 170:
```

PR #5102, opened 2025-09-14
Move stale extensions down and update a few version numbers

ErikEJ:


roji (approved):
Thanks as always @ErikEJ!

PR #5158, opened 2025-11-17
Consistency fixes in EF Core Database Provider docs

RadoslavGatev:
The same configuration approach should be followed in DI and non-DI scenarios. Therefore, for consistency, I propose using `UseAzureSql` in the Azure SQL tab and `UseAzureSynapse` for Azure Synapse Analytics respectively.


roji (approved):
Makes sense, thanks!

PR #5159, opened 2025-11-17
Fix dotnet counters monitor invocation

RadoslavGatev:
The `--counters` option must be specified in order to only monitor a specific provider like Microsoft.EntityFrameworkCore.

More info: https://learn.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-counters#dotnet-counters-monitor

Otherwise it outputs the following message:
> --counters is unspecified. Monitoring System.Runtime counters by default.


roji (approved):
Thanks!

PR #5247, opened 2026-01-17
Change BillingAddress to allow NULL values

flcdrg:
"NOT NULL NULL" is not valid T-SQL. Given it refers to the previous type, allowing this column to be nullable seems to be what was intended.

### DIFF
--- a/entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
+++ b/entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
@@ -304,7 +304,7 @@ CREATE TABLE [Customers] (
     [Id] int NOT NULL IDENTITY,
     [Name] nvarchar(max) NOT NULL,
     [ShippingAddress] json NOT NULL,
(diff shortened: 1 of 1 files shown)


entity-framework/core/what-is-new/ef-core-10.0/whatsnew.md
```diff
@@ -304,7 +304,7 @@ CREATE TABLE [Customers] (
     [Id] int NOT NULL IDENTITY,
     [Name] nvarchar(max) NOT NULL,
     [ShippingAddress] json NOT NULL,
-    [BillingAddress] json NOT NULL NULL,
+    [BillingAddress] json NULL,
```
roji:
```suggestion
    [BillingAddress] json NOT NULL,
```

flcdrg:
But it isn't!

If you refer to the most recent definition above it is defined like this:

```csharp
public class Customer
{
    ...

    public Address ShippingAddress { get; set; }
    public Address? BillingAddress { get; set; }
}
```

PR #5100, opened 2025-09-11
Add documentation for cosmos db provider transactional batches

JoasE:
Adds documentation for the usage of TransactionalBatch in the Cosmos DB provider when saving changes.
See: https://github.com/dotnet/efcore/issues/17308 and https://github.com/dotnet/efcore/pull/36693

PR #5256, opened 2026-02-02
Preview 1 docs

roji:
Adds documentation for EF 11 preview.1 (note that this targets the preview-1 branch, so will not go live when merged).

This is notably missing documentation for https://github.com/dotnet/efcore/issues/17308, as we have https://github.com/dotnet/EntityFramework.Docs/pull/5100 pending.

@JoasE can you please take a look as this covers two significant features you're contributed? A review from you would be much appreciated!

### DIFF
--- a/entity-framework/core/providers/cosmos/saving.md
+++ b/entity-framework/core/providers/cosmos/saving.md
@@ -22,7 +22,7 @@ Starting with EF 11, the EF Core Azure Cosmos DB provider leverages transactiona
 
 * **Auto** (default) – EF automatically groups operations into transactional batches, which are executed sequentially. If a batch fails, execution stops immediately and no subsequent changes are saved, while any previously successful batches remain committed. This generally provides good performance when performing multiple operations that can be grouped by partition key value, with a best-effort approximation of atomicity.
 * **Never** – All operations are performed individually and sequentially, in the exact order they were tracked. This avoids batching and can be slower, especially for large numbers of changes. This was the behavior prior to version 11.
-* **Always** – Requires that all operations can be executed as a single transaction batch; if any operation cannot be included in a batch (see [Limitations](#Limitations)), an exception is thrown. This allows you to guarantee full atomicity (and a single roundtrip) when executing `SaveChangesAsync`, but it is then up to you to manually ensure that all operations can be performed in a transactional batch.
+* **Always** – Requires that all operations can be executed as a single transaction batch; if any operation cannot be included in a batch (see [Limitations](#limitations)), an exception is thrown. This allows you to guarantee full atomicity (and a single roundtrip) when executing `SaveChangesAsync`, but it is then up to you to manually ensure that all operations can be performed in a transactional batch.
 
 Here is an example of using <xref:Microsoft.EntityFrameworkCore.AutoTransactionBehavior.Always>, which causes `SaveChangesAsync` to fail because too many operations are attempted:
 


entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+The following modes are available:
+
+Mode               | Description
+-------------------|------------
+`FullyAutomatic`   | The default mode. Uses the underlying Cosmos DB SDK automatic session token management. `GetSessionTokens` and `UseSessionTokens` methods will throw when invoked.
+`SemiAutomatic`    | Allows overwriting the SDK's automatic session token management via `UseSessionTokens`. If `UseSessionTokens` has not been called for a container, the SDK's automatic management is used. EF tracks session tokens which can be retrieved via `GetSessionTokens`.
+`Manual`           | Fully replaces SDK automatic session token management. Only session tokens specified via `UseSessionTokens` or tracked by EF operations are used.
+`EnforcedManual`   | Same as `Manual`, but throws an exception if `UseSessionTokens` was not called before executing a read operation.
```
JoasE: Currently EF also throws for write operations with EnforcedManual if UseSessionTokens ins't invoked

JoasE: And this is also container scoped, so maybe something like 'if UseSessionTokens wasn't called for the container before executing an operation in it'

entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,153 @@
...
+{
+    await using var context = new MyCosmosContext();
+
+    // Get the session token from the request header
+    if (Request.Headers.TryGetValue("x-cosmos-session-token", out var sessionToken))
+    {
+        context.Database.UseSessionToken(sessionToken!);
+    }
+
+    var document = await context.Documents.FindAsync(id);
```
JoasE, line 173: technically the document could have been updated in the meantime, and the read will update the session token. To ensure retrieving the same version again on the next read, you'd need to set a response header or cookie aswell

entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,153 @@
...
+
+    var document = new Document { Name = dto.Name };
+    context.Documents.Add(document);
+    await context.SaveChangesAsync();
+
+    // Retrieve the session token to return to the client
+    var sessionToken = context.Database.GetSessionToken();
+
+    // Store in response header for the client to send back in subsequent requests
+    Response.Headers["x-cosmos-session-token"] = sessionToken;
```
JoasE, line 157: In my experience, most commonly cookies are used to manage this, as you'd want the token to expire at some point. But headers for the example are fine ofcourse aswell

entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,153 @@
...
+Note that unlike with relational databases, Azure Cosmos DB doesn't provide transaction guarantees by default across multiple documents.
+
+## Bulk execution
+
+> [!NOTE]
+> Bulk execution is being introduced in EF Core 11, which is currently in preview.
+
+By default, EF Core executes document operations sequentially when calling `SaveChangesAsync`. When saving a large number of entities, this can be slow as each operation waits for the previous one to complete before starting.
+
+Azure Cosmos DB supports _bulk execution_, which allows multiple document operations to be executed in parallel, significantly improving throughput when saving many entities at once. This is especially useful for data loading scenarios, batch operations, or any situation where you need to save many entities.
```
JoasE: Maybe nice to mention? But doesn't have to be in there
```suggestion
Azure Cosmos DB supports _bulk execution_, which allows multiple document operations to be executed in parallel and across DbContext instances, significantly improving throughput when saving many entities at once. This is especially useful for data loading scenarios, batch operations, or any situation where you need to save many entities.
```

entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,153 @@
...
+Saving data with the Azure Cosmos DB provider works in a similar fashion to other providers. You add, modify or remove entities, and then call <xref:Microsoft.EntityFrameworkCore.DbContext.SaveChangesAsync*> to persist those changes to the database. For more information on the basics of saving, see [Saving data](xref:core/saving/index).
+
+Note that unlike with relational databases, Azure Cosmos DB doesn't provide transaction guarantees by default across multiple documents.
+
+## Bulk execution
+
+> [!NOTE]
+> Bulk execution is being introduced in EF Core 11, which is currently in preview.
+
+By default, EF Core executes document operations sequentially when calling `SaveChangesAsync`. When saving a large number of entities, this can be slow as each operation waits for the previous one to complete before starting.
```
JoasE: This isn't completely true anymore. As you mentioned, #5100 still has to be merged, but we need to either update it here or not forget to update it in that pr

entity-framework/core/providers/cosmos/saving.md
JoasE: If we want to mention it, maybe also update it in whatsnew

entity-framework/core/providers/cosmos/saving.md
roji: Remind me where we landed - I thought we don't enable transaction batches by default? In any case, we can definitely tweak this as part of #5100 as well.

JoasE: I think this is the last discussion we had on it: https://github.com/dotnet/efcore/issues/36903#issuecomment-3444417357
And default in code is currently WhenNeeded which will group in batches where possible

entity-framework/core/providers/cosmos/saving.md
roji: Sure - this is just a simple conceptual example.

entity-framework/core/providers/cosmos/saving.md
roji: You mean in case a 2nd read needs to happen after this FindAsync, right? I'm a fan of showing the simple, minimal case in these docs. Does the current code seems reasonable enough without adding more complexity?

JoasE: > Does the current code seems reasonable enough without adding more complexity?

I'd say that it does yes

> You mean in case a 2nd read needs to happen after this FindAsync, right? I'm a fan of showing the simple, minimal case in these docs.

Not really, IIUC. Let me try to explain it a bit more, tho again I think this example is fine enough for these docs.
Let's say there was also a UpdateDocument here.
User 1 calls CreateDocument (v1) then UpdateDocument (v2)
User 2 calls GetDocument -> retrieves document v2
User 2 refreshes the pages, calls GetDocument again -> Could potentially retrieve document v1
Because we didn't set a session token cookie/header on the read.

Even tho in my experience most apps don't handle this case, it could cause stale reads

roji: Hmm, OK. Though on the other hand, if User 2 **does** set the session token on their 2nd read (the refresh), aren't they then guaranteed to get v2 and not v3 (in case some User 3 updated the document in between the two reads)? I'm not familiar enough with Cosmos session tokens to know if they function as a sort of "minimal version" or as an exact version; if it's the latter, then there would be no point in doing a refresh at all, as you're guaranteed to get the same version. Though if it does function as minimum, what you say makes perfect sense.

In any case, I think we both agree this sort of scenario should be out of scope fo this particular docs sample... At the end of the day, we're not a place for advanced Cosmos documentation - we just need to show how to access Cosmos functionality via EF. Actual usage patterns - especially advanced ones - should be in more general, non-EF docs.

JoasE: To awnser your question: It's a minimum version guarantee

entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,179 @@
...
+* Transactional batches can contain only up to 100 operations, and cannot surpass 2MB of data in total.
+* Azure Cosmos DB does not allow document writes with [pre- or post-triggers](/azure/cosmos-db/nosql/stored-procedures-triggers-udfs#triggers) to be part of a transactional batch. Because of this, any entities configured with triggers are executed separately and before any transactional batches. This can affect ordering and consistency in mixed scenarios.
+
+## Bulk execution
+
+> [!NOTE]
+> Bulk execution is being introduced in EF Core 11, which is currently in preview.
+
+Prior to version 11, the Azure Cosmos DB provider executed document operations sequentially when calling `SaveChangesAsync`; when saving a large number of entities, this was slow as each operation waits for the previous one to complete before starting. Version 11 enables [transactional batches](#transactionality-and-transactional-batches), which allow operations to be batched together for better performance. However, transactional batches can only be used against a single partition, and have various limitations (see the documentation above.
```
JoasE:
```suggestion
Prior to version 11, the Azure Cosmos DB provider executed document operations sequentially when calling `SaveChangesAsync`; when saving a large number of entities, this was slow as each operation waits for the previous one to complete before starting. Version 11 enables [transactional batches](#transactionality-and-transactional-batches), which allow operations to be batched together for better performance. However, transactional batches can only be used against a single partition, and have various limitations (see the documentation above).
```

entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,179 @@
...
+### Configuring session token management mode
+
+To enable manual session token management, configure `SessionTokenManagementMode()` when setting up your context:
+
+```csharp
+protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+    => optionsBuilder.UseCosmos(
+        "<connection string>",
+        databaseName: "OrdersDB",
+        options => options.SessionTokenManagementMode(SessionTokenManagementMode.SemiAutomatic));
```
roji, line 91: @AndriySvyryd should this be WithSessionTokenManagementMode (also the other new options maybe)?

entity-framework/core/providers/cosmos/saving.md
JoasE: I have a couple of suggestions which may be a bit nit-picky/explaining cosmos too much. If you think it's not needed just ignore them
```suggestion
Prior to version 11, the Azure Cosmos DB provider executed document operations sequentially when calling `SaveChangesAsync`; when saving a large number of entities, this was slow as each operation waits for the previous one to complete before starting. Version 11 enables [transactional batches](#transactionality-and-transactional-batches), which allow operations to be batched together. However, this only improves performance when document writes can be grouped by partition key value, along with other limitations (see the documentation above).
```

entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,179 @@
...
+
+Starting with EF 11, the EF Core Azure Cosmos DB provider leverages transactional batches by default whenever possible, providing a best-effort approximation of atomicity (and optimal performance) when saving changes. The batching behavior can be controlled by the <xref:Microsoft.EntityFrameworkCore.AutoTransactionBehavior> property, allowing developers to trade off between performance, consistency guarantees, and failure behavior depending on the application’s needs.
+
+* **Auto** (default) – Operations are grouped into transactional batches by the container and partition they affect, and with a maximum of 100 changes per batch; these batches are then executed sequentially. If a batch fails, execution stops immediately and no subsequent changes are saved. Any batches that were successfully saved before the failure remain saved. This generally provides good performance for performing multiple operations within the same partition with a best-effort approximation of atomicity.
```
JoasE: Maybe it's better to move this to just the limitations
```suggestion
* **Auto** (default) – EF automatically groups operations into transactional batches, which are executed sequentially. If a batch fails, execution stops immediately and no subsequent changes are saved, while any previously successful batches remain committed. This generally provides good performance when performing multiple operations that can be grouped by partition key value, with a best-effort approximation of atomicity.
```

entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,179 @@
...
+* **Never** – All operations are performed individually and sequentially, in the exact order they were tracked. This avoids batching and can be slower, especially for large numbers of changes. This was the behavior prior to version 11.
+* **Always** – Requires that all operations can be executed as a single transaction batch; if any operation cannot be included in a batch (e.g. because they affect different partitions), an exception is thrown. This allows you to guarantee full atomicity (and a single roundtrip) when executing `SaveChangesAsync`, but it is then up to you to manually ensure that all operations can be performed in a transactional batch.
```
JoasE:
```suggestion
* **Always** – Requires that all operations can be executed as a single transaction batch; if any operation cannot be included in a batch (see [Limitations](#Limitations)), an exception is thrown. This allows you to guarantee full atomicity (and a single roundtrip) when executing `SaveChangesAsync`, but it is then up to you to manually ensure that all operations can be performed in a transactional batch.
```

entity-framework/core/providers/cosmos/saving.md
roji: Thanks for the info, good to know these things!

entity-framework/core/providers/cosmos/saving.md
AndriySvyryd: It's too late, none of the Cosmos option methods start with a verb

entity-framework/core/providers/cosmos/saving.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Appending session tokens
+
+You can also append session tokens to the existing ones. This is useful when you have session tokens from multiple sources and want to combine them:
+
+```csharp
+// Append a session token for the default container
+context.Database.AppendSessionToken(sessionToken);
+
+// Append session tokens for all containers
```
AndriySvyryd:
```suggestion
// Append session tokens for the specified containers
```

AndriySvyryd (approved):

PR #5270, opened 2026-02-18
Document recording the latest migration id in the model snapshot

roji:
See https://github.com/dotnet/efcore/issues/37688

### DIFF
--- a/entity-framework/core/managing-schemas/migrations/teams.md
+++ b/entity-framework/core/managing-schemas/migrations/teams.md
@@ -7,10 +7,9 @@ uid: core/managing-schemas/migrations/teams
 ---
 # Migrations in Team Environments
 
-When working with Migrations in team environments, problems can arise when migrations are added by multiple developers around the same time; note that migrations aren't simply SQL scripts but also include a snapshot of the model at the time of that migration. Possible issues include:
+When working with Migrations in team environments, various problems can arise when migrations are added by multiple developers around the same time; note that migrations aren't simply SQL scripts but also include a snapshot of the model at the time of that migration.
 
-1. Developer A renames some table to X, while developer B renames the same table to Y.
-2. Developer A and B both create work branches at the same time, and generate a migration in their branches. If developer A merges their branch and then developer B does the same, the latest migration (developer B's) will have a context snapshot that does not include the changes from developer A's migration.
+For example, imagine developer A and B both create work branches at the same time, and generate a migration in their branches. If developer A merges their branch and then developer B does the same, the latest migration (developer B's) will have a context snapshot that does not include the changes from developer A's migration. This can cause various forms of corruption in later migrations.
 
 As a result, it is highly recommended to coordinate in advance and to avoid working concurrently on migrations in multiple branches when possible.
 


entity-framework/core/managing-schemas/migrations/teams.md
```diff
@@ -2,56 +2,34 @@
...
+ms.date: 02/18/2026
 uid: core/managing-schemas/migrations/teams
 ---
 # Migrations in Team Environments
 
-When working with Migrations in team environments, pay extra attention to the model snapshot file. This file can tell you if your teammate's migration merges cleanly with yours or if you need to resolve a conflict by re-creating your
-migration before sharing it.
+When working with Migrations in team environments, problems can arise when when migrations are added by multiple developers around the same time; recall that migrations aren't simply SQL scripts, but also include a snapshot of the model at the time of that migration. Possible issues include:
 
-## Merging
```
roji, line 13: @AndriySvyryd I removed this previous section of the docs, maintaining a harsher stance towards merging; following the previous instructions, you would indeed have a correct latest model snapshot (assuming the two conflicting migrations didn't change the exact same area of the model); however, the new migration would still have an incorrect model snapshot (without the changes from the previously-merged one), so IIUC if a user later rolled back to it and generated a new migration, things would get corrupt.

Let me know if this all makes sense to you... I think the new, simpler guidance is healthier - do not attempt to merge, but rather **always** roll back and recreate. We make this much simpler to detect in 11 with the new migration ID feature.

entity-framework/core/managing-schemas/migrations/teams.md
```diff
@@ -2,56 +2,34 @@
...
 author: SamMonoRT
-ms.date: 10/30/2017
+ms.date: 02/18/2026
 uid: core/managing-schemas/migrations/teams
 ---
 # Migrations in Team Environments
 
-When working with Migrations in team environments, pay extra attention to the model snapshot file. This file can tell you if your teammate's migration merges cleanly with yours or if you need to resolve a conflict by re-creating your
-migration before sharing it.
+When working with Migrations in team environments, problems can arise when when migrations are added by multiple developers around the same time; recall that migrations aren't simply SQL scripts, but also include a snapshot of the model at the time of that migration. Possible issues include:
```
AndriySvyryd:
```suggestion
When working with Migrations in team environments, problems can arise when migrations are added by multiple developers around the same time; note that migrations aren't simply SQL scripts but also include a snapshot of the model at the time of that migration. Possible issues include:
```

entity-framework/core/managing-schemas/migrations/teams.md
```diff
@@ -2,56 +2,34 @@
...
 
-## Merging
+1. Developer A renames some table to X, while developer B renames the same table to Y.
```
AndriySvyryd: This can be removed, the second point pretty much encompasses all issues

entity-framework/core/managing-schemas/migrations/teams.md
```diff
@@ -2,56 +2,34 @@
...
+2. Developer A and B both create work branches at the same time, and generate a migration in their branches. If developer A merges their branch and then developer B does the same, the latest migration (developer B's) will have a context snapshot that does not include the changes from developer A's migration.
 
-When you merge migrations from your teammates, you may get conflicts in your model snapshot file. If both changes are unrelated, the merge is trivial and the two migrations can coexist. For example, you may get a merge conflict in the customer entity type configuration that looks like this:
+As a result, it is highly recommended to coordinate in advance and to avoid working concurrently on migrations in multiple branches.
```
AndriySvyryd:
```suggestion
As a result, it is highly recommended to coordinate in advance and to avoid working concurrently on migrations in multiple branches when possible.
```

entity-framework/core/managing-schemas/migrations/teams.md
```diff
@@ -2,56 +2,34 @@
...
-=======
-b.Property<int>("LoyaltyPoints");
->>>>>>> Theirs
-```
+## Detecting diverged migration trees
 
-Since both of these properties need to exist in the final model, complete the merge by adding both properties. In many
-cases, your version control system may automatically merge such changes for you.
+> [!NOTE]
+> This feature is being introduced in EF Core 11, which is currently in preview.
```
AndriySvyryd:
```suggestion
> This feature is being introduced in EF Core 11 from preview-3 onwards.
```

entity-framework/core/managing-schemas/migrations/teams.md
```diff
@@ -2,56 +2,34 @@
...
 1. Abort the merge and rollback to your working directory before the merge
 2. Remove your migration (but keep your model changes)
 3. Merge your teammate's changes into your working directory
 4. Re-add your migration
 
-After doing this, the two migrations can be applied in the correct order. Their migration is applied first, renaming
-the column to *Alias*, thereafter your migration renames it to *Username*.
-
-Your migration can safely be shared with the rest of the team.
+After doing this, your later migration is cleanly based on top of any migrations that have been added in the meantime, and its context snapshot contains all previous changes. Your migration can safely be shared with the rest of the team.
```
AndriySvyryd:
```suggestion
After doing this, your migration is cleanly based on top of any migrations that have been added in the other branch, and its context snapshot contains all previous changes. Your migration can now be safely shared with the rest of the team.
```

AndriySvyryd (approved):

entity-framework/core/managing-schemas/migrations/teams.md
roji: No objection, but note that these docs will only get merged once preview-3 is out, so I think there's little value in specifying the preview version in the docs.

PR #5328, opened 2026-04-14
Fix SQL stored procedure syntax by using SELECT instead of RETURN

srdrtmy:
In SQL Server, RETURN is intended only for integer status codes so that SELECT is the correct for returning computed string results from a stored procedure


roji (approved):
Thanks, makes sense.

PR #5332, opened 2026-04-14
Convert the docs generation custom agent to a skill

roji:
Easier to use locally and more standard nowadays.

### DIFF
--- a/.agents/skills/generate-docs/SKILL.md
+++ b/.agents/skills/generate-docs/SKILL.md
@@ -1,12 +1,11 @@
 ---
-name: EF Documentation Generator
-description: This agent creates documentation PRs in the EF documentation site when new features are implemented in EF Core.
-disable-model-invocation: true
+name: Generate EF documentation
+description: Create documentation PRs in the EF documentation site when new features are implemented in EF Core.
 ---
 
 # Document new EF features
 
-Given an EF issue by the user, this custom agent generates documentation for features introduced in that issue and submits a PR to the EF docs repo (dotnet/EntityFramework.Docs).
+Given an EF issue and/or PR, this skill generates documentation for features introduced in that issue/PR and submits a PR to the EF docs repo (dotnet/EntityFramework.Docs).
 
 ## Target branch
 


AndriySvyryd (approved):

valuDex:
Here's a previous response from the AI ​​alongside your code, but based on the code provided, setting `MyDbContext` to inject a custom telemetry interceptor is programmatically correct and follows best practices in EF Core. However, since you are injecting `reTelemetryInterceptor` into the constructor, there is a major architectural trap to avoid related to service lifetimes. Service Lifecycle Trap: In ASP.NET Core, `DbContext` is typically registered as a Scoped service - that is, it is created once per HTTP request. If `reTelemetryInterceptor` is registered as a Transient or Scoped service, the code will work perfectly. If it is registered as a Singleton service - which is common for logging clients - you risk a “captive dependency” problem if the interceptor attempts to hold scoped resources. Conversely, if you register `DbContext` as a Singleton service - for example using `AddDbContextFactory` - injecting a scoped interceptor will raise a runtime exception related to the lifecycle check. How to register in `Program.cs`: To ensure that this setup runs smoothly and without runtime errors, register both the interceptor and context in the `Program.cs` file as follows: `var builder = WebApplication.CreateBuilder(args);`

// 1. Register the custom interceptor first
// (use AddScoped if it depends on other scoped services, or AddSingleton if it doesn't hold state/data)
builder.Services.AddScoped<reTelemetryInterceptor>();

// 2. Register DbContext
builder.Services.AddDbContext<MyDbContext>((serviceProvider, options) =>
{
builder.Configuration.GetConnectionString("DefaultConnection");

// Note: Since you are dealing with AddInterceptors inside OnConfiguring,
// EF Core will automatically resolve the dependency via Dependency Injection (DI) when the constructor is run.
});
Use the code with caution. Alternatives: Modern approach in EF Core (7.0 and above): If you want to keep the `DbContext` constructor clean without passing the interceptor manually, EF Core lets you fetch resolving interceptors directly from the Application Service Provider inside `OnConfiguring`: `protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)`
{
// Fetch the interceptor directly from the Dependency Injection (DI) system without having to inject it through the constructor
optionsBuilder.AddInterceptors();
} }
Use the code with caution. Note: For the `AddInterceptors()` method (which is non-parameterized) to work, you must register the interceptor in the `Program.cs` file using the command: `builder.Services.AddScoped<IInterceptor, reTelemetryInterceptor>();`. To help you run this code on a Mac, please tell me: What database provider are you connecting to (eg: SQL Server, PostgreSQL, or SQLite)? Do you get a specific error message (such as `Cannot resolve scoped service from root provider`) when you try to run the application?
We continue to warn and warn and warn, and we state that any similar product or one born from the womb of the rule of measurement belongs to: The link above represents the remainder of the application interface. Following the breach, we require robust security measures to ensure seamless integration and rapid project completion, paving the way to hold the perpetrators accountable. Mastering the work and its mechanization in digital rooting, based on the standard framework model: standards framework + design framework + organizational framework + evaluation framework + dictionary mechanization framework.
License to publish and distribute
02508065457329
Entity Name:

TASNEF COMPANY

Legal Reference:

Registered Project Title:

Standardized Cloud Computing with Digital Standards
Deposit Number: 1447/6257
ISBN: 978-603-06-0050-2
Deposited with: King Fahd National Library (Kingdom of Saudi Arabia)
Publisher: Media Regulatory Authority

CLASSIFICATION COMPANY
Royal Commission Protection Number: 122431087 (issued 2022)
Renewed in 2025
To:
SA 1020256781
OMB 0280-0625 FR 3046-
Standardized Cloud Computing with Digital Standards
IP: 192.168.1.64
SAP: P2005320033

PR #5366, opened 2026-05-23
docs: Document that ApplyConfigurationsFromAssembly requires a parameterless constructor

### LINKED ISSUE #3207: IEntityTypeConfiguration.Configure() is not fired for entities without default constructor
## File a bug

I looked through [open](https://github.com/dotnet/efcore/issues) and [closed](https://github.com/dotnet/efcore/issues?q=is%3Aissue+is%3Aclosed) issues and haven't found any similar so here is my report:

For simple enity 
```
    public class User: IEntityTypeConfiguration<User>
    {
        public long Id { get; set; }

        public string Name { get; set; } = string.Empty;
        
        public void Configure(EntityTypeBuilder<User> user)
        {
            user.HasIndex(x => x.Name);
        }
    }
```
when I use IEntityTypeConfiguration to reverse DbContext configuration from context `OnModelCreating()` into entity itself it works fine generating migration having `migrationBuilder.CreateIndex()`

```
        protected override void Up(MigrationBuilder migrationBuilder)
        {
            migrationBuilder.CreateTable(
                name: "user",
                columns: table => new
                {
                    id = table.Column<long>(type: "bigint", nullable: false)
                        .Annotation("Npgsql:ValueGenerationStrategy", NpgsqlValueGenerationStrategy.IdentityByDefaultColumn),
                    name = table.Column<string>(type: "text", nullable: false)
                },
                constraints: table =>
                {
                    table.PrimaryKey("pk_user", x => x.id);
                });

            **migrationBuilder.CreateIndex(
                name: "ix_user_name",
                table: "user",
                column: "name");**
        }
```

If I want to have custom constructor like this

```
    public class User : IEntityTypeConfiguration<User>
    {
        public User(string name)
        {
            Name = name;
        }

        public long Id { get; set; }

        public string Name { get; set; } = string.Empty;

        public void Configure(EntityTypeBuilder<User> user)
        {
            user.HasIndex(x => x.Name);
        }
    }
```

it generates empty migration unless `public DbSet<User> Users { get; set; } = null!;` is included into DbContext explicitly. Migration generated then is as follows:
```
protected override void Up(MigrationBuilder migrationBuilder)
        {
            migrationBuilder.CreateTable(
                name: "users",
                columns: table => new
                {
                    id = table.Column<long>(type: "bigint", nullable: false)
                        .Annotation("Npgsql:ValueGenerationStrategy", NpgsqlValueGenerationStrategy.IdentityByDefaultColumn),
                    name = table.Column<string>(type: "text", nullable: false)
                },
                constraints: table =>
                {
                    table.PrimaryKey("pk_users", x => x.id);
                });
        }
```

which is obvious proof that `IEntityTypeConfiguration.Configure() ` is not fired during configuration.

_Expected behavior is to have the same migration no matter what constructors entity has cause DB model does not depend on them._

Ofcourse I can add default constructor explicitly, but how can I have custom one only e.g. for the purpose to have insuranse `Name` property is set ?

### Include provider and version information

EF Core version: 5.0.2
Database provider: Npgsql.EntityFrameworkCore.PostgreSQL
Target framework:  .NET 5.0)
Operating system: Windows 10
IDE:  Visual Studio 2019 16.8.6

Aykuttonpc:
## Summary

Fixes #3207.

`ApplyConfigurationsFromAssembly` only instantiates configuration types that have a parameterless constructor, but this isn't currently documented. Users hit surprising behavior when their `IEntityTypeConfiguration<T>` implementations have non-default constructors (e.g. for DI).

This PR adds a `[!NOTE]` block to `entity-framework/core/modeling/index.md` next to the existing "order is undefined" note, describing:

- A parameterless constructor is required — may be public or non-public (implementation: `BindingFlags.Public | BindingFlags.NonPublic` + `Activator.CreateInstance(type, nonPublic: true)` in [ModelBuilder.cs](https://github.com/dotnet/efcore/blob/main/src/EFCore/ModelBuilder.cs)).
- Types lacking such a constructor are skipped, with a `SkippedEntityTypeConfigurationWarning` logged.
- To apply a configuration whose constructor requires arguments, instantiate it manually and pass it to `ApplyConfiguration`.

Mirrors the style of the adjacent NOTE.

### DIFF
--- a/entity-framework/core/modeling/index.md
+++ b/entity-framework/core/modeling/index.md
@@ -42,6 +42,9 @@ It is possible to apply all configuration specified in types implementing `IEnti
 > [!NOTE]
 > The order in which the configurations will be applied is undefined, therefore this method should only be used when the order doesn't matter.
 
+> [!NOTE]
+> `ApplyConfigurationsFromAssembly` only instantiates configuration types that have a parameterless constructor; the constructor may be public or non-public. Types whose constructors require parameters are skipped, and a `SkippedEntityTypeConfigurationWarning` is logged. To apply such configurations, instantiate them manually and pass them to <xref:Microsoft.EntityFrameworkCore.ModelBuilder.ApplyConfiguration%2A>.
+
 #### Using `EntityTypeConfigurationAttribute` on entity types
 
 Rather than explicitly calling `Configure`, an <xref:Microsoft.EntityFrameworkCore.EntityTypeConfigurationAttribute> can instead be placed on the entity type such that EF Core can find and use appropriate configuration. For example:


AndriySvyryd (approved):

AndriySvyryd:
Thanks for your contribution!

PR #5420, opened 2026-07-09
Update warning about split queries and ordering

ali-alo:
Removes a redundant use of "each" in the EF Core documentation.


AndriySvyryd (approved):
Thanks for your contribution

